PR #3, opened 2018-10-24
Organising by folders

RuiAAPeres:


mluisbrown:
We could (should?) add a section for Proposals, so I have somewhere to put the `Current` proposal.

PR #4, opened 2018-10-28
[Proposal] `Current` alternative to dependency injection.

mluisbrown:
Add the proposal to use the `Current` alternative to dependency injection.

(Added by Wagner)
You can more easily read the full document [here](https://github.com/Babylonpartners/ios-playbook/blob/michael/control-the-world/TechnicalDocuments/ControlTheWorld.md)


ilyapuchka:
@mluisbrown I think we should touch a bit more on how `Current` is going to be used with features that are extracted into frameworks, i.e. appointments, clinical records, chat bot, and how this relates to SDK (even if it's just mention that it's not needed by SDKs and so should not be used there).

mluisbrown:
@Ben-Henshall Making current mutable only for testing is likely to be a bit cumbersome. The point free guys suggest having a linter rule for checking for mutations of `Current`. In our environment it will be an issue because we will have to initialise `Current` in it's framework, and only set some of its values on App Startup (eg, Analytics, AB testing), so they will be mutated at runtime. I'm not sure if there is a good way around this 😕

ilyapuchka:
@mluisbrown good point.
Alternative to linter we might do a runtime check like this, though I didn't try if observers work on global variables, I'd expect them to work though.

```swift
var Current = Current() {
  didSet {
     assert(#function == "function we allow to mutate Current in, probably something in AppDelegate")
  }
}
```

Of course it will be not compile time check as if it would be with linter. And we are going into implementation detail that we can discuss later when we come to implementing this proposal.

stephencelis:
Very cool! We definitely find it useful to mutate the world in playgrounds and development, so an alternative might be:

``` swift
#if RELEASE
let Current = World()
#else
var Current = World()
#endif
```

TechnicalDocuments/ControlTheWorld.md
```diff
@@ -0,0 +1,116 @@
...
+
+Instead I propose that, for now, we put everything that is currently in `AppDependencies` into `Current`, along with `date`, `calendar`, `timezone` and `locale` (which we currently are not even injecting).
+
+This will already save us a lot of DI boilerplate and give us the ability to simulate different date, calendar, timezone and locale settings easily.
+
+Once we get more familiar with using `Current`, we can explore using the `struct` approach above for analytics and AB testing services.
+
+## Where is the boundary between Current and using Dependency Injection?
+
+In previous discussions with the team it was agreed that we can / should use `Current` as far as our Builders and no further. The Builders should still continue to inject any dependencies required by View Models, View Controllers and Flow Controllers.
```
RuiAAPeres: I think it would be relevant to have a diagram of our architecture and the boundary.

mluisbrown: @RuiAAPeres here's a diagram for you 😛

TechnicalDocuments/ControlTheWorld.md
```diff
@@ -0,0 +1,120 @@
...
+
+* Avoid having to pass singleton dependencies down through multiple layers of injection.
+
+* Ability to easily create different app "environments" to simulate running on different dates, timezones and locales, for example.
+
+* Ability to have "mock" environments for testing. For example, testing that analytics events are correctly being sent is typically quite difficult. With a mock analytics service in a mock app environment this becomes trivial, and without the need to inject an analytics service protocol to every part of the app.
+
+## Implementation
+
+Define a `struct` to provide dependencies of the environment external to the app (the "World" in which the app lives), and static configuaration dependencies:
```
wltrup: Misspelling: `configuaration` -> `configuration`

TechnicalDocuments/ControlTheWorld.md
```diff
@@ -0,0 +1,120 @@
...
+
+Now all we need is to instantiate an instance of `World` called `Current` which is our current environment. By default this will instantiate the production dependencies:
+
+```swift
+var Current = World()
+```
+
+## Usage
+
+Everywhere where we are referring to one of the dependencies now stored in `Current`, use `Current`: eg, instead of `Date()` use `Current.date()`.
```
wltrup: > Everywhere where we are referring to one of the dependencies now stored in `Current`, use `Current`

That's a strangely worded sentence, imo. Perhaps

"Everywhere where we'd be using a dependency, use `Current` to access it instead. For example, instead of `Date()` use `Current.date()`."

TechnicalDocuments/ControlTheWorld.md
wltrup:
```suggestion
Define a `struct` to provide dependencies of the environment external to the app (the "World" in which the app lives), and static configuration dependencies:
```

TechnicalDocuments/ControlTheWorld.md
wltrup:
```suggestion
Everywhere where we'd be using a dependency, use `Current` to access it instead. For example, use `Current.date()` as opposed to `Date()`.
```

TechnicalDocuments/ControlTheWorld.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Introduction
+
+This proposal is heavily based on episodes [16](https://www.pointfree.co/episodes/ep16-dependency-injection-made-easy) and [18](https://www.pointfree.co/episodes/ep18-dependency-injection-made-comfortable) of [Pointfree](https://www.pointfree.co) and Stephen Celis's talk [How to Control the World](https://vimeo.com/291588126) at NSSpain 2018. Slides of that talk are [also available](https://speakerdeck.com/stephencelis/how-to-control-the-world).
+
+## Motivation
+
+To overcome some of the disadvantages of dependency injection:
+
+* Adding a new dependency a long way down the chain can mean passing that dependency all the way down through many layers. Many of those layers only need to get passed the dependency in order to pass it down to the next layer. This leads to a lot of source files being changed just to add a single dependency.
```
diegopetrucci:
```suggestion
* Adding a new dependency down the chain can mean passing that dependency all the way down through many layers. Many of those layers only need to get passed the dependency in order to pass it down to the next layer. This leads to a lot of source files being changed just to add a single dependency.
```

diegopetrucci:
I have added a bunch of comments, but keep in mind that they can be ignored. This is a great proposal and I am looking forward to implement it in our codebase. Great stuff Michael!

TechnicalDocuments/ControlTheWorld.md
```diff
@@ -0,0 +1,121 @@
...
+
+In previous discussions with the team it was agreed that we can / should use `Current` as far as our Builders and no further. The Builders should still continue to inject any dependencies required by View Models, View Controllers and Flow Controllers.
+
+
+## Where will Current live?
+
+In order for `Current` to be accessible from all of the Babylon frameworks, we should create a new framework 🙈 , maybe named `BabylonWorld`, which can be imported into all our frameworks that require it.
+
+<p align="center">
+<img src="arch.png">
```
wltrup: Too large! I'm suggesting 50% of its current size.

```suggestion
<img src="arch.png" width="596" height="405">
```

TechnicalDocuments/ControlTheWorld.md
```diff
@@ -0,0 +1,120 @@
...
+
+Instead I propose that, for now, we put everything that is currently in `AppDependencies` into `Current`, along with `date`, `calendar`, `timezone` and `locale` (which we currently are not even injecting).
+
+This will already save us a lot of DI boilerplate and give us the ability to simulate different date, calendar, timezone and locale settings easily.
+
+Once we get more familiar with using `Current`, we can explore using the `struct` approach above for analytics and AB testing services.
+
+## Where is the boundary between Current and using Dependency Injection?
+
+In previous discussions with the team it was agreed that we can / should use `Current` as far as our Builders and no further. The Builders should still continue to inject any dependencies required by View Models, View Controllers and Flow Controllers.
```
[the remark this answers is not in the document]
mluisbrown: I've used a slightly different wording. I'm not keen about "in the most specific way possible", which kind of implies that instead of injecting `VisualDependenciesProtocol` you would have to inject only the bits of it that you need (eg `brandColors` or `appColors`) separately.

TechnicalDocuments/ControlTheWorld.md
mluisbrown: You only need to specify the width. The height will follow. I'm changing this from my side 👍

mluisbrown:
@RuiAAPeres can this be merged?

TechnicalDocuments/ControlTheWorld.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Benefits
+
+* Avoid having to pass singleton dependencies down through multiple layers of injection.
+
+* Ability to easily create different app "environments" to simulate running on different dates, timezones and locales, for example.
+
+* Ability to have "mock" environments for testing. For example, testing that analytics events are correctly being sent is typically quite difficult. With a mock analytics service in a mock app environment this becomes trivial, and without the need to inject an analytics service protocol to every part of the app.
+
+## Implementation
```
andersio: I am not sure if the `Analytics` example should be included here, or should it at least be clarified as an example but not as part of the proposal.

mluisbrown: I've added a not about this.

RuiAAPeres:
This proposal has been approved. The planning on how it will be implemented will be done internally. ✅

PR #9, opened 2018-11-08
[Style guide] Adds several more sections

wltrup:
⚠️⚠️⚠️
In addition to reviewing the document as a whole, please **add comments (if you have any) to the specific threads I have already created because, in this PR, those are really the more important points of potential contention.**
⚠️⚠️⚠️

Full document [here](https://github.com/Babylonpartners/ios-playbook/blob/wagner/style-guide/naming/Style-guide/README.md)


Style-guide/README.md
```diff
@@ -71,8 +68,8 @@ Descriptive and consistent naming makes software easier to read and understand.
 
 - striving for clarity at the call site
 - prioritizing clarity over brevity
-- using camel case (not snake case)
-- using uppercase for types (and protocols), lowercase for everything else
+- using `camelCase` (not `snake_case`)
+- using uppercase initials for types (and protocols), lowercase first-initial for everything else
```
wltrup: Do we need examples here or is it already clear enough?

Style-guide/README.md
```diff
@@ -71,8 +68,8 @@ Descriptive and consistent naming makes software easier to read and understand.
...
 - including all needed words while omitting needless words
```
wltrup: ditto

Style-guide/README.md
```diff
@@ -71,8 +68,8 @@ Descriptive and consistent naming makes software easier to read and understand.
...
 - using names based on roles, not types
```
wltrup: ditto

Style-guide/README.md
```diff
@@ -71,8 +68,8 @@ Descriptive and consistent naming makes software easier to read and understand.
...
 - sometimes compensating for weak type information
```
wltrup: This isn't clear to me. Is it to anyone else?

Style-guide/README.md
```diff
@@ -81,7 +78,7 @@ Descriptive and consistent naming makes software easier to read and understand.
 - naming methods for their side effects
   - verb methods follow the -ed, -ing rule for the non-mutating version
```
wltrup: I think this needs clarification and examples.

Style-guide/README.md
```diff
@@ -81,7 +78,7 @@ Descriptive and consistent naming makes software easier to read and understand.
...
   - noun methods follow the formX rule for the mutating version
```
wltrup: ditto

Style-guide/README.md
```diff
@@ -81,7 +78,7 @@ Descriptive and consistent naming makes software easier to read and understand.
...
-  - boolean types should read like assertions
+  - boolean types should read like assertions (`isLoading`, `isHidden`, etc)
   - protocols that describe _what something is_ should read as nouns
```
wltrup: We currently have many violations of this, eg, `SomethingProtocol`

Style-guide/README.md
```diff
@@ -81,7 +78,7 @@ Descriptive and consistent naming makes software easier to read and understand.
...
   - protocols that describe _a capability_ should end in _-able_ or _-ible_
```
wltrup: `ImageCatalogueAware` violates this. There may be others.

Style-guide/README.md
```diff
@@ -92,28 +89,13 @@ Descriptive and consistent naming makes software easier to read and understand.
 - giving the same base name to methods that share the same meaning
 - avoiding overloads on return type
```
wltrup: Needs clarification and examples

Style-guide/README.md
```diff
@@ -81,39 +78,24 @@ Descriptive and consistent naming makes software easier to read and understand.
...
   - verb methods follow the -ed, -ing rule for the non-mutating version
   - noun methods follow the formX rule for the mutating version
-  - boolean types should read like assertions
+  - boolean types should read like assertions (`isLoading`, `isHidden`, etc)
   - protocols that describe _what something is_ should read as nouns
   - protocols that describe _a capability_ should end in _-able_ or _-ible_
 - using terms that don't surprise experts or confuse beginners
-- generally avoiding abbreviations
-- using precedent for names
+- generally avoiding abbreviations // a few well-known ones are allowed, such as, `Html` and `Id`. Any others?
```
wltrup: (I've had to change the line above to trigger a diff, else we couldn't comment on the line. The inline comment will be removed.)

A few well-known ones are allowed, such as, `Html` and `Id`. Any others?

Style-guide/README.md
```diff
@@ -81,39 +78,24 @@ Descriptive and consistent naming makes software easier to read and understand.
...
+- using precedent for names // ???
```
wltrup: (I've had to change the line above to trigger a diff, else we couldn't comment on the line. The inline comment will be removed.)

Does anyone understand what this is about?

Style-guide/README.md
```diff
@@ -81,39 +78,24 @@ Descriptive and consistent naming makes software easier to read and understand.
...
 - preferring methods and properties to free functions
-- casing acronyms and initialisms uniformly up or down
+- casing acronyms and initialisms uniformly up or down // I think we use camelcase, as in `Id` and `Html`
```
wltrup: (I've had to change the line above to trigger a diff, else we couldn't comment on the line. The inline comment will be removed.)

I think we use camelcase, as in `Id` and `Html` instead.

Style-guide/README.md
```diff
@@ -289,7 +269,7 @@ var deviceModels: [String]
 
 ## Spacing
 
-* Indent using 2 spaces rather than tabs to conserve space and help prevent line wrapping. Be sure to set this preference in Xcode and in the Project settings as shown below:
+* Indent using 4 spaces rather than tabs to conserve space and help prevent line wrapping. Be sure to set this preference in Xcode and in the Project settings as shown below:
 
 ![Xcode indent settings](screens/indentation.png)
```
wltrup: Need to update the picture with the correct number of tabs, once agreed to.

Style-guide/README.md
diegopetrucci: i think it's fine given it's a swift convention

Style-guide/README.md
diegopetrucci: I would add an example, maybe with something like `tvShowID: String`

Style-guide/README.md
```diff
@@ -344,12 +324,14 @@ class TestDatabase : Database {
 
 ## Comments
 
-When they are needed, use comments to explain **why** a particular piece of code does something. Comments must be kept up-to-date or deleted.
+When they are needed, use comments to explain **why** a particular piece of code does something. Comments must be kept up-to-date, or altogether deleted.
```
diegopetrucci: I would change this to be something more like comments should be avoided at all costs, with the code being self-documenting. The only exceptions are when the code deals with a complex domain-specific API that not everyone is used to working with, or the code is complicated because of legacy or bugs reasons.

Style-guide/README.md
RuiAAPeres: is HTML an abbreviation or an acronym?

Style-guide/README.md
RuiAAPeres: We use `ID`, unless it's a prefix `idOfSomething`.

Style-guide/README.md
wltrup: > adding our rule about associating a TODO comment with a Jira ticket would be useful

Good recommendation. Will do.

Style-guide/README.md
wltrup: Right, so it's full upper in the case of a type and full lower in the case of part of an identifier.

Style-guide/README.md
wltrup: Acronym. HyperText Markup Language.

Style-guide/README.md
zzcgumn: Our frequent use of the `SomethingProtocol` is a side effect of wrapping our business controllers in protocols. I am not sure how we should change this. To have `PaymentsController` conform to `PaymentsControlling` instead of `PaymentsControllerProtocol` does not sound right to me. Perhaps the underlying problem is that we should model business controllers as protocols.

Style-guide/README.md
```diff
@@ -60,7 +58,6 @@ Uncontroversial, non-additive changes such as misspellings, grammar or compiler
 * [Smiley Face](#smiley-face)
```
mr-v: This section can be removed :)

Style-guide/README.md
mr-v: Swift Guidelines already have good examples, so maybe let's use it. Maybe not copy examples verbatim, but add links to specific sections. Each paragraph has its own `id` in the HTML, so it's doable. 😉 Would be super cool to have those Swift Guidelines links here!

https://swift.org/documentation/api-design-guidelines/#include-words-to-avoid-ambiguity
https://swift.org/documentation/api-design-guidelines/#follow-case-conventions

Style-guide/README.md
mr-v: Do you think those Swift Guidelines links I mentioned would help here with clarification?

Style-guide/README.md
mr-v: > If an associated type is so tightly bound to its protocol constraint that the protocol name is the role, avoid collision by appending Protocol to the protocol name

Swift Guidelines say that at times it's "kinda ok".

Style-guide/README.md
mr-v: >  Any others?

URL, HTTP. I think the rule of thumb could be if abbreviation already appears in iOS SDK or stdlib, it's ok to use it. Swift Guidelines have pretty nice tip for this:

> The intended meaning for any abbreviation you use should be easily found by a web search.

Style-guide/README.md
mr-v: Another instance where Swift Guidelines link would be helpful. I think it's ok to keep this list of takeaways from Swift Guidelines succinct as long as we provide a link, so there's a reference for detailed explanation.
Otherwise we would just be duplicating swift.org content.

Style-guide/README.md
mr-v: In iOS SDK acronyms are always uppercased in types , e.g. `HTTPURLResponse`. If there are places we're not doing it, we should fix it.

Style-guide/README.md
mr-v: > I would change this to be something more like comments should be avoided at all costs, with the code being self-documenting. The only exceptions are when the code deals with a complex domain-specific API that not everyone is used to working with, or the code is complicated because of legacy or bugs reasons.

I think the **why** covers it in general, but we could add extra clarification about not commenting **what** code does and point about documenting with code instead.

Style-guide/README.md
wltrup: I haven't got that far yet.

PR #10, opened 2018-11-19
Switch to SnapshotTesting 📸

mluisbrown:
As a result of [this slack conversation](https://babylonhealth.slack.com/archives/GD6J0QWQ5/p1542619318004600) I've added a proposal that we switch our snapshot testing library to [SnapshotTesting](https://github.com/pointfreeco/swift-snapshot-testing).

📸


mluisbrown:
@ilyapuchka yes, `SnapshotTesting` allows you to set the tolerance.

TechnicalDocuments/SnapshotTesting.md
```diff
@@ -0,0 +1,50 @@
...
+## Introduction
+
+This proposal is to replace the library that we use for snapshot testing ([`iOSSnapshotTestCase`](https://github.com/uber/ios-snapshot-test-case/)) with [`SnapshotTesting`](https://github.com/pointfreeco/swift-snapshot-testing).
+
+## Motivation
+
+### Cons of `iOSSnapshotTestCase`
+* `iOSSnapshotTestCase` is an old library written in Objective-C that can only snapshot a `UIView` or `CALayer`.
+* It has on occasion seemed to be quite fragile, reporting diffs when nothing changed, to the extent that we have had to increase the failure tolerance to quite a high value, and have even on a couple of occasions considered dropping snapshot testing altogether.
+* it stores all references images in a single folder for all tests
```
ilyapuchka: in fact it can be configured with launch arguments, we just do it like so for some reason

TechnicalDocuments/SnapshotTesting.md
```diff
@@ -0,0 +1,50 @@
...
+* It is easy to extend the library to add explicit support for snapshotting of other types, instead of just using the `Any` strategy.
+* Snapshots are stored in a `__Snapshots__` folder in the same folder hierarchy as the test source. This makes it easy to find relevant snapshots.
+* It uses `XCTAttachment`s to store failure snapshots and diffs, which makes these available directly from the Xcode test report navigator, without requiring any other software or needing to hunt for them.
+* It is fully Linux compatiable. Allowing snapshot testing to be used on projects that are not iOS or macOS (eg our merge bot).
+* Even if `record` is set to `false` it will still record reference snapshots if none currently exist, so you only have to set it to `true` when updating snapshots becuase of changes.
+
+### Cons of `SnapshotTesting`
+* It's a very new library and in active development. The developers themselves state *"This library should be considered alpha, and not stable. Breaking changes will happen often."* However, provided we stick to a specific version or commit in the Podfile, this should not be a big issue.
```
stephencelis: This changes today 😄

https://www.pointfree.co/blog/posts/23-snapshottesting-1-0-delightful-swift-snapshot-testing

TechnicalDocuments/SnapshotTesting.md
```diff
@@ -0,0 +1,50 @@
...
+
+Average times for running `Cmd+U` on my machine in Xcode for `BabylonChatBotUI`:
+
+iOSSnapshotTestCase | SnapshotTesting
+-------------------|-----------------
+11 seconds | 15 seconds
+
+So, it would seem that `SnapshotTesting` is roughtly 36% slower than `iOSSnapshotTestCase` 😞 .
+
+I have no idea where the speed difference is. Could well be just that unoptimized Swift is slower than Objective-C 🤷🏼‍♀️ . Still, I don't think this is a reason to reject the benefits of `SnapshotTesting`.
```
mbrandonw: Thanks for these benchmarks!

I think this could be explained by the fact that we support async snapshots (for snapshotting things that require callbacks or delegates, like `WKWebView`), and so every assertion is using an `XCTWaiter` behind the scenes. We could probably skip that for assertions that we know are not async.

We'll investigate this soon.

TechnicalDocuments/SnapshotTesting.md
```diff
@@ -0,0 +1,50 @@
...
+### Cons of `iOSSnapshotTestCase`
+* `iOSSnapshotTestCase` is an old library written in Objective-C that can only snapshot a `UIView` or `CALayer`.
+* It has on occasion seemed to be quite fragile, reporting diffs when nothing changed, to the extent that we have had to increase the failure tolerance to quite a high value, and have even on a couple of occasions considered dropping snapshot testing altogether.
+* it stores all references images in a single folder for all tests
+* you have to manually find the failure images and diffs to see what went wrong (or use an app like FBSnapshotsViewer)
+* You always have to explicityly set `recordMode` to `true` even if the test has no reference images yet.
+
+### Pros of `SnapshotTesting`
+* `SnapshotTesting` is a brand new library written in Swift by the [Pointfree](pointfree.co) duo. It's a great example of a protocol based API using struct based protocol witnesses as explained in the Pointfree videos on the subject.
+* It supports snapshotting anything, not just `UIView` or `CALayer`. For example, you could snapshot a `URLRequest` instead of having to laboriously type out the expected values by hand.
```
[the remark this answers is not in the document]
stephencelis: In the JavaScript/React community, snapshot testing most commonly refers to the text-based version of snapshotting. We think it’s appropriate for any kind of data and hope that the Swift community will get used to expanding the term’s use 😄

TechnicalDocuments/SnapshotTesting.md
ilyapuchka: This will remove some headache from our side as well as we had to implement exactly this case on our side (waiting for web view to load in snapshot tests) 🎉

TechnicalDocuments/SnapshotTesting.md
```diff
@@ -0,0 +1,50 @@
...
+### Benchmarks
+I used `BabylonChatBotUI` for running some rudimentary benchmarks between the two libraries as it seems to have the most snapshot tests and, more importantly, it seems to have *only* snapshot test so nothing else would pollute the times.
+
+Average times for running `Cmd+U` on my machine in Xcode for `BabylonChatBotUI`:
+
+iOSSnapshotTestCase | SnapshotTesting
+-------------------|-----------------
+11 seconds | 15 seconds
+
+So, it would seem that `SnapshotTesting` is roughtly 36% slower than `iOSSnapshotTestCase` 😞 .
```
stephencelis: This is good to know! Thanks for benchmarking. There are definitely some avenues to explore here for optimization.

TechnicalDocuments/SnapshotTesting.md
mluisbrown: Oh wow, I totally missed the async snapshot support. This might explain why in the few async snapshot scenarios we have using `SnapshotTesting` are causing a crash (`EXC_BAD_ACCESS`). I didn't mention this before as I was trying to nail down the cause a bit better and get an easily reproducible case.

Curiously, the test runs fine, but it's when the XCTest UIApplication is torn down that there is an `objc_msgSend` crash. I will see if switching to the Async API solves this 👍

TechnicalDocuments/SnapshotTesting.md
```diff
@@ -0,0 +1,50 @@
+Improved Snapshot Testing
+=========================
+
+## Introduction
+
+This proposal is to replace the library that we use for snapshot testing ([`iOSSnapshotTestCase`](https://github.com/uber/ios-snapshot-test-case/)) with [`SnapshotTesting`](https://github.com/pointfreeco/swift-snapshot-testing).
+
+## Motivation
+
+### Cons of `iOSSnapshotTestCase`
```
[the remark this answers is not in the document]
mluisbrown: I think we would still continue to use image snapshots for UIViews. We might use `recursiveDescription` also, but IMO the images are important.

RuiAAPeres:
@mluisbrown were you able to make any progress on 2ef7513?

stephencelis:
@mluisbrown We have a proof-of-concept branch here that optimizes how XCTWaiter is used to avoid unnecessary XCTest expectations: https://github.com/pointfreeco/swift-snapshot-testing/compare/optimize-async

It might help avoid the crash you were seeing and hopefully will improve performance? Care to re-run your benchmark and let us know?

mluisbrown:
@RuiAAPeres I made some progress on the crashing issue by using Async pullback, but it’s still not totally resolved. I think it is solvable though, just needs a bit more work. 

@stephencelis thanks for the POC. I’ll try out as soon as I have time, probably next week 👍

mluisbrown:
@stephencelis I just tried the XCWaiter POC from the `optimize-async` branch and there's is no change in the benchmark unfortunately, still running around 15s each time for the same batch of tests 😕 

It also didn't help with the crashing, although I have made more progress here. I discovered two things:
* if we have any tests that use Nimble's `toEventually()` async testing they crash in [Nimble's code to manage the runloop](https://github.com/Quick/Nimble/blob/d76a0ef09ee7614f69e58b9001d4668d26b26f4c/Sources/Nimble/Utils/Await.swift#L267)
* With those tests disabled, the test suite completes, but when the app is shutting down after testing (in `UIApplicationMain`) we get a crash trying which appears to me in a call to `objc_msgSend`.

I enabled Zombie object diagnostics and in both of the above situations just before the crash I see this in the console:

`*** -[CALayer setNeedsLayout]: message sent to deallocated instance 0x600000582da0`

I'm now convinced that this is the root cause of both the issues above. In fact it's likely the cause of all the crashing we have seen. 

In all our codebase, the only two places where `setNeedsLayout` is being called directly on a `CALayer` are in `CALayer.swift` in `SnapshotTesting` [here](https://github.com/pointfreeco/swift-snapshot-testing/blob/master/Sources/SnapshotTesting/Snapshotting/CALayer.swift#L18) and [here](https://github.com/pointfreeco/swift-snapshot-testing/blob/master/Sources/SnapshotTesting/Snapshotting/CALayer.swift#L42).

It's also relevant since we are doing all our snapshot tests using `CALayer` instead of `UIView` or `UIViewController` as that was the only way I could ensure the entire window, including the nav bar, was included in the snapshot.

I still can't figure why this would be happening, but we're getting closer to the root cause!

mluisbrown:
> It's also relevant since we are doing all our snapshot tests using CALayer instead of UIView or UIViewController as that was the only way I could ensure the entire window, including the nav bar, was included in the snapshot.

@stephencelis I tried changing our tests to use `UIView` and `UIViewController` instead of `CALayer` but the problem persisted.

At this point I think it's probably better if I open an issue on the `SnapshotTesting` repo to continue the discussion about the async crashes.

mluisbrown:
For now, the issue with crashing when running with other async tests is insurmountable so I'm closing this proposal 😞

mluisbrown:
On second thoughts, I'm going to keep this open but mark it as WIP.

RuiAAPeres:
Any update on this @mluisbrown ?

mluisbrown:
Unfortunately not @RuiAAPeres 😢  I have a branch where I want to try some more debugging to try and isolate the issue with the crashing. It's not as simple as just adding a single async test.

There are targets where we have a bunch of snapshot tests (and not many other types of test) where this isn't an issue at all. If I add a single async test to this target, it still doesn't crash. So there has to be something more. I'm going to try to find out what it is.

TechnicalDocuments/SnapshotTesting.md
nunogoncalves: Actually Took me six months to have a Pr merged where you can actually set each test folder.

mluisbrown:
Ok, I have narrowed down the problem to the async snapshot tests that we are doing in `IntroViewControllerSnapshotTests` and `PrivacyNoticeViewControllerSnapshotTests` (and possibly others). I created a sample project which repros the issue every time: https://github.com/mluisbrown/SnapshotBug

We could now adopt `SnapshotTesting` if we decide we can abandon (or somehow remove the async nature of) the tests which are causing the issue.

RuiAAPeres:
Any update on this @mluisbrown ?

mluisbrown:
@RuiAAPeres see my previous comment. 

> We could now adopt `SnapshotTesting` if we decide we can abandon (or somehow remove the async nature of) the tests which are causing the issue.

The crash situation is identified, and we can work around it by removing or re-working the small handful of async snapshot tests that we have. So, IMO if there is agreement, we can go ahead with implementing this.

mluisbrown:
> we can work around it by removing or re-working the small handful of async snapshot tests that we have

Turns out I had already done this for one of our async tests whilst investigating the issue, so we *already* have a workaround for async tests using `.asyncPullback`.

<details>
<summary>For example: (expand for source code)</summary>

```swift
final class PrivacyNoticeViewControllerSnapshotTests: SnapshotTestCase {
    let asyncWebViewSnapshotting: Snapshotting = Snapshotting<CALayer, UIImage>.image
        .asyncPullback { (window: UIWindow) in
            Async { callback in
                guard let nav = window.rootViewController as? UINavigationController,
                    let vc = nav.viewControllers.first as? PrivacyNoticeViewController
                    else {
                        fatalError()
                    }

                vc.reactive.signal(for: #selector(vc.webView(_:decidePolicyFor:decisionHandler:)))
                    .observe(on: QueueScheduler.main)
                    .take(duringLifetimeOf: vc)
                    .observeValues { _ in
                        DispatchQueue.main.asyncAfter(deadline: DispatchTime.now() + TimeInterval(0.5)) {
                            callback(window.layer)
                        }
                }
            }
        }

    override func setUp() {
        super.setUp()
        self.recordMode = false
    }

    func testViewControllerWithSummary() {
        Device.all.forEach { device in
            let viewController = PrivacyNoticeViewController(
                viewModel: PrivacyNoticeTestViewModel(),
                appConfiguration: makeAppConfiguration(),
                visuals: MockVisualDependencies.make(),
                shouldHideToolbar: false,
                shouldShowCloseButton: false,
                showSummary: true)

            let navigation = BabylonNavigationController(rootViewController: viewController)
            let window = self.makeWindow(for: navigation, device: device)

            assertSnapshot(
                matching: window,
                as: asyncWebViewSnapshotting,
                named: device.indentifier,
                record: self.recordMode
            )
        }
    }
```
</details>

So there really is *no* reason to not adopt SnapshotTesting now.

PR #18, opened 2019-01-04
Refactor Flow to be a Struct.

zzcgumn:
Proposal to refactor `Flow` from a protocol oriented design to a witness oriented design.

If you want to get a feel for how this would change our code base you can have a look at the branch. `martin/flow-struct-experiment ` in our main repository. Bear in mind that it is experimental code and needs dome tidying up to reach production quality.

Related ticket https://babylonpartners.atlassian.net/browse/CNSMR-451.

### DIFF
--- a/TechnicalDocuments/Proposals/FlowAsStruct.md
+++ b/TechnicalDocuments/Proposals/FlowAsStruct.md
@@ -1,24 +1,73 @@
-# <Title>
+# Refactor `Flow` to be a Struct
 
-* Author(s): Author1, Author2
-* Review Manager: ReviewManager
+* Author: Martin Nygren
+* Review Manager: Anders Ha
 
 ## Introduction
 
-// A short explanation about this proposal's goals.
+This proposal is inspired by the [Pointfree](https://www.pointfree.co) episodes on Witness Oriented design, [33](https://www.pointfree.co/episodes/ep33-protocol-witnesses-part-1), [34](https://www.pointfree.co/episodes/ep34-protocol-witnesses-part-2), [35](https://www.pointfree.co/episodes/ep35-advanced-protocol-witnesses-part-1), [36](https://www.pointfree.co/episodes/ep36-advanced-protocol-witnesses-part-2) and [39](https://www.pointfree.co/episodes/ep39-witness-oriented-library-design).
 
 ## Motivation
 
-// This section should the answer the `why?`.
+Currently `Flow` is defined as a `Protocol` and implemented by a couple of helper classes (e.g. `EmptyFlow`, `NavigationFlow`, `ModalFlow`) that are made available as calculated properties (`Flow.empty`, `UINavigationController.navigationFlow`, `UIViewController.modalFlow`). This implementation already indicates that modelling the concept of a `Flow` as struct with function variables is closer to what we want then defining `Flow` as a protocol. We would have preferred to extend various `UIKit` classes if modelling `Flow` as a protocol is a natural choice.
+
+```
+extension UIViewController: Flow {
+  public func present(_ viewController: UIViewController,
+                      animated: Bool,
+                      completion: (() -> Void)?) {
+                self.present(viewController,
+                             animated: animated,
+                             completion: completion)
+         }
+
+  public func replace(with viewController: UIViewController,
+                      animated: Bool) {
+                present(viewController, animated: true)
+         }
+
+  public func dismiss(animated: Bool,
+                      completion: (() -> Void)?) {
+
+                if self.presentedViewController == nil {
+                    DispatchQueue.main.async { completion?() }
+                } else {
+                    self.dismiss(animated: animated, completion: completion)
+                }
+          }
+}
+```
+
+The above example would have several problems. What if we want to have a view controller that defines a flow for embedding one or more view controllers inside subviews? How sure are we that calling the completion block is something that we always want to do?
+
+Another reason to change `Flow` from a protocol to a struct is that it makes it easier to inject side effects without impacting any business logic. One example is a requirement to post screen view events to the analytics service. Another to save user journeys to a local database to support a personalised home screen.
 
 ## Proposed solution
 
-// This section should the answer the `how?`.
+Redefine `Flow` like this.
+
+```
+/// Represents the root or containing view of a stack of view controllers.
+public struct Flow {
+
+    public let presentViewController: (UIViewController, Bool, (() -> Void)?) -> Void
+
+    public let replaceViewController: (UIViewController, Bool) -> Void
+
+    public let dismissViewController: (Bool, (() -> Void)?) -> Void
+
+    public let currentViewController: () -> UIViewController?
+
+    // Initialiser omitted.
+}
+```
+
+The variable names have been chosen to allow us to keep all existing convenience overrides. The new method `currentViewController` has been added to support side effects that need to record the current screen. It is a piece of state that every `Flow` instance must maintain to work correctly, but it has thus far never been needed by any business logic.
 
 ## Impact on existing codebase
 
-// This section should explain, assuming this proposal is accepted, how much effort it would require for it to be implemented in our codebase. Other concerns should be raised, if it's a significant deviation from our stack. 
+We would need to update the implementation of all structs and classes that implement `Flow`, there are eight of them in our code base. A couple of them (I think three or four) would need to use composition with a `Flow` instance in their public interface. With the exception of referencing a flow instance in a small number of places there would be no further changes to our code bases.
 
 ## Alternatives considered
 
-// This section describes what other approaches were considered and why this one was chosen.
+Main alternative is, of course, to keep `Flow` as a protocol. Other than being a less flexible design, I believe it would also make it a more difficult and painful to support requirements for posting analytics data. The only alternative to decorating `Flow` instances with a side effect to post screen view events that I can think of is to override, or observe calls to, `viewDidAppear`.
--- a/TechnicalDocuments/Proposals/FlowAsStruct.md
+++ b/TechnicalDocuments/Proposals/FlowAsStruct.md
@@ -1,0 +1,24 @@
+# <Title>
+
+* Author(s): Author1, Author2
+* Review Manager: ReviewManager
+
+## Introduction
+
+// A short explanation about this proposal's goals.
+
+## Motivation
+
+// This section should the answer the `why?`.
+
+## Proposed solution
+
+// This section should the answer the `how?`.
+
+## Impact on existing codebase
+
+// This section should explain, assuming this proposal is accepted, how much effort it would require for it to be implemented in our codebase. Other concerns should be raised, if it's a significant deviation from our stack. 
+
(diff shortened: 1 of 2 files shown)


TechnicalDocuments/Proposals/FlowAsStruct.md
```diff
@@ -0,0 +1,73 @@
...
+* Author: Martin Nygren
+* Review Manager: Anders Ha
+
+## Introduction
+
+This proposal is inspired by the [Pointfree](https://www.pointfree.co) episodes on Witness Oriented design, [33](https://www.pointfree.co/episodes/ep33-protocol-witnesses-part-1), [34](https://www.pointfree.co/episodes/ep34-protocol-witnesses-part-2), [35](https://www.pointfree.co/episodes/ep35-advanced-protocol-witnesses-part-1), [36](https://www.pointfree.co/episodes/ep36-advanced-protocol-witnesses-part-2) and [39](https://www.pointfree.co/episodes/ep39-witness-oriented-library-design).
+
+## Motivation
+
+Currently `Flow` is defined as a `Protocol` and implemented by a couple of helper classes (e.g. `EmptyFlow`, `NavigationFlow`, `ModalFlow`) that are made available as calculated properties (`Flow.empty`, `UINavigationController.navigationFlow`, `UIViewController.modalFlow`). This implementation already indicates that modelling the concept of a `Flow` as struct with function variables is closer to what we want then defining `Flow` as a protocol. We would have preferred to extend various `UIKit` classes if modelling `Flow` as a protocol is a natural choice.
```
andersio: > We would have preferred to extend various `UIKit` classes if modelling `Flow` as a protocol is a natural choice.

Not necessarily. There are two key reasons:

1. They hide the nature of the presenter & may also hold a weak reference to it, so as not to exhibit retain cycles. This is especially important since the presenter behind `Flow` is pretty much guaranteed to have retained the view model observed by the flow controller.

    Arguably speaking if the `Flow` protocol is constrained to `AnyObject`, flow controllers could hold a weak reference to `Flow`. But then it makes custom `Flow` implementations impossible if we universally enforce this, as they would be deallocated right away unless being retained in other places. If we don't enforce this, we are breaking the interchangeability by binding the slots with a specific implementation detail.

2. `Flow` requirements have similar names to existing UIKit signatures. Direct conformance would clutter autocompletion, especially since it covers also more than UIViewController modal presentation which guarantees mismatches of method signatures.

TechnicalDocuments/Proposals/FlowAsStruct.md
```diff
@@ -0,0 +1,73 @@
...
+                if self.presentedViewController == nil {
+                    DispatchQueue.main.async { completion?() }
+                } else {
+                    self.dismiss(animated: animated, completion: completion)
+                }
+          }
+}
+```
+
+The above example would have several problems. What if we want to have a view controller that defines a flow for embedding one or more view controllers inside subviews? How sure are we that calling the completion block is something that we always want to do?
```
andersio: I wasn't able to see how the proposed solution addresses the question here, or how it could potentially do better than what we have right now.

My answer to the question would also be that you should probably have multiple `Flow` slots, since in a majority of the circumstances involving multiple embedded VCs each of them generally has a designated semantic.

If what you wish to have is a very generalised solution for embedding content, I would be skeptical if going through this path is ideal as compared to Bento and friends.

TechnicalDocuments/Proposals/FlowAsStruct.md
```diff
@@ -0,0 +1,73 @@
...
+
+Another reason to change `Flow` from a protocol to a struct is that it makes it easier to inject side effects without impacting any business logic. One example is a requirement to post screen view events to the analytics service. Another to save user journeys to a local database to support a personalised home screen.
```
andersio: I think this is a legit motivation for the move, but the solution does not showcase any leverage of the composability over the current solution.

Also under the current framework, one could design mechanisms to inject side effects without any ergonomic impact:

```swift
struct SideEffectFlow: Flow {
    let wrapped: Flow
    let didStartPresenting: (() -> Void)?

    func present(_ viewController: UIViewController, animated: Bool, completion: (() -> Void)?) {
        wrapped.present(viewController, animated: animated, completion: completion)
        didStartPresenting?()
    }
}

extension Flow {
    public func on(didStartPresenting: (() -> Void)? = nil) {
        return SideEffectFlow(wrapped: self, didStartPresenting: didStartPresenting)
    }
}
```

TechnicalDocuments/Proposals/FlowAsStruct.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Proposed solution
+
+Redefine `Flow` like this.
+
+```
+/// Represents the root or containing view of a stack of view controllers.
+public struct Flow {
+
+    public let presentViewController: (UIViewController, Bool, (() -> Void)?) -> Void
```
andersio: The `ViewController` suffix can be dropped since the signature already conveyed the type of the subject to be presented.

TechnicalDocuments/Proposals/FlowAsStruct.md
ilyapuchka: As we [have seen](https://www.pointfree.co/episodes/ep33-protocol-witnesses-part-1), protocols are not the only way to define abstractions.

andersio: It is not the only way. I am just saying it is the blessed way by the language.

TechnicalDocuments/Proposals/FlowAsStruct.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Introduction
+
+This proposal is inspired by the [Pointfree](https://www.pointfree.co) episodes on Witness Oriented design, [33](https://www.pointfree.co/episodes/ep33-protocol-witnesses-part-1), [34](https://www.pointfree.co/episodes/ep34-protocol-witnesses-part-2), [35](https://www.pointfree.co/episodes/ep35-advanced-protocol-witnesses-part-1), [36](https://www.pointfree.co/episodes/ep36-advanced-protocol-witnesses-part-2) and [39](https://www.pointfree.co/episodes/ep39-witness-oriented-library-design).
+
+## Motivation
+
+Currently `Flow` is defined as a `Protocol` and implemented by a couple of helper classes (e.g. `EmptyFlow`, `NavigationFlow`, `ModalFlow`) that are made available as calculated properties (`Flow.empty`, `UINavigationController.navigationFlow`, `UIViewController.modalFlow`). This implementation already indicates that modelling the concept of a `Flow` as struct with function variables is closer to what we want then defining `Flow` as a protocol. We would have preferred to extend various `UIKit` classes if modelling `Flow` as a protocol is a natural choice.
+
+```
```
dmcrodrigues: Can we re-indent this snippet and mark it as being `swift`?

zzcgumn: 👍

dmcrodrigues:
I've check both the proposal and an implementation using it and in the end I'm not convinced if one is definitely better than the other because in essence we are doing exactly the same thing just differently. 

### What problem are we trying to fix?

The main problem we want to address is the need to track when something happens inside the flow, e.g. presenting a new view controller, which with the current proposal is done by decoration where we can decorate a Flow and perform our custom logic when something happens.

But as @andersio pointed out, we can do exactly the same thing by extending the current implementation and include an observable to notify about any event happening inside the flow either through decoration again or by including this capability in the flow itself.

```swift
enum FlowEvent {
    case presented(viewController: UIViewController, ...)
    case replaced(...)
    case dismissed(...)
}

public protocol Flow {
    var events: SomeKindOfObservable<FlowEvent> { get }

    func present(_ viewController: UIViewController, animated: Bool, completion: (() -> Void)?)
    func replace(with viewController: UIViewController, animated: Bool)
    func dismiss(animated: Bool, completion: (() -> Void)?)
}
```

In the end I'm not sure if there's enough to justify one approach over the other.

### Potential problems

This approach however can raise a potential problem because we cannot always disassociate the navigation from the actual screens and we can see that when we pass multiple flows like navigation and modal. 

Right now we don't enforce this through the type system, we always work with `Flow` and we declare our intentions through named parameters so consumers can be aware of that they need to provide but this can always go wrong because nothing prevents consumers to pass a navigation flow wrongly to an expected modal flow and in some sense I think we should moving to strongly types when we have this kind of constraints in terms of navigation.

Using specific types for each kind of flow enables that change while moving to a single value type `Flow` would require additional changes to be able to enforce receiving a particular flow either through tagging or something else which in my opinion reveals a disadvantage over the current approach we have.

### Conclusion

I'm not entirely sure if we really should move to a different approach or if enabling observation would not solve our main problem/limitation.

TechnicalDocuments/Proposals/FlowAsStruct.md
mr-v: > Another reason to change `Flow` from a protocol to a struct is that it makes it easier to inject side effects without impacting any business logic.

@zzcgumn First, sorry for reviewing this late (was unfamiliar with this part of the codebase). Could you show specific example that shows how it’s improved?

Decorating the `Flow` the way Anders proposes seems to match your needs and keep existing infrastructure untouched.

mr-v:
> If you want to get a feel for how this would change our code base you can have a look at the branch. martin/flow-struct-experiment in our main repository.

If anyone else also wants to take a look, here are the bits that seemed most relevant:

https://github.com/Babylonpartners/babylon-ios/blob/3027516aa4ddf6bb1e8cc6e5c9b6b743ddac70bc/Babylon/Octopus/Features/Home/HomeFlowController.swift#L60-L79

[TrackingFlow](https://github.com/Babylonpartners/babylon-ios/blob/martin/flow-struct-experiment/Babylon/Octopus/Common/Analytics/TrackingFlow.swift) - contains `screenViewTracking` that produces `Flow`.
[ScreenNaming](https://github.com/Babylonpartners/babylon-ios/blob/martin/flow-struct-experiment/BabylonUI/BabylonUI/Analytics/ScreenNaming.swift)

TechnicalDocuments/Proposals/FlowAsStruct.md
```diff
@@ -0,0 +1,73 @@
...
+/// Represents the root or containing view of a stack of view controllers.
+public struct Flow {
+
+    public let presentViewController: (UIViewController, Bool, (() -> Void)?) -> Void
+
+    public let replaceViewController: (UIViewController, Bool) -> Void
+
+    public let dismissViewController: (Bool, (() -> Void)?) -> Void
+
+    public let currentViewController: () -> UIViewController?
```
mr-v: Information like this was previously encapsulated in the internals of `Flow` protocol implementations. Exposing this opens it up for API misuse. From my understanding `Flow` is concerned with navigation and so far the currently presented `ViewController` was irrelevant. Could we achieve what you want without exposing this?

zzcgumn:
As per discussion this proposal has been rejected. We should investigate strongly typed Flow instances instead.

PR #21, opened 2019-01-11
Post Screen Events to Analytics as Flow Side Effects.

zzcgumn:
This proposal introduces a reasonably general mechanism for posting screen events when a new screen is rendered. My design goals have been to:

- High degree of code re-use.
- Easy to add screen event tracking.
- Avoid updating business logic.

Examples for how it would work in our code base can be found in the `martin/flow-struct-experiment` branch in the main repository. 

It is important to note that even though my examples are based on `Flow` instances that have been refactored as structs it is not necessary to refactor our current `Flow` implementations for this approach to work. In my view, refactoring `Flow` to be a struct makes adding side effects easier and more intuitive. I am, however, happy to add screen event tracking as side effects to our current `Flow` implementations if consensus is to not refactor `Flow`.

### DIFF
--- a/TechnicalDocuments/Proposals/TrackScreenEventsAsSideEffects.md
+++ b/TechnicalDocuments/Proposals/TrackScreenEventsAsSideEffects.md
@@ -1,28 +1,197 @@
-# <Title>
+# Track Screen Events as Flow Side Effects.
 
-* Author(s): Author1, Author2
-* Review Manager: ReviewManager
+* Author: Martin Nygren
+* Review Manager: Anders Ha
 
 ## Introduction
 
-// A short explanation about this proposal's goals.
+There is a push to improve product analytics, both in terms of the technology
+stack and coverage. To increase the coverage there is an ambition to post a
+screen view event whenever a new screen is rendered. Screen event is a somewhat
+loose term but for iOS it is a reasonable default interpretation to say that a
+screen event happens when the `view` of `UIViewController` becomes visible.
 
 ## Motivation
 
-// This section should the answer the `why?`.
+Since we will have to post a lot of screen events we want it to be easy to add
+code to track them.
+
+Posting analytics events should ideally not have any impact
+on the business logic. In an ideal world it would always be possible to collect
+product analytics data without any impact on the business or user interface
+logic, reality tends to be messier but the more analytics events we can collect
+without interfering with the business logic the better it is.
+
+It might happen that we get requirements to post some extra data in addition to
+the screen name. Although there is an ambition to keep data points consistent
+and simple it cannot be guaranteed that all of them will be.
+
+Experimenting with decorating flows with side effects I have arrived at the
+conclusion that this approach makes it quite quick to add screen event tracking.
+
+Adding screen event tracking as side-effects make sure that they don't have
+any impact on the business logic.
+
+Collecting more complicated data points can be done by decorating the flow
+differently or adding a second decoration.
+
+You can check out `martin/flow-struct-experiment` if you want to have a look at
+how this would impact our codebase. Screen event tracking has been added to the
+GP @ Hand registration journeys. Things are a bit experimental, so would need
+some cleaning up but does hopefully show how it would work.
 
 ## Proposed solution
 
-// This section should the answer the `how?`.
+Creating a general side-effect that can be used to post the screen name requires
+a mechanism for reading the screen name from the view controller. Part one of
+this proposal is thus to add define a protocol `ScreenNaming`
+
+```
+public protocol ScreenNaming {
+    func screenName() -> String?
+}
+
+```
+
+with a default implementation for `UIViewController`
+
+```
+extension UIViewController: ScreenNaming {
+    @objc open func screenName() -> String? { return view.accessibilityIdentifier }
+}
+```
+
+Using the `accessibilityIdentifier` to store the screen name has been discussed
+within the team and was considered a bit risky. I do, however, feel that in
+nearly all cases it will be fine to use the same string constant for the screen
+name and the accessibility identifier.
+
+It has been discussed that we should add a default mechanism for assigning
+accessibility identifiers to Bento components. If this happens we might not want
+to use that accessibility identifier as the screen name. In my view adding this
+override to `BabylonBoxViewController` offers a convenient way to define a
+different screen name for bento style view controllers.
+
+```
+open class BabylonBoxViewController<ViewModel, Renderer>: BoxViewController<ViewModel, Renderer, BabylonAppAppearance> {
+  ...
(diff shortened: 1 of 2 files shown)


TechnicalDocuments/Proposals/TrackScreenEventsAsSideEffects.md
```diff
@@ -0,0 +1,197 @@
...
+
+I added this helper function to make `|>` available.
+
+```swift
+static func screenViewTracking(_ flow: Flow) -> Flow {
+    return screenViewTracking(flow, with: Current.analyticsService)
+}
+```
+
+It is important to note that `Flow` instances should only be decorated at the
```
ilyapuchka: This can be error prone, if such mistake is done then it will result in double tracking. Can we maybe make it a bit safer by doing something like this:

```swift
extension UIViewController {
    struct Tracking {
       let vc: UIViewController
       var modalFlow: Flow {
          return FlowAnalyticsEvent.screenViewTracking(vc.modalFlow)
       }
    }
    var tracking: Tracking {
       return Tracking(vc: self)
    }
}
```

Then it will not be possible to decorate flow anywhere outside `BabylonUI`, for that `screenViewTracking` should be made internal. The usage will be then

```
vc.tracking.modalFlow -> flow with tracking
vc.modalFlow -> flow without tracking (if we ever need to suppress tracking in any parts of the app)
```

What do you think?

ilyapuchka:
I love how switching from protocol to value type allows us to enhance navigation like that, good idea to use it 👏

TechnicalDocuments/Proposals/TrackScreenEventsAsSideEffects.md
zzcgumn: I don't mind experimenting with designs that make it less likely that we get bugs from double posting. 

At the moment I am working on creating a mock analytics service that we can use in our UI tests and unit tests. I am hoping that this will make it inexpensive to include analytics in our automation tests.

andersio:
IMO all side effects of any given screen should be contained within its view model at best effort.

> In my view this approach is inferior for two reasons:
> - Less flexible. By decorating flows it is possible to change the tracking depending on how the screen has been reached without adding state to the view controller or view model.

We may achieve the same effect with a generalised view lifecycle observation mechanism in the VM. Such mechanism does not necessarily have to force one to react with a state transition, and it is still able to differentiate first and subsequent appearances (relevant to view controllers on navigation flows).

mluisbrown:
I have to say that I agree with @andersio's approach and his [proposal](#22) to achieve the same result.

zzcgumn:
@mluisbrown These two proposals are not exactly equivalent. For view controllers that are not based on Bento we would need to use a different mechanism to post analytics events. Presumably by overriding `viewDidAppear`.

ilyapuchka:
I don't fully agree that approach with putting everything in the view model is superior to this, though I do agree that in theory side effects should be contained in the view model. Analytics events on the other hand as @zzcgumn points ideally should not affect business logic as they just pollute it. Also relying on UIKit callbacks is relying on something we don't control and it can lead to additional boilerplate like checks if this view controller have already been displayed before when `viewWillAppear` is called if we don't want to track the same view after modal was dismissed.

I don't see how using decorated flows is less flexible either - it aims to address the issue of tracking how users navigate the app in the first place and thats what flows are being used for through out the app, whether view models are powering the business logic which IMO such analytics is not part of. 
I don't see any harm if we would track navigation one way and track user actions on views another in another way - probably it should be implemented in a way similar to how flow controllers observe routes of the view model, it should expose tracking signal that the same way as routes will be derived from the state and the tracker will send appropriate events.

I do need to go through the VM proposal more in depth though to have a better understanding of its impact and have a final conclusion of my own.

andersio:
Since the ambition is to introduce a tracking event for all screens universally, a potential alternative to doing it at `Flow` is to extend `FormViewController` and `BoxViewController` to support this particular analytics event.

In this way, we do not need to change flows at each flow injection point, and would also be able to enforce all view models to provide a name for such purpose.

andersio:
I think we need to identify the real scope of this. Apparently we cannot really have a very general solution that fits everything, or having multiple approaches co-exist without any caveat.

1. Intercepting `Flow` covers only programmatical presentation & dismissal.
    It doesn't cover interactive pop gestures, for example.

2. `viewWillAppear` and `viewWillDisappear` covers both programmatical & interactive presentation & dismissal ().

ilyapuchka:
That's a good point about interactive transitions, though I believe this can be implemented purely by the flows observing navigation controller state (I suppose that's the only interactive transition we may have in the app). My opinion on that is that neither relying on UIKit callbacks in view models, neither doing things implicitly with flows will have no trade-offs and edge cases. But flows seem to be conceptually a better fit for that as their whole purpose is navigation (and everything related to it). Otherwise view models are a dumping ground and we can as well stop using flow controllers as a separate concept and just do everything through view models, because really it's just a question of adding one method to it.

When talking about tracking we need to make clear what is the purpose of the tracking, what are we tracking, user intention to view particular screen in the app (eventually the app will do anything in response to user action in most cases) or just funnels users go through. This will help us to understand if tracking things like screen is presented when modal is dismissed or navigation stack is popped is needed or not (IMO it's not needed unless we are tracking it for building funnels).

Implementation based on flows can be also done automatically for every view controller because we are getting flows using computed properties of view controller, so we can change them to return tracking flows.

Extending Forms and Bento directly for that may be a viable solution but only if every screen in the app is implemented with them, which is not necessary a constraint we want.

I think we should wrap up the discussion as there is no perfect way of doing things and decide on what approach we go forward with (I'll be fine with any), or drop the idea entirely and do tracking on demand as it is done now.

andersio:
> Otherwise view models are a dumping ground and we can as well stop using flow controllers as a separate [...]

I think there is a misunderstanding here. While I did originally say that we might do this with the Screen Lifecycle Onservation, what I mean in  https://github.com/Babylonpartners/ios-playbook/pull/21#issuecomment-454840313 would have the tracking done in `FormVC` and `BabylonBoxVC`, and the view model is merely supplying the screen name in the same way this proposal has suggested.

> Extending Forms and Bento directly for that may be a viable solution but only if every screen in the app is implemented with them, which is not necessary a constraint we want.

A large majority of them is though, and will continue to be. I don’t think it should be written off like this. Extending Forms and Bento covers a majority of screens, and doing the rest manually is a sensible model.

zzcgumn:
@andersio Do I understand you correctly that you agree in principle to add `ScreenNaming`?

If so, I will move that to a separate proposal in order to discuss how it should be implemented.

zzcgumn:
This is no longer relevant.

PR #28, opened 2019-01-25
Code owners: Restoring developer's efficiency

dmcrodrigues:


dmcrodrigues:
> If the approval check if a pre-merge check, is it possible to re-trigger the check without updating the branch?

@Ben-Henshall I didn't understand exactly your question but this will constantly monitor every review part of the PR and change the state of the check accordingly, until someone from the team approves this check will be failing, after approving it will be passing so there's no need to re-trigger the check.

PR #29, opened 2019-01-25
Unified Definition for Screen Names.

zzcgumn:
There are several suggestions around how we should go about observing that the current has been replaced and then post a screen event. All of them rely on a unified and convenient method to define the screen name.

This proposal suggests that we define a protocol `ScreenNaming` with default implementations for `UIViewController`, `FormViewController` and `BabylonBoxViewController`.

### DIFF
--- a/TechnicalDocuments/Proposals/ScreenNaming.md
+++ b/TechnicalDocuments/Proposals/ScreenNaming.md
@@ -1,0 +1,78 @@
+# <Title>
+
+* Author(s): Martin Nygren
+* Review Manager: Sergey Shulga
+
+## Introduction
+
+There is ongoing work to improve our product analytics, both in terms of the technology stack and data coverage. We have an upcoming requirement to post a screen view event every time a new screen is shown. This splits naturally into two parts
+
+1. Defining screen names.
+2. Observing that the current screen has changed and post its screen name.
+
+This proposal concerns only the first part.
+
+## Motivation
+
+To be able to post screen events for all view controllers we will need to be able to define screen names easily. I also believe that in most cases it will not be a problem to use the same string constant for the screen and accessibility identifier. Adding accessibility identifiers would improve the stability of our UI test suite.
+
+## Proposed solution
+
+Define a protocol
+
+```swift
+	public protocol ScreenNaming {
+	    func screenName() -> String?
+	}
+
+```
+
+with default implementations for `UIViewController`, `BabylonBoxViewController` and `FormViewController`
+
+```swift
+extension UIViewController: ScreenNaming {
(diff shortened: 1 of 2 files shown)


TechnicalDocuments/Proposals/ScreenNaming.md
```diff
@@ -0,0 +1,78 @@
+# <Title>
```
daniloaliberti: define the title: 
# Unified Definition for Screen Names

zzcgumn: 👍

TechnicalDocuments/Proposals/ScreenNaming.md
```diff
@@ -0,0 +1,78 @@
...
+
+To be able to post screen events for all view controllers we will need to be able to define screen names easily. I also believe that in most cases it will not be a problem to use the same string constant for the screen and accessibility identifier. Adding accessibility identifiers would improve the stability of our UI test suite.
+
+## Proposed solution
+
+Define a protocol
+
+```swift
+public protocol ScreenNaming {
+	  func screenName() -> String?
```
daniloaliberti: [nit] 2 tabs ..

zzcgumn:
@daniloaliberti I think there are other scenarios in which we might use screen names. One example I am thinking of is unit tests for builders.

dmcrodrigues:
@zzcgumn couple of questions:
1. Can we rename `ScreenNaming` to `ScreenNamingProtocol`?
2. How are we going to handle the optionality? Because `screenName` is defined as an optional string.
3. How big is the impact if we make this explicit? Do we have accessibility identifiers in every screen? Because if not we won't be able to tell which ones are missing.

andersio:
Is there a reason why it needs to be optional? Semantically if you declare you are able to provide a screen name, you should have the screen name ready i.e. non-optional.

zzcgumn:
> We have an upcoming requirement to post a screen view event every time a new screen is shown.

This is no longer true. It has been reduced to supply screen names for a couple of selected flows. This greatly simplifies our technical solution.

1. There is no longer any real need to create a solution that allows for default implementations for `UIViewController`, `BabylonBoxViewController` and `FormsViewControllerV2`.

2. The reason it was modelled as optional was that for the foreseeable future we would have had screen names for nearly all view controllers, but needed a global mechanism to post screen events. 

3. I believe we can fulfil the current requirements with a small change to `BabylonBoxViewController` and making view models implement `ScreenNaming` (or `ScreenNamingProtocol`).

I am going to test 3 and if successful open a normal PR.

zzcgumn:
This proposal is now redundant and hence I am closing it.

PR #30, opened 2019-01-28
Babylon Design Library

gtsiap:
This proposal introduces a Design Library for the Babylon app.
The idea is to have reusable components which will be applied to the whole application.

This implementation also covers the existence of multiple themes during the runtime, like switching to dark mode.

### DIFF
--- a/TechnicalDocuments/Proposals/BabylonDesignLibrary.md
+++ b/TechnicalDocuments/Proposals/BabylonDesignLibrary.md
@@ -103,7 +103,7 @@ extension DesignLibrary {
         }
 
         fileprivate init(configuration: Configuration, traits: UITraitCollection) {
-            self.Configuration = Configuration
+            self.configuration = configuration
             self.traits = traits
         }
 
--- a/TechnicalDocuments/Proposals/BabylonDesignLibrary.md
+++ b/TechnicalDocuments/Proposals/BabylonDesignLibrary.md
@@ -151,7 +151,7 @@ extension DesignLibrary: DesignLibraryProtocol {}
 /// Note: DesignLibrary can't be used directly here due to
 /// framework dependencies.
 struct World<DesignLibrary: DesignLibraryProtocol> {
-    let DesignLibrary: MutableProperty<DesignLibrary>
+    let designLibrary: MutableProperty<DesignLibrary>
 }
 
 /// Use Current in BabylonBoxViewController
--- a/TechnicalDocuments/Proposals/BabylonDesignLibrary.md
+++ b/TechnicalDocuments/Proposals/BabylonDesignLibrary.md
@@ -48,31 +48,31 @@ Furthermore, `DesignLibrary` can also host alerts, custom viewcontroller animati
 ``` swift
 
 struct DesignLibrary {
-    let configuration: Configuration
+    let tokens: Tokens
 
     var viewControllerBuilder: ViewControllerBuilder {
-        return ViewControllerBuilder(configuration: configuration)
+        return ViewControllerBuilder(tokens: Tokens)
     }
 
     var componentsBuilder: ComponentsBuilder {
-        return ComponentsBuilder(configuration: configuration, traits: UITraitCollection())
+        return ComponentsBuilder(tokens: Tokens, traits: UITraitCollection())
     }
 
     init(
-        configuration: Configuration
+        tokens: Tokens
     ) {
-        self.configuration = configuration
+        self.tokens = tokens
     }
 }
 
 extension DesignLibrary {
-    struct Configuration {
+    struct Tokens {
         let typography: Typography
         let colors: Colors
     }
 }
 
-extension DesignLibrary.Configuration {
+extension DesignLibrary.Tokens {
     struct Typography {
         let headline1: UIFont
         let headline2: UIFont
@@ -88,13 +88,13 @@ extension DesignLibrary.Configuration {
 
 extension DesignLibrary {
     struct ComponentsBuilder: BoxAppearance {
-        let configuration: Configuration
+        let tokens: Tokens
         var traits: UITraitCollection
 
-        @available(*, deprecated, message: "Use DesignLibrary.Configuration")
+        @available(*, deprecated, message: "Use DesignLibrary.Tokens")
         public var brandColors: BrandColorsProtocol = DefaultBrandColor()
 
-        @available(*, deprecated, message: "Use DesignLibrary.Configuration")
+        @available(*, deprecated, message: "Use DesignLibrary.Tokens")
         public var appColors: AppColors = AppColors()
 
         @available(*, deprecated, message: "Use DesignLibrary.componentBuilder")
@@ -102,8 +102,8 @@ extension DesignLibrary {
             return ModernPalette(self)
         }
 
-        fileprivate init(configuration: Configuration, traits: UITraitCollection) {
-            self.configuration = configuration
+        fileprivate init(tokens: Tokens, traits: UITraitCollection) {
+            self.tokens = tokens
             self.traits = traits
         }
 
@@ -117,7 +117,7 @@ extension DesignLibrary {
 
 extension DesignLibraryService {
     struct ViewControllerBuilder {
-        let configuration: Configuration
+        let tokens: Tokens
 
         func makeConfirmationModal(
             title: String,
--- a/TechnicalDocuments/Proposals/BabylonDesignLibrary.md
+++ b/TechnicalDocuments/Proposals/BabylonDesignLibrary.md
@@ -1,7 +1,7 @@
 # Babylon Design Library
 
 * Author(s): Giorgos Tsiapaliokas
-* Review Manager: ReviewManager
+* Review Manager: Anders Ha
 
 ## Introduction
 This proposal is about introducing a Design Library.
@@ -12,22 +12,29 @@ we provide different styesheets based on the design.
 
 The idea behind this is to have reusable components with a predefined stylesheet displaying different data.
 The benefit of having a design library is
-1. to develop new UIs faster
+1. to develop new UIs faster.
 2. to catch inconsistencies in the UI due to a deisgn flaw or due to an engineering mistake.
 
 The requirement is to cover most of the common use cases and *not* to extend to edge cases.
 Also we must reuse our current tools like `Bento`
 
 ## Proposed solution
-We can introduce a new service which will be responsible for creating these components following
+We can introduce `AppearenceService` which will be responsible for creating these components following
 the semantics of the design styleguide which will be provided to us.
 
-The purpose of this service isn't to expose the full potential of `Bento` components but to follow the
-given design guidelines. This is an example.
-Also with this service we want to limit the exposure of specific components, in order to be able to introduce  
-drastic design changes in the future as easy as possible.
+The purpose of  `AppearenceService` isn't to expose the full potential of `Bento` components but to 
+follow specific design guidelines.
+As an example of this is: 
+A `Bento` component supports different behaviours like  `Deletable` the goal of this
+proposal isn't to replicate them or to make a wrapper around them but to follow specific design guidelines
+and to offer reusable UI elements.
 
-Furthermore, this service can also host alerts, custom viewcontroller animations, etc. 
+Also with `AppearenceService` we want to limit the exposure of specific components, in order to be able to introduce  
+drastic design changes in the future as easy as possible.
+For example `AppearenceService.Appearence.makeButton` returns `AnyRenderable` instead of `Component.Button`
+because in the future we may decide that we need to use a different button component. We want to depend on abstractions.
+
+Furthermore, `AppearenceService` can also host alerts, custom viewcontroller animations, etc.
 
 ``` swift
 final class AppearenceService {
--- a/TechnicalDocuments/Proposals/BabylonDesignLibrary.md
+++ b/TechnicalDocuments/Proposals/BabylonDesignLibrary.md
@@ -16,7 +16,16 @@ The benefit of having a design library is
 2. to catch inconsistencies in the UI due to a deisgn flaw or due to an engineering mistake.
 
 The requirement is to cover most of the common use cases and *not* to extend to edge cases.
-Also we must reuse our current tools like `Bento`
+Also we must reuse our current tools like `Bento`.
+
+For instance, in our codebase a common UI element is a rounded button with background color. 
+For that we use `Component.Button` and a stylesheet while this approach is functional is has a few drawbacks.
+The attributes of the button (background color, alpha, corner radius etc) can either 
+1. have been redefined in another part of the codebase which is good from a design point of view *and if* we search
+for them we can use them, *if we don't* search for them we will just reimplement something which already exists.
+2. or the attributes of the button haven't been redefined in another part of the application and we may not
+notice that so a design flaw will be implemented. We are humans after all :).
+
 
 ## Proposed solution
 We can introduce `AppearenceService` which will be responsible for creating these components following
--- a/TechnicalDocuments/Proposals/BabylonDesignLibrary.md
+++ b/TechnicalDocuments/Proposals/BabylonDesignLibrary.md
@@ -87,7 +87,7 @@ extension DesignLibrary.Configuration {
 }
 
 extension DesignLibrary {
-    struct ComponetsBuilder: BoxDesignLibrary {
+    struct ComponetsBuilder: BoxAppearance {
         let configuration: Configuration
         var traits: UITraitCollection
 
--- a/TechnicalDocuments/Proposals/BabylonDesignLibrary.md
+++ b/TechnicalDocuments/Proposals/BabylonDesignLibrary.md
@@ -54,8 +54,8 @@ struct DesignLibrary {
         return ViewControllerBuilder(configuration: configuration)
     }
 
-    var componentBuilder: ComponetsBuilder {
-        return ComponetsBuilder(configuration: configuration, traits: UITraitCollection())
+    var componentsBuilder: ComponentsBuilder {
+        return ComponentsBuilder(configuration: configuration, traits: UITraitCollection())
     }
 
     init(
@@ -87,7 +87,7 @@ extension DesignLibrary.Configuration {
 }
 
 extension DesignLibrary {
-    struct ComponetsBuilder: BoxAppearance {
+    struct ComponentsBuilder: BoxAppearance {
         let configuration: Configuration
         var traits: UITraitCollection
 
@@ -145,7 +145,7 @@ protocol DesignLibraryProtocol {
     var componentsBuilder: ComponentsBuilder { get }
 }
 
-extension DesignLibrary.ComponetsBuilder: BoxAppearance {}
+extension DesignLibrary.ComponentsBuilder: BoxAppearance {}
 extension DesignLibrary: DesignLibraryProtocol {}
 
 /// Note: DesignLibrary can't be used directly here due to
@@ -158,13 +158,13 @@ struct World<DesignLibrary: DesignLibraryProtocol> {
 class BabylonBoxViewController<ViewModel, Renderer>: BoxViewController<
     ViewModel,
     Renderer,
-    DesignLibrary.ComponetsBuilder
+    DesignLibrary.ComponentsBuilder
 > where
     ViewModel: BoxViewModel,
     Renderer: BoxRenderer,
     ViewModel.State == Renderer.State,
     ViewModel.Action == Renderer.Action,
-    Renderer.DesignLibrary == DesignLibrary.ComponetsBuilder {
+    Renderer.DesignLibrary == DesignLibrary.ComponentsBuilder {
     init(
         viewModel: ViewModel,
         renderer: Renderer.Type,
--- a/TechnicalDocuments/Proposals/BabylonDesignLibrary.md
+++ b/TechnicalDocuments/Proposals/BabylonDesignLibrary.md
@@ -1,0 +1,120 @@
+# Babylon Design Library
+
+* Author(s): Giorgos Tsiapaliokas
+* Review Manager: ReviewManager
+
+## Introduction
+This proposal is about introducing a Design Library.
+
+## Motivation
+With our current infrastrure we have have reusable components which don't reuse the same style,
+we provide different styesheets based on the design.
+
+The idea behind this is to have reusable components with a predefined stylesheet displaying different data.
+The benefit of having a design library is
+1. to develop new UIs faster
+2. to catch inconsistencies in the UI due to a deisgn flaw or due to an engineering mistake.
+
+The requirement is to cover most of the common use cases and *not* to extend to edge cases.
+Also we must reuse our current tools like `Bento`
+
+## Proposed solution
+We can introduce a new service which will be responsible for creating these components following
+the semantics of the design styleguide which will be provided to us.
+
+The purpose of this service isn't to expose the full potential of `Bento` components but to follow the
+given design guidelines. This is an example.
+Also with this service we want to limit the exposure of specific components, in order to be able to introduce  
+drastic design changes in the future as easy as possible.
+
+Furthermore, this service can also host alerts, custom viewcontroller animations, etc. 
+
+``` swift
+final class AppearenceService {
+    let appearence: MutableProperty<Appearence>
+    init(appearence: Appearence) {
+        self.appearence = MutableProperty(appearence)
+    }
+}
+
+extension AppearenceService {
+    struct Appearence {
+        let typography: Typography
+        let colors: Colors
+    }
+}
+
+extension AppearenceService.Appearence {
+    struct Typography {
+        typealias Font = (UIFont, UIColor)
+        let headline1: Font
+        let headline2: Font
+        let body1: Font
+        let body2: Font
+    }
+
+    struct Colors {
+        let primary: UIColor
+        let secondary: UIColor
+    }
+}
+
+extension AppearenceService.Appearence {
+
+     func makeButton(title: String, isEnabled: Bool, didTap: (() -> Void)?) -> AnyRenderable {
+        let buttonStyleSheet = ButtonStyleSheet()
+
+        let colorForState = isEnabled ? colors.primary : colors.primary.withAlphaComponent(0.5)
+
+        buttonStyleSheet.compose(\.backgroundColor, colorForState)
+
+        let activityIndicatorStyleSheet = ActivityIndicatorStyleSheet()
+        let styleSheet = Component.Button.StyleSheet(
+            button: buttonStyleSheet,
+            activityIndicator: activityIndicatorStyleSheet,
+            hugsContent: true,
+            autoRoundCorners: false
+        )
+
+        return Component.Button(
+            title: title,
+            isEnabled: isEnabled,
+            isLoading: false,
+            didTap: didTap,
+            styleSheet: styleSheet
+        )
+    }
+    
+    func makeLoader() -> AnyRenderable {
+        let buttonStyleSheet = ButtonStyleSheet()
+
+        let activityIndicatorStyleSheet = ActivityIndicatorStyleSheet()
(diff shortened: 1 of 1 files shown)


TechnicalDocuments/Proposals/BabylonDesignLibrary.md
```diff
@@ -0,0 +1,120 @@
...
+
+The purpose of this service isn't to expose the full potential of `Bento` components but to follow the
+given design guidelines. This is an example.
+Also with this service we want to limit the exposure of specific components, in order to be able to introduce  
+drastic design changes in the future as easy as possible.
+
+Furthermore, this service can also host alerts, custom viewcontroller animations, etc. 
+
+``` swift
+final class AppearenceService {
```
andersio: `BabylonAppAppearance` has been setup to serve a similar role `AppearenceService.Appearence`, and all `BabylonBoxViewController`s are set to observe `BabylonAppAppearance.shared`. Is there any reason why we should diverge from it?

gtsiap: I didn't mention anything regarding the integration of `AppearenceService` to our current infrastructure because in my opinion this should be a different discussion.

The current proposal has been made in order to evaluate the need of having something like `AppearenceService`.
I would like to have this conversation in another proposal or PR.

Still this is a very good question regarding why `AppearenceService` is implemented in this proposal under its current form. 
The purpose of `AppearenceService` is to provide `AppearenceService.Appearence` without using a singleton or something like `NotificationCenter`.
Since `Current` has been introduced to our codebase injecting dependencies has become easier.

Also neither `AppearenceService` nor `AppearenceService.Appearence` depend on system behaviours (like `UITraitCollection`) because their purpose is to offer a brand mechanism which won't be restricted just to `Bento` components.

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
```diff
@@ -0,0 +1,136 @@
...
+For that we use `Component.Button` and a stylesheet while this approach is functional is has a few drawbacks.
+The attributes of the button (background color, alpha, corner radius etc) can either 
+1. have been redefined in another part of the codebase which is good from a design point of view *and if* we search
+for them we can use them, *if we don't* search for them we will just reimplement something which already exists.
+2. or the attributes of the button haven't been redefined in another part of the application and we may not
+notice that so a design flaw will be implemented. We are humans after all :).
+
+
+## Proposed solution
+We can introduce `AppearenceService` which will be responsible for creating these components following
```
ilyapuchka: [nit] It's not a "service" in the meaning we put in other services. Just calling it `Appearance` would be enough.

gtsiap: I don't have strong feelings about the name.
I just picked something.

Just `Appearance ` isn't ok since its `AppearanceService. Appearance`

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
```diff
@@ -0,0 +1,136 @@
...
+
+    struct Colors {
+        let primary: UIColor
+        let secondary: UIColor
+    }
+}
+
+extension AppearenceService.Appearence {
+
+     func makeButton(title: String, isEnabled: Bool, didTap: (() -> Void)?) -> AnyRenderable {
```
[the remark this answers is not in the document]
sergdort: should we also have e.g `primaryButton`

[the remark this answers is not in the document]
gtsiap: we can have an `enum` like

```
enum Style {
    case primary
    case secondary
}

func button(style: Style, title: String, isEnabled: Bool, didTap: (() -> Void)?) -> Component.Button {
```

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
gtsiap: `AppearanceContainer` ? 
Since it will responsible for providing `AppearanceContainer. Appearance`.

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
ilyapuchka: This proposal does not touch on how appearance will be accessed by the renderers. I think it is required to move forward.

ilyapuchka: I'm also concerned that this does not touch on how this will go along with existing `BabylonAppAppearance`, injecting yet another thing in all Bento renderers does not seem like a good idea to me.

I think `BabylonAppAppearance` could be explored in a direction of adapting it. I'm particularly not a fan of `ModernPallete`, so would like to understand how it can be replaced with something (it might be just a name change + integration with colours/fonts which are currently hardcoded into it)

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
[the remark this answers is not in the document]
ilyapuchka: IMO there shouldn't be any "styles". If button looks different it's a different design system component which deserves a separate method.

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
ilyapuchka: Why do we even need such level of nesting? Why `typography` or `colours` can't be a property of `Appearance` which would be a top level type?
If we want to keep colors & fonts and separate from Bento components then there should be `Appearance` for just fonts and colors and components factory that would be created with a an instance of appearance.

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
mluisbrown: So this will mean we can only have this style of button? What if, in a given renderer, you want to modify the style? Will you have to do it the existing way?

ilyapuchka: @mluisbrown I think the idea is that you will not do it on demand but instead will introduce a new design component if the need for it is agreed by UX team or will use what is there already. So that we don't have dozens of similar but slightly different components.

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
andersio: From my view, this is more or less a replica of the appearance system that has been put in place. If you drop everything inside `BabylonAppAppearance` and put in everything introduced here, we end up with the same outcome. 

Moreover, as @ilyapuchka said, ultimately it is about how things propagate into actual screens, and we should not treat it as a separate, independent work.

On an ending note, whatever _inside_ BabylonAppAppearance can be sidelined and deprecated, and how it is populated *now* does not imply it is desirable & extensions should follow it. It is merely born as a mean to share HC and CM styles in the first place.

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
mluisbrown: @ilyapuchka tbh I don't know how well that will work with UX and Design. At the very least, we should make the designers aware that we have a fixed set of components.

ilyapuchka:
I agree, but if they think different on their side then I miss the point of such design system =/

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
gtsiap: >  we should make the designers aware that we have a fixed set of components.

They are aware of the fact that we *will* have a fixed set of components.
Also as I have been told they don't want to have stuff which doesn't exist in
the arranged fixed set of components (aka `Babylon Design Library`).

If a part of a UI can't be implemented with the `Babylon Design Library` then 
it has to be discussed.

sergdort:
I have one concern with `Appearence` returning `AnyRenderable ` is that you no longer can change the style in the `Renderer`

ilyapuchka:
@sergdort true, but I suppose the idea is that you would never need to do that, everything will be done by these factory methods, so that consistency is ensured.

gtsiap:
> I have one concern with `Appearence` returning `AnyRenderable` is that you no longer can change the style in the `Renderer`

If you are able to change the stye in the `Renderer` then we have an issue.
Let me give you an example, if we agree that there are 3 options for button and
you have to implement something which isn't one of these options, then why does it 
exist? 
We have already agreed on these 3 options.
Also we want to depend on abstractions.

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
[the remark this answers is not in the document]
zzcgumn: Is not `makeButton` consistent with Apple's guide lines for factory methods?

zzcgumn:
Intuitively I feel that is something that we should move towards. I need to have a think about adding a new entity compared to enhancing our existing.

One question, if it is a design library, why don't we call it `DesignLibrary` instead of `Appearance`?

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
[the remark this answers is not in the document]
sergdort: I kind feel like `make` prefix would add unnecessary noice ¯\_(ツ)_/¯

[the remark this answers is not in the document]
gtsiap: `beginning factory methods with make`

From: https://github.com/Babylonpartners/ios-playbook/tree/master/Style-guide

This is the main reason that I used `make` as a prefix, probably we can make an exception
for this one due to the number of methods.

gtsiap:
> One question, if it is a design library, why don't we call it DesignLibrary instead of Appearance?

Because I didn't think of it :) .
I will wait until tomorrow and then I will change the name.

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
```diff
@@ -0,0 +1,196 @@
...
+    struct Colors {
+        let primary: UIColor
+        let secondary: UIColor
+    }
+}
+
+extension DesignLibrary {
+    struct ComponetsBuilder: BoxAppearance {
+        let configuration: Configuration
+        var traits: UITraitCollection
```
andersio: Ditto, while this shall delegate to `configuration.traits` afterwards.

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
```diff
@@ -0,0 +1,196 @@
...
+
+    init(
+        configuration: Configuration
+    ) {
+        self.configuration = configuration
+    }
+}
+
+extension DesignLibrary {
+    struct Configuration {
```
andersio: It appears to me that `var traits: UITraitCollection` should be moved here, so that the config here can be updated with altered trait collections, and generate tokens _also_ based on the preferred content size category and (potentially) the user interface style (light/dark).

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
```diff
@@ -0,0 +1,196 @@
...
+
+Furthermore, `DesignLibrary` can also host alerts, custom viewcontroller animations, etc.
+
+``` swift
+
+struct DesignLibrary {
+    let configuration: Configuration
+
+    var viewControllerBuilder: ViewControllerBuilder {
+        return ViewControllerBuilder(configuration: configuration)
```
andersio: The `ViewControllerBuilder` might want to receive the whole `DesignLibrary` thing, e.g. a Bento based model.

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
andersio: and [nit] `Tokens`.

gtsiap: if `var traits: UITraitCollection` moves there then it will make `DesignLibrary`depend into `UITraitCollection` which isn't idea since we may also add a `ViewControllerBuilder` in the future.

We can have support for multiple themes by modifying `Current.designLibrary` .
No?

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
ilyapuchka: TBH I'm very skeptical about this dynamic theming conversations. One thing is to allow it to be configured at just at startup, another - to make it react just on system changes like change of content size category, and another thing is to make it completely dynamic in runtime, like switching to dark mode. Do we really need this? cc @RuiAAPeres

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    var componentBuilder: ComponetsBuilder {
```
ilyapuchka: that can be address at implementation stage I guess, but I think dropping `builder` suffix will slightly improve how this code will look like at call site, i.e. `designLibrary.component.button`, `designLibrary.viewController.bentoVC`

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
gtsiap: This was actually an example, for now we won't have something like this.
But yes this is a valid comment, in the future if we want to add something like this 
then `ViewControllerBuilder.configuration` should be replaced with `ViewControllerBuilder.designLibrary`

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
andersio: The question is more about how fonts (and maybe colors) are generated with regard to `UITraitCollection`, which has to live somewhere & has access to the current `UIContentSizeCategory`, if not through a `UITraitCollection` directly.

I don't quite understand how it is going to work with the current structure, unless you are proposing to ignore the trait collection provided by BentoKit `BoxViewController`, and just updating `Current.designLibrary` whenever the preferred content size category changed notification is posted.

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
```diff
@@ -0,0 +1,196 @@
...
+
+    struct Colors {
+        let primary: UIColor
+        let secondary: UIColor
+    }
+}
+
+extension DesignLibrary {
+    struct ComponetsBuilder: BoxAppearance {
+        let configuration: Configuration
```
ilyapuchka: again, this is implementation detail, but accessing `designLibrary.configuration.typography` is very long, having `designLibrary.typography` (via computed property) would be more ergonomic IMO.

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
andersio: > Do we really need this?

Yes, at least for the new markets AFAIK.

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
```diff
@@ -0,0 +1,196 @@
...
+    init(
+        viewModel: ViewModel,
+        renderer: Renderer.Type,
+        rendererConfig: Renderer.Config
+    ) {
+        super.init(
+            viewModel: viewModel,
+            renderer: renderer,
+            rendererConfig: rendererConfig,
+            appearance: Current.DesignLibrary.map(\.componentsBuilder)
```
ilyapuchka: Passing this as a parameter with `Current.designLibrary` as default value would have the same effect but will allow more flexibility in case we need to replace `appearance` with something fake keeping everything else in `BabylonBoxViewController`.

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
ilyapuchka: @andersio this does not answer what exactly out of three options I highlighted we need to think about. Having different styles for different partners is something we already have and it does not require any runtime updates. Or are we talking about changing how app looks like depending on what partner you register with?
Designing solution that will fit dark mode ATM is tricky I think as it's not very clear what API it will have (just a boolean in trait collections? notification? or the system colours will be inverted automatically somehow?). 
So I think we need to clarify what exactly we mean by "dynamic theming" to understand how dynamic our solution should be and to not overthink it.

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
ilyapuchka: `Tokens` is a good name 👍

TheAdamBorek:
similar to @sergdort I have bad feelings about `AnyRenderable` and the fact that we won't be able to change anything in a renderer. It's hard for me to believe that designers will use components we have without *any* change :/

Maybe we could have multiple type erasers like: `AnyButtonRenderable`, `AnyLabelRenderable`. We lose some flexibility in changing the implementation in the future however it's hard for me to imagine that we are going to change a `Button` to be a `TextField` one day.
With such a distinction we could expose stylesheet in those `AnyXXXRenderable` and renderer would have a possibility to adjust it. I'm thinking out loud here 🤔 

On the other hand, if there is a change in a component we can still use the current approach and define Stylesheets from scratches in particular renderer 🤔 


If we decide to implement a design library I think it would be nice to write snapshot tests for all those predefined components and create a `DesignLibraryComponents.md` file with a preview of how every component looks like. It will make it easier to get started (for new starters ;))

gtsiap:
> If we decide to implement a design library I think it would be nice to write snapshot tests for all those predefined components and create a `DesignLibraryComponents.md` file with a preview of how every component looks like. It will make it easier to get started (for new starters ;))

We will have snapshot tests :)
Also we can have swift playgrounds.

I believe with swift playgrounds it will be easier to maintain them.

> similar to @sergdort I have bad feelings about `AnyRenderable` and the fact that we won't be able to change anything in a renderer. It's hard for me to believe that designers will use components we have without _any_ change :/

This defeats the whole purpose. 
If you are able to apply custom stylesheets when the abstracted component changes the stylesheet may be invalid visually. 

The purpose of a style guide is to follow it :)

Yes there will be cases where we will have to deal with custom elements but these
elements should appear only *once* in the application, otherwise they should be components.

If we have an element which is slightly different than its component then this should be communicated, because probably its a design flaw.

sergdort:
What I also would like to see in this proposal is how are we going to do about the spacing between the components

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
andersio: > Having different styles for different partners is something we already have and it does not require any runtime updates

The _plan_ for the new market, AFAICR, is having one app where themes change based on the associated partner of the user. So it is going to be different from static configurations we have per white label app.

> Designing solution that will fit dark mode ATM is tricky I think as it's not very clear what API it will have

Arguably speaking the API has already landed in `UITraitCollection` as `userInterfaceStyle`, albeit not being used for iOS at the moment. Since tvOS uses it for its dark mode, as an educated guess it is very unlikely that iOS diverges from this.

andersio: Most importantly, regardless of needing backend driven themes or dark mode or both, building these on top of a design system while respecting accessibility settings are going to result in similar infrastructure given the dynamic factors involved, unless we are going after the path of "just kill the app to see your new theme".

ilyapuchka: I agree that there should be some kind of observation of trait collections that should result in UI refresh, but I think weather it should be done through design system which should be made a `Property` or wether observation should be done separately and design system should just accept current traits as a parameter can be addressed later as such integration is a separate issue from having design library on its own.

gtsiap:
> What I also would like to see in this proposal is what are we going to do about the spacing between the components

we will be provided with fixed spacing options.

I am not sure about the naming yet, we will learn more about this one soon.
We can only catch this one in the implementation.

andersio:
I guess spacing components would inevitably return, unless we allow certain components to accept extra margins to be applied on top of their own defined one.

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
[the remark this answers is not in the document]
RuiAAPeres: we have been pretty consistent with `make` across our app, I don't see why we should change that now. The method is clearly a factory method (it has a return type and is referentially transparent).

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
RuiAAPeres: Dark theme and the ability to change it at runtime is still on the table. How this is to be achieved, that's a different matter.

RuiAAPeres:
As an FP fanboy, erasing the type and not allowing further composition, or manipulation leaves a bad taste in my mouth. I might be missing something as to why we should erase the type tho.

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
```diff
@@ -0,0 +1,215 @@
...
+This proposal is about introducing a Design Library.
+
+## Motivation
+With our current infrastrure we have have reusable components which don't reuse the same style,
+we provide different styesheets based on the design.
+
+The idea behind this is to have reusable components with a predefined stylesheet displaying different data.
+The benefit of having a design library is
+1. to develop new UIs faster.
+2. to catch inconsistencies in the UI due to a deisgn flaw or due to an engineering mistake.
```
zzcgumn: [NIT] `deisgn` -> `design`

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
```diff
@@ -0,0 +1,215 @@
...
+        public var modern: ModernPalette {
+            return ModernPalette(self)
+        }
+
+        fileprivate init(tokens: Tokens, traits: UITraitCollection) {
+            self.tokens = tokens
+            self.traits = traits
+        }
+
+        func font(fontAttributes: Tokens.Typography.FontAttributes) -> UIFont {
```
zzcgumn: Guidelines for factory methods recommend `makeFont(attributes:  Tokens.Typography.FontAttributes) -> UIFont`.

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
```diff
@@ -0,0 +1,215 @@
...
+            return UIFont.preferredFont(
+                for: fontAttributes.textStyle,
+                weight: fontAttributes.weight,
+                compatibleWith: traits
+            )
+        }
+
+        func primaryButton(title: String, isEnabled: Bool, didTap: (() -> Void)?) -> AnyRenderable { ... }
```
zzcgumn: [NIT] Naming should `makePrimaryButton(title: String, isEnabled: Bool, didTap: (() -> Void)?))

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
```diff
@@ -0,0 +1,215 @@
...
+
+        func secondaryButton(title: String, isEnabled: Bool, didTap: (() -> Void)?) -> AnyRenderable { ... }
```
zzcgumn: [NIT] Naming should `makeSecondaryButton(title: String, isEnabled: Bool, didTap: (() -> Void)?))

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
```diff
@@ -0,0 +1,215 @@
...
+
+        func loader() -> AnyRenderable { ... }
```
zzcgumn: [NIT] `loader()` -> `makeLoadSpinner()`

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
zzcgumn: 👍

zzcgumn:
I have looked at this proposal and the discussion a couple of times and I think we should add `DesignLibrary` to `Current` with initialiser dependency injection. 

I do, however, have two queries about the proposed implementation.

1. I don't like the type erasure in the factory methods. I would prefer to keep the strong typing, I don't think we will ever have to worry about maintaining two completely different button types. If we do, I am not convinced that type erasure is the best solution.

2. Why would we need to make `World` a generic type? Where does the circular dependency come from and why does it require `World` to become parametrised?

andersio:
https://drive.google.com/file/d/1kffWsohKuFACR_b3TU29ecP67PKGSRKt/view?usp=sharing

mr-v:
> Also we can have swift playgrounds.
> 
> I believe with swift playgrounds it will be easier to maintain them.

Yes, please!

As a side-note, please make sure they are compiled on a CI, so they don't stop compiling after some changes. For example of setup see [Running Xcode Playgrounds on Travis CI](https://flight.school/articles/running-xcode-playgrounds-on-travis-ci/).

ilyapuchka:
Based on my personal experience separate app target will be much easier to maintain and use 🤷‍♂️ but that's details

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
```diff
@@ -0,0 +1,208 @@
...
+extension DesignLibrary {
+    struct Tokens {
+        let typography: Typography
+        let colors: Colors
+    }
+}
+
+extension DesignLibrary.Tokens {
+    struct Typography {
+        struct FontAttributes {
```
mr-v: Can we use `UIFont.TextStyle` instead of `FontAttributes`? As far as I can tell it already contains information about the font's weight.

Alternatively we may define our own `UIFont.TextStyle`, to better match our design needs.

TheAdamBorek:
Btw, one question has just come to my mind. What about icons? Assets are also a part of "a design" and we already had a conversation on it (ImageCatalogAware vs "ImageProtocol"). 

If the design library is a thing which should help us in maintaining multiple app-versions/brands, icons seems to be also a part of that

ilyapuchka:
@adampbaker good point. We can extend this concern to localisation, because it's also part of "assets" and can differ not just based on the language, but also targets. Not sure though if design library should be concerned, but agree that we should at least touch on this point on the meeting.

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
```diff
@@ -0,0 +1,208 @@
...
+because in the future we may decide that we need to use a different button component. We want to depend on abstractions.
+
+Furthermore, `DesignLibrary` can also host alerts, custom viewcontroller animations, etc.
+
+``` swift
+
+struct DesignLibrary {
+    let tokens: Tokens
+
+    var viewControllerBuilder: ViewControllerBuilder {
```
andersio: As mentioned in the call, IMO we probably shouldn't vend this as an instance property of `DesignLibrary`, which represents an a particular snapshot of the design library parameters. The relationship would be kinda weird, in the sense that we would have view controllers produced from a snapshot, that could set to observe the `Property<DesignLibrary>` in `Current`.

Particularly:
1. Bento modals being made would be observing `Current.designLibrary` updates itself, so the builder method is essentially a factory of `(Property<DesignLibrary>) -> UIViewController`.
2. Other non-Bento modals can pull out the current design library & get the necessary parameters from there, so the builder method is essentially a factory of `(Property<DesignLibrary>) -> DesignLibrary >>> (DesignLibrary) -> UIViewController`.

We may keep the modal builders namespaced under `DesignLibrary` for discoverability purpose. But `ViewControllerBuilder` should use `Current.designLibrary` or take `Property<DesignLibrary>`, and be treated like any other screen builds.

TechnicalDocuments/Proposals/BabylonDesignLibrary.md
```diff
@@ -0,0 +1,208 @@
...
+
+        func primaryButton(title: String, isEnabled: Bool, didTap: (() -> Void)?) -> AnyRenderable { ... }
+
+        func secondaryButton(title: String, isEnabled: Bool, didTap: (() -> Void)?) -> AnyRenderable { ... }
+
+        func loader() -> AnyRenderable { ... }
+    }
+}
+
+extension DesignLibraryService {
```
andersio:
```suggestion
extension DesignLibrary {
```

PR #35, opened 2019-02-10
Composable Style Sheets.

andersio:
Implementation: https://github.com/Babylonpartners/Bento/pull/88


TechnicalDocuments/Proposals/StyleSheets.md
```diff
@@ -0,0 +1,119 @@
...
+
+1. As soon as the new `StyleSheet` type is available in Bento, new components shall start using it.
+
+2. Extensions to either `UIView` or `Base*View` shall be made to provide a convenience which stores & applies style sheets on one's behalf. For example:
+
+   ```swift
+   view.backgroundColor = .green
+   view.clipsToBound = false
+
+    // NOTE: `apply` automatically keeps the inverse.
```
RuiAAPeres: What do you mean by "keeps"?

TechnicalDocuments/Proposals/StyleSheets.md
```diff
@@ -0,0 +1,119 @@
...
+
+1. Ease of Use: Requiring upfront declarations of all styleable properties.
+
+    We would have to constantly match UIKit changes. Custom views would require manual implementation.
+
+2. Mental Model: Awkwardness when being applied to views that are opened for extension.
+
+    When one defines a non generic `ExtensibleViewStyleSheet` for an `ExtensibleView`, the `StyleSheetProtocol.Element` associated type is immediately satisfied with the concrete view type. This means in a generic environment bound to `StyleSheetProtocol`, only `apply(to _: ExtensibleView)` would be recognised since `Element == ExtensibleView` per the conformance. In other words, all stylesheet subclasses must override this base class typed method & are required to downcast at runtime to access stuff defined in the subclass.
+
+    On the other hand, when one defines a generic `ExtensibleViewStyleSheet<T: ExtensibleView>` stylesheet, it does address the said issue by satisfying `StyleSheetProtocol.Element` with a generic parameter. However, it leads to awkward and verbose spellings at use sites in general, e.g. `ExtensibleViewStyleSheet<ExtensibleView>`.
```
RuiAAPeres: I don't understand if 2) is a point in favour or against. 😞

TechnicalDocuments/Proposals/StyleSheets.md
```diff
@@ -0,0 +1,119 @@
...
+
+3. Correctness: Unintended Sharing
+
+    Stylesheets are currently classes so as to be able to take advantage of inheritance, and this makes them prone to unintended sharing. With Swift offering first-class value type, this might amplify its possibility due to common perception of configurations being modelled as value types.
```
mluisbrown:
```suggestion
    Stylesheets are currently classes so as to be able to take advantage of inheritance. Since there is a common perception in Swift of configurations being modelled as value types this creates a risk of unintended sharing
```

TechnicalDocuments/Proposals/StyleSheets.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Proposed Solution
+
+A new value type `StyleSheet<View>` shall be introduced, with the following features:
+
+1. Copy on write & value semantics: Backed by CoW `Dictionary`. Support `Equatable` out of the box.
+
+2. Built around Key Paths introduced since Swift 4, and has proper support of entries setting key paths that are partially overlapping with each other.
+
+3. Support producing **inverses** for deterministic view reuse.
```
AliSoftware: I'm not a fan of the term "inverse" in that context, as in my mind that carries the idea that the function inverse in the mathematical sense is computed in some magic way. Like magically creating a `view.alpha *= 2.0` function from a `view.alpha *= 0.5` function.

I think the term "snapshot" that you used in the description is more accurate and understandable —and less confusing— to describe some state that will be _restored_ / _restorable_ before the next call to `apply` or during `prepareForReuse()`

TechnicalDocuments/Proposals/StyleSheets.md
```diff
@@ -0,0 +1,119 @@
...
+styleSheet.apply(to: &button)
+```
+
+## Impact on existing APIs
+The value type `StyleSheet` is an additive change alone.
+
+However, assuming we subsequently replace all existing BentoKit component style sheets, this shall be a massive source breaking change.
+
+### Potential Migration Path
+While script-based and staged migration is a viable solution, additional effort on top of snapshot testing might be required to verify that the removal of style sheet defaults does not lead to changes in visual appearance.
```
AliSoftware: We could suggest to use the `@available(*, renamed: ...)` attribute to facilitate the migration via fix-its?

andersio: It could be a possible to have both old stylesheets to co-exist, and then Bento components could do that. But I haven’t explored that path yet.

TechnicalDocuments/Proposals/StyleSheets.md
```diff
@@ -0,0 +1,119 @@
...
+   view.apply(
+       StyleSheet()
+            .setting(\.backgroundColor, .red)
+            .setting(\.clipsToBound, true)
+    )
+   expect(view.backgroundColor) == .red
+   expect(view.clipsToBound) == true
+
+    // NOTE: The stored inverse is applied before the new style sheet is applied.
+    //       So `clipsToBound` would be reverted to the original value.
```
inamiy: Can you update the code above to _manually_ apply inverted style that is described in https://github.com/Babylonpartners/Bento/pull/88 ?

andersio: Sure. But in any case I needed to rewrite part of the proposal. Just didn't have time to do it...

RuiAAPeres:
An outcome is needed for this proposal:

1) Close it.
2) Schedule a sync meeting.

andersio:
Meeting is scheduled!

PR #38, opened 2019-02-12
Introduce fish operator (aka Kleisli composition)

ilyapuchka:


TechnicalDocuments/Proposals/Fish_Operator.md
```diff
@@ -0,0 +1,124 @@
...
+  return { a in
+    a.flatMap(a2b)
+  }
+}
+```
+
+With this we can convert example code above to much simpler form:
+
+```swift
+didTap: ^helpText >=> Action.didTapWhatDoesThisMean >=> observer
```
sergdort: Why is there a `^` operator?

TechnicalDocuments/Proposals/Fish_Operator.md
```diff
@@ -0,0 +1,124 @@
...
+}
+
+public func flatMap<A>(_ a2b: @escaping (A) -> Void) -> (A?) -> Void {
+    return { a in
+        a.flatMap(a2b)
+    }
+}
+```
+
+This can result in ambiguty in some cases and will force to specify types explicitly, so to avoid that we can alternatively introduce new operator to work specifically with `Void` types:
```
NSMyself: Small typo here ("ambiguity", not "ambiguty")

TechnicalDocuments/Proposals/Fish_Operator.md
ilyapuchka: Because we can compose functions only, `^` lifts value `T` to function `() -> T`. It is similar to `const`, but `const` will create `(()) -> T` in this case, which is not the same as `() -> T`

mluisbrown: To lift the value into a function returning the value, so it can be composed with other functions.

TechnicalDocuments/Proposals/Fish_Operator.md
mluisbrown: So, is this the proposal? To introduce the `>->` operator only? Does this operator conflict with anything commonly used?

ilyapuchka: This is alternative solution to avoid possible ambiguities (though I think such cases will be rare so it is not needed and can be addressed when we face this issue.) I'll move this to "alternatives" section.

ilyapuchka: I mean the proposal is to introduce `>=>` as a single operator with overloads for `Void` types, but alternatively we can introduce both `>=>` and `>->` (last just for `Void` types).

dmcrodrigues: I was also confused with that was being proposed, my suggestion is to keep this proposal focused on fish operator and not introduce anything else, the rest should go into its own proposal for discussion.

ilyapuchka: Removed these details and moved this to the alternatives section

dmcrodrigues:
> This will not compile though as Swift treats `() -> A` and `(()) -> A` as different types. And even if it would treat them the same the final function that would be created by this chain will have signature `() -> Void?` which is not the same as `() -> Void`.

But @ilyapuchka aren't we trying to solve a separate issue in the operator itself? Why can't we define a function similar to `const` but that simply discards any values received and always returns `Void`? I'm a bit intrigued around this extra overloads specially considering they do not exist in Prelude which makes me think if we are not trying to solve a different issue caused by that particular example where we need to a resulting function `() -> Void` in particular.

ilyapuchka:
@dmcrodrigues if you mean to define a separate function to handle `Void` types only, then yeah, probably, but operator itself is more generic. Though with this function it will not look as clean as with operator overloads.

dmcrodrigues:
> but operator itself is more generic. Though with this function it will not look as clean as with operator overloads.

@ilyapuchka but being generic I think it shouldn't care about `Void` at all, Prelude does not include that for instance, because it seems we are tying that specific overload to a specific usage in our end where we can't use `Void?`.

`func >=> <A, B>(lhs: @escaping () -> A?, rhs: @escaping (A) -> B?) -> () -> B?`

Optionality is inherently part of the operator but we are making a specialisation to accommodate our usage, and potentially others, where maybe we should be using function composition to make it explicitly we want to "discard" everything and getting a `Void` at the end.

TechnicalDocuments/Proposals/Fish_Operator.md
```diff
@@ -0,0 +1,95 @@
...
+
+public func >=> <A, B, C>(lhs: @escaping (A) -> B?, rhs: @escaping (B) -> C?) -> (A) -> C? {
+  return lhs >>> flatMap(rhs)
+}
+
+public func flatMap<A, B>(_ a2b: @escaping (A) -> B?) -> (A?) -> B? {
+  return { a in
+    a.flatMap(a2b)
+  }
+}
```
RuiAAPeres: I am having trouble understanding this `flatMap`, since this doesn't resemble any `flatMap` I have seen. Could you point me to a reference where this is used (outside Swift for example)

ilyapuchka:
Yes, I agree, defining variants for Void is a specialization. I’ll see how we can do it with function, but at the same time I don’t see anything wrong with such specialisation 🤷‍♂️

TechnicalDocuments/Proposals/Fish_Operator.md
```diff
@@ -0,0 +1,95 @@
...
+
+public func flatMap<A>(_ a2b: @escaping (A) -> Void) -> (A?) -> Void {
+    return { a in
+        a.flatMap(a2b)
+    }
+}
+```
+
+Comparing with original implementation based on `Optional.map` this approach gives a bit different result in the sense that it always creates a closure, that inside will terminate chain of operations on first `nil` value, whether using `Optional.map` will produce `nil` value for a closure itself if mapped value is `nil`. In most cases this is an insignificant difference though, unless semantics require to pass `nil` instead of a closure that contains a terminatable chain of calls. In this case the original approach can be still used. (But personally I'd prefer us to avoid using optional closures unless it is required by API semantics, which is usually not the case)
```
RuiAAPeres: I can't understand the first sentence. 😞

TechnicalDocuments/Proposals/Fish_Operator.md
ilyapuchka: That's one defined in Prelude. We can make it private if it's an issue or use closure in-place. I don't have examples from outside Swift.

TechnicalDocuments/Proposals/Fish_Operator.md
ilyapuchka: When you use `helpText.map { ... }` the result of this will be `nil` if `helpText` is `nil`. With `>=>` there result of this will be always a non-nil closure, composed of functions that can return `nil`.

TechnicalDocuments/Proposals/Fish_Operator.md
ilyapuchka: Here is another example from Prelude of the similar function applied to any monoid (and Void is monoid as well, same as `Prelude.Unit`)

```
public func foldMap<A, M: Monoid>(_ f: @escaping (A) -> M) -> (A?) -> M {
  return { xs in
    xs.foldMap(f)
  }
}
```

ilyapuchka:
Regular composition in form of `>>>` will not work either, because the issue is not just optionality of return type or function arguments, but also that `() -> A != (()) -> A`.

Speaking of function that will make signatures match I'm having troubles coming up with something more sensible than overloads. Here is an example:

```swift
let a: () -> Int = { return 1 }
let b: (Int) -> (String) = { "\($0)" }
let c: (String) -> () = { print($0) }

let f = a >=> b >=> c
// let f: (()) -> ()?

let didTap: () -> ()

didTap = f
// cannot assign value of type '(()) -> ()?' to type '() -> ()'
```

Even if we use `Prelude.Unit` type instead of `Void` without any overloads of `>=>` we will get `(Unit) -> Unit?` as a result, which is also not useful in the same way.

There is an example of such custom handling of `Prelude.Unit` type in [swift-web](https://github.com/pointfreeco/swift-web/blob/master/Sources/ApplicativeRouter/SyntaxRouter.swift#L125). To compose two `Router` types there is this operator:

```swift
static func <%> <A, B> (lhs: Router<A>, rhs: Router<B>) -> Router<(A, B)>
```

so it effectively has signature of `(M<A>, M<B>) -> M<A, B>`

But then there is variant that discards `Unit` value on the left

```swift
func %> <A>(x: Router<Prelude.Unit>, y: Router<A>) -> Router<A>
```

or on the right:

```swift
func <% <A>(x: Route<A>, y: Router<Prelude.Unit>) -> Router<A>
```

which are `(M<Void>, M<A>) -> M<A>` and `(M<A>, M<Void>) -> M<A>` which are similar to proposed overloads of `>=>` which are 

Again if overloading the same operator is confusing we can come up with alternatives like `>->` that I also proposed.

We could workaround first issue with `(()) -> A != () -> A` by defining yet another variant of `^`, but will it be better than overload of `>=>`?
And what we can do with issue of `() -> Void? != () -> Void` without overloading or using specific operator? It's not possible to write function like `(() -> Void?) -> (() -> Void)` without involving `!` which is meh...

ilyapuchka:
@dmcrodrigues I was playing with separate function for `Void`, so we can have this:

```swift
func void(_ a: @escaping (()) -> Void?) -> () -> Void {
    return { a(()) }
}

func void<A>(_ a: @escaping (A) -> Void?) -> (A) -> Void {
    return { a($0) }
}

let f = a >=> b |> void
```

But IMO no matter what this will be called it's more confusing than just having an overload of `>=>`. And it results in the similar kind of overloads of this function itself, it's just being applied to the result, not to the composed functions.

TechnicalDocuments/Proposals/Fish_Operator.md
```diff
@@ -0,0 +1,89 @@
...
+```swift
+() -> String? >=> (String) -> Action >=> (Action) -> Void
+```
+
+This will not compile though as Swift treats `() -> A` and `(()) -> A` as different types. And even if it would treat them the same the final function that would be created by this chain will have signature `() -> Void?` which is not the same as `() -> Void`.
+
+To solve that we can introduce few overloads of `>=>` operator to handle these special cases:
+
+```swift
+public func >=> <A>(lhs: @escaping (()) -> A?, rhs: @escaping (A) -> Void) -> () -> Void {
```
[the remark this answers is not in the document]
inamiy: FYI the L57 code is the void-specialization of https://github.com/Babylonpartners/ios-playbook/pull/38#pullrequestreview-204541759 `func ifExists` with argument swapped.

ilyapuchka:
@inamiy I like your idea with concluding the chain with `ifExists`, though I wrote it differently like `ignoreOutput`, it will solve the issue with `Void?` in the end, but there is still an issue with `()` input which in real life will not allow us to pass such composition into a method that accepts just a `() -> Void`, which is how callbacks are always defined (and I guess we are not going to change it)

So instead of just doing this: 

```swift
didTap: ^helpText.map >=> Action.didTapWhatDoesThisMean >=> observer
``` 

we will need to do 

```swift
didTap: { (^helpText >=> Action.didTapWhatDoesThisMean |> ifExists(observer))(()) }

```

which defeats the purpose. We would need a second trailing operator like `ignoreInput` to solve that:

```swift
didTap: ^helpText >=> Action.didTapWhatDoesThisMean |> ifExists(observer) |> ignoreInput
```

which works but considering that we never have callbacks with `(()) -> T` signature we will need to use `ignoreInput` everywhere where we would want to apply `>=>`. Or to have a not less weird single `void` (or what ever other name is better) function that will adjust types. 
In this case I'd just prefer not to use `>=>` at all and do it the old way with opening closure and using `$0` 😢 

I agree that overloading `>=>` with special cases is probably not the best practice and it breaks the shape of Kleisli as it is defined, but considering trade-offs of doing it other ways (unless we haven't found yet a better way) I'd say its still a worth tradeoff 🤷‍♂️ The main objective of this proposal is to find a better shape for cases where `>>>` does not work.

And btw we already have an overload `func >>> <B, C>(f: @escaping () -> B, g: @escaping (B) -> C) -> () -> C`  to workaround the same issue with `(())` for `>>>` 🤷‍♂️

ilyapuchka:
Looking closer at the original issue now I wonder If the best way to write it would be `didTap: helpText.map(Action.didTapWhatDoeaThisMean >>> observer)`

inamiy:
@ilyapuchka

You are right, it's not practical to have `(()) -> T` signature everywhere in the production code 🤔 

> `didTap: ^helpText >=> Action.didTapWhatDoesThisMean |> ifExists(observer) |> ignoreInput`

I think this is the best approach so far, or we would want to ease the verbose `ignoreInput` by introducing specialized overloads as mentioned in https://github.com/Babylonpartners/ios-playbook/pull/38#pullrequestreview-204541759 . 
(NOTE: Just as a reminder, in either case, `ifExists` can't be replaced with `>=>`)

> In this case I'd just prefer not to use `>=>` at all and do it the old way with opening closure and using `$0`

I think it's still worth using `>=>` even when manually adding `ignoreInput`,
e.g. imagine 1. string to url -> 2. read file from url -> 3. decode the content, which are all failurable but want to put them into 1 single action.
`>=>` becomes more important when heavily composed, so I'm +1 to add the "canonical" fish operator.

ilyapuchka:
Yeah, `>=>` will definitely be useful by itself, and its rather easy to understand.

Btw I've found a way to write this without any overloads and new functions:

```swift
didTap: helpText.map(.didTapWhatDoesThisMean >>> observer).map(^)
```

Which so far seems to be the best option, though the trailing `map(^)` may seem weird because what it does is it converts `()?` into `(() -> ())?`.

ilyapuchka:
Yeah, I think `>=>` by itself will be useful and it's easy to understand.

The best way to rewrite the code from original issue is probably just this:

```swift
didTap: { input ?|> action >>> observer }
```

inamiy:
> `didTap: { input ?|> action >>> observer }`

Yes, I like closure-wrapping approach 👍

TechnicalDocuments/Proposals/Fish_Operator.md
RuiAAPeres: Can you please re-phrase it what you wrote?

TechnicalDocuments/Proposals/Fish_Operator.md
```diff
@@ -0,0 +1,89 @@
+# Introduce "fish" operator (Kleisli composition)
+
+* Author: Ilya Puchka
+* Review Manager:TBD
```
RuiAAPeres: Since @inamiy has been providing good feedback, it makes sense for him to be the review manager.

ilyapuchka: I'd like to add @dmcrodrigues as well, as he also was more involved.

RuiAAPeres:
**Note:** There is a lot of comments here, so my comment is regarding what I am reading (or how the current proposal is). If you have address similar comments to mine, please point me to the right place. 🙏 

---

After reading it in full, I don't completly understand how we are to compose actions going forward. The proposal states the following:

image

But immediatly after it reads `This will not compile though`. 

My next thought was: `"ok, I guess the proposal will have a way to solve this language limitation, in a ergnomic way"`:

image

While I am fan of functional aspects/concepts, the former proposed solution is difficult to read and ultimately it feels like a hack with `<| ()`. While the latter doesn't make use of the fish operator. 

What I would like to see in this proposal is a clear distinction between `how we do things` versus `how we will do things going forward`. Ultimately people (in this thread) will want to debate the `how we will do things going forward`. 

--- 

I am unsure as well about this later example. 

image

In the first example, it assumes `a` is optional, since it tries to safely unwrap with a `guard`. But on the second, (given the signature is `(A) -> B?`) it is not. There is also the mention of a `_ value: A` that is mentioned but never used. Or maybe I am missing something fundamental.

ilyapuchka:
@RuiAAPeres I have adjusted one of the examples. 
The first example was my initial reason for such suggestion, but at the end it turned out not to be the best example. Originally I've suggested to use overloads of `>=>` to work around mentioned issues of type mismatch, but I agreed with @dmcrodrigues and @inamiy comments that suggested not to do so. There are some more alternatives mentioned later in the proposal. 

It's hard to find a very good example in our massive code base, but here is one arguably better (though one can still prefer a variant with closure and `?|>`):

```swift
        let onSelectTab = ReactiveSwift.Action<BabylonTabBarItem, Void, NoError> { (item) in
            return .run {
                item
                    |> viewControllerForTabBarItem
                    ?|> presentSelectedViewController
            }
        }
```

```swift
        let onSelectTab = ReactiveSwift.Action<BabylonTabBarItem, Void, NoError> { (item) in
            let present = ^item
                >=> viewControllerForTabBarItem
                >=> presentSelectedViewController
            return .run(present |> runLazy)
        }
```

The area where this operator will be most useful is when we need to pass a closure value of `(A) -> B?` type composed of other closures of the same form as parameter to other functions. It will be an alternative to opening a closure and unwrapping optionals inside it. In this way it is similar to why we started to use `>>>` instead of closures with `$0` and `|>` inside. As it's often the case there are multiple ways of achieving the same result. Before we had `>>>` we were fine with closures with `$0` (we still use it in fact, but much less). This proposal follows the same direction. Though it's true that `>=>` will be probably less often applicable than `>>>`.  But I think the feeling and taste for it will come with time, the same way as it had with `>>>`.

inamiy:
> `didTap: ^helpText >=> Action.didTapWhatDoesThisMean >=> observer`
> `didTap: { ^helpText >=> Action.didTapWhatDoesThisMean >=> observer <| () }`
> `didTap: { helpText ?|> Action.didTapWhatDoesThisMean >>> observer }`

I recommend these code to be removed from proposal because **composing `observer` (i.e. side-effect) is not a good idea for pure functional composition**.
As I mentioned earlier, we can use function application `|>` instead  (with side-effect though) to ease this problem.
That's why I introduced `ifExists`.

For `didTap: () -> Void` and `(()) -> Void` type mismatch issue, as @ilyapuchka mentioned in https://github.com/Babylonpartners/ios-playbook/pull/38#issuecomment-464447969 , it seems `>>>` already has void-specialized overload in the project, so it's probably consistent to have void-specialized `>=>` as well, so that we don't have to add a hacky `|> ignoreInput` or `<| ()`.

So, in "Proposed solution", I think it's good to **define specialized overload on the first place** so that overall discussion becomes more simplified:

```swift
infix operator >=>: ForwardComposition
public func >=> <A, B, C>(lhs: @escaping (A) -> B?, rhs: @escaping (B) -> C?) -> (A) -> C? {
  return lhs >>> flatMap(rhs)
}

// Void-argument-specialized overload.
// This is used to bridge `(()) -> T` and `() -> T` type mismatch.
public func >=> <B, C>(lhs: @escaping (()) -> B?, rhs: @escaping (B) -> C?) -> () -> C? {
  return lhs >>> flatMap(rhs)
}

public func flatMap<A, B>(_ a2b: @escaping (A) -> B?) -> (A?) -> B? {
  return { a in
    a.flatMap(a2b)
  }
}
```

RuiAAPeres:
There is a lot to unpack here. Let's go sentence by sentence. The following was stated:

> It's hard to find a very good example in our massive code base

Isn't this almost a contradiction? We have a massive codebase, and there is a struggle to find a good use case for this operator, doesn't it lend itself to conclude that it's not a good fit? 

I have problems accepting the `runLazy` as a necessity for this operator to exist. 

> In this way it is similar to why we started to use `>>>` instead of closures with `$0` and `|>` inside. As it's often the case there are multiple ways of achieving the same result. Before we had `>>>` we were fine with closures with `$0` (we still use it in fact, but much less). This proposal follows the same direction.

I don't believe this is true. When we introduced `>>>` there were clear benefits, with plenty of examples where we could use it. The struggle to find a good spot where to use this operator, without oddities (like `runLazy`) is what makes me be against it in principle. 

I do agree with you on this:

> But I think the feeling and taste for it will come with time, the same way as it had with `>>>`.

As someone said:

> The power of these abstractions is very hard to convey to someone who has never had the privilege of learning them.

The codebase barely uses Either, so I am unsure it's the best argument in favour of this operator. 

---

I would like to clarify my stand on this proposal:

1. I am very open up for new operators and functional aspects/influences, but I am yet to be convinced of its usage in this particular problem. The proposal for the [functional composition operator](https://wiki.haskell.org/Function_composition) (`>>>`) was almost a no brainer, because its appliance and usefulness could be seen in a vast number of cases across the codebase. (it's used 295 times in our codebase). I would expect, at a smaller scale ofc, the same examples, our use-cases, for this operator. From this proposal, I have yet to see that. 
2. The example of its usage and the reason why we are introducing this operator doesn't seem to be a natural fit (either because of the Swift compiler limitations, or some other reason). To prove what I am saying, we are in a conversation reaching almost 40 comments and I have yet to see something that clearly improves the ergonomics of actions composition.

---

Looking at what @inamiy wrote and his edited-proposal:

```
infix operator >=>: ForwardComposition
public func >=> <A, B, C>(lhs: @escaping (A) -> B?, rhs: @escaping (B) -> C?) -> (A) -> C? {
  return lhs >>> flatMap(rhs)
}

// Void-argument-specialized overload.
// This is used to bridge `(()) -> T` and `() -> T` type mismatch.
public func >=> <B, C>(lhs: @escaping (()) -> B?, rhs: @escaping (B) -> C?) -> () -> C? {
  return lhs >>> flatMap(rhs)
}

public func flatMap<A, B>(_ a2b: @escaping (A) -> B?) -> (A?) -> B? {
  return { a in
    a.flatMap(a2b)
  }
}
```

Where is this going to be used? And why? If the goal is to introduce the operator for the sake of it, then I don't see much strength sustaining this initiative.

ilyapuchka:
It's hard to find good example for this operator because its application will require more refactoring than `>>>` had where the only thing we had to search for was basically usage of `$0`. Most of the time we are still thinking in terms of "application" rather than "composition" and using higher order functions so we use optional unwrapping and pass unwrapped values further in other functions rather than composing functions and passing them around. I think more of the examples of usefulness of `>=>` will be revealed if we start to think more about operating on functions rather than on values.

I'd argue that profits of using `>>>` can be also argued because we can do exactly the same with existing `|>` operator.

Regarding `runLazy` and other options it's hard to find variant which will please everyone (I'm sure `ifExists` also will not please every one), and it's only purpose is to be a workaround of type mismatch. It's also is part of "alternatives considered" so is not part of proposed change.

I'm surprised `>=>` is not seen as a natural evolution of `>>>`, as the only new thing it introduces basically is wrapping of function return values, and how it is less a no brainer.

P.S.  `<*>` is used in two places only, where in one of them it brings 0 value (I'd say it's actually -) it terms of clarity of intent, yet it exists.

RuiAAPeres:
> Most of the time we are still thinking in terms of "application" rather than "composition" and using higher order functions so we use optional unwrapping and pass unwrapped values further in other functions rather than composing functions and passing them around

The reason we introduced `>>>` is exactly why we want to move way from function application and move to function composition. 

> I'm surprised `>=>` is not seen as a natural evolution of `>>>`, as the only new thing it introduces basically is wrapping of function return values, and how it is less a no brainer.

It's less of a no-brainer because this proposal fails to show a good, clear example of its usage. Ultimately people, based on past experience, want to see a problem and how something (in this case `>=>`) solves it. *This proposable as yet to demonstrate that.* 

> P.S. `<*>` is used in two places only, where in one of them it brings 0 value (I'd say it's actually -) it terms of clarity of intent, yet it exists.

I am fine in removing it.

inamiy:
FYI here are opinions from @mbrandonw and @stephencelis (thanks!): 
- https://twitter.com/mbrandonw/status/1098250494215143425
- https://twitter.com/stephencelis/status/1098252959270559747

> But I also think it’s maybe not useful for functions of the form `() -> A?`, `(A) -> ()`, that feels to run against the grain. 

> We lift a lot of our side-effecting into IO and Kleisli compose with that: `(A) -> IO<B> >=> (B) -> IO<C>`.

Their example can be found in [pointfreeco/swift-web/Middleware.swift](https://github.com/pointfreeco/swift-web/blob/ad19b2d1090f46e6671c266ca3817395189bf6da/Sources/HttpPipeline/Middleware.swift#L113-L117) with `IO` as monad.

Also, you can find my `OptionalKleisli` example in [inamiy/SwiftRewriter/TokenHandler.swift](https://github.com/inamiy/SwiftRewriter/blob/9074a32cfee323c206cd5dd174ef8ff84e4b2522/Sources/SwiftRewriter/Utilities/TokenHandler.swift#L7-L42).
In this example, `struct OptionalKleisli` does not use `>=>` but uses `>>>` instead.
In theory, **`>=>` will become `>>>` by wrapping Kleisli arrow (`a -> m b`) with a new type**.

So... From theoretical perspective, `>=>` is strongly related to `>>>`, 
so I'm +1 to this addition even when we rarely use it.

But for iOS team best practice, I'm also feeling it might be overkill.
Please consider `f >=> g >=> h` can be rewritten as: 

1. `{ f($0).flatMap(g).flatMap(h) }` , or
2. `{ if let a = f($0), let b = g(a), let c = h(b) { return .some(c) } else { return .none } }` (in Optional context)

If we often write 1. or 2., then we should definitely ship `>=>`.
But in Optional context, we luckily have **Optional chaining in Swift** so that `$0?.f?.g?.h` often works when functions are not free but instance methods.
With the help of Optional chaining, I believe we have less chance of encountering 1. or 2. and thus less usage of Optional `>=>`.

ilyapuchka:
Regarding optional chaining, I think ATM team would probably prefer to use key paths composition rather than closure `$0` so in that case it could be probably re-written with Kleisli as something like `^\.f. >=> ^\.g >=> ^\.h` (`^` transforms `KeyPath<Root, Value>` to `(Root) -> Value`) which of course is not more compact, but does not require to open a closure and use `$0` (depending on the context this might still not fit I think).

Another common application of `>=>` I can think of is `filterMap` that accepts a closure in the form of `(A) -> B?` which is what `>=>` produces.

I agree with @RuiAAPeres in the sense that without good example it's hard to justify the proposal, but at the same time I don't see it as a blocker - we can easily remove it if we find it useless. Without it on the other hand we have less chances to discover nice applications for it. Also as we are now comfortable with `>>>` it shouldn't be an issue to understand `>=>`.

RuiAAPeres:
> I agree with @RuiAAPeres in the sense that without good example it's hard to justify the proposal, but at the same time I don't see it as a blocker - we can easily remove it if we find it useless. Without it on the other hand we have less chances to discover nice applications for it. Also as we are now comfortable with `>>>` it shouldn't be an issue to understand `>=>`.

But Ilya, as it stands it is useless. If there is a good example where we can apply this operator, please add it to the proposal? We can find a different solution for the action composition problem.

ilyapuchka:
An example for `filterMap`:

```swift
            let avatarSelection = avatarSelection.producer.filterMap { image in
                return image
                    |> PatientUpdateRequest.prepareAvatar
                    ?|> Event.userDidSelectAvatar
            }

```

becomes

```swift
            let avatarSelection = avatarSelection.producer.filterMap(
                PatientUpdateRequest.prepareAvatar >=> Event.userDidSelectAvatar
            )
```

With a help of some method that would match enum cases (subject to a separate proposal) we can rewrite

```swift
        return Feedback { state -> Signal<Event, NoError> in
            guard
                let loaded = state.loadedState,
                let selecting = loaded.selectingDeliveryMethodState,
                case let .selectingDeliveryOption(address) = selecting else {
                    return .empty
            }
```

into

```swift
        return Feedback(filterQuery:
            ^\.loadedState
                >=> ^\.selectingDeliveryMethodState
                >=> patternMatch(SelectingDeliveryMethodState.selectingDeliveryOption)
        ) { (address) -> Signal<Event, NoError> in

// patternMatch<T, A>(_ `case`: @escaping (A) -> T) -> (T) -> A?
```

Wether or not this is better and useful - up to the team to decide, I don't have anything else to add to the discussion.

inamiy:
@ilyapuchka https://github.com/Babylonpartners/ios-playbook/pull/38#issuecomment-465827342

Though second example is just a rewrite of **Swift Opitional-Binding** which is another powerful tool we should probably stick to, I like **first example that eliminates `?|>`**. 
(I think `?|>` is not a common operator in funcitonal programming)

We can say that `>>>` eliminates `|>`, so `>=>` is useful for eliminating `?|>`.
So, I'm 👍 for this change!

ilyapuchka:
💯 agree @inamiy

RuiAAPeres:
Those are good examples. Can you please add them to the proposal? Once that is done, I will gladly approve this iniative 👍

ilyapuchka:
@RuiAAPeres added

TechnicalDocuments/Proposals/Fish_Operator.md
```diff
@@ -0,0 +1,174 @@
...
+public func >=> <A, B, C>(lhs: @escaping (A) -> B?, rhs: @escaping (B) -> C?) -> (A) -> C? {
+  return lhs >>> flatMap(rhs)
+}
+
+public func >=> <A, B>(lhs: @escaping (()) -> A?, rhs: @escaping (A) -> B?) -> () -> B? {
+    let f = lhs >>> { b in b.flatMap(rhs) }
+    return { f(()) }
+}
+
+public func flatMap<A, B>(_ a2b: @escaping (A) -> B?) -> (A?) -> B? {
```
ilyapuchka: we can make this `private` or use `|>` instead =)

PR #39, opened 2019-02-12
move checkboxes to the PR template

ilyapuchka:
It occurred to me that these checkboxes may be better in PR template rather than proposal document itself.


RuiAAPeres:
They can be in both, no?

ilyapuchka:
"what's the point?" 😄

ilyapuchka:
I mean one place is enough, if in doc is fine, then fine, just thought that PR template is a better place and in doc it is a bit redundant

RuiAAPeres:
Not all PR's against this repo will be a proposal.

ilyapuchka:
Sure, but we can have multiple templates, you'll be able to select type of PR when creating it. Quite handy.

ilyapuchka:
Some more details here https://github.blog/2018-01-25-multiple-issue-and-pull-request-templates/

RuiAAPeres:
If the goal is to use PR's templates, then delete the proposal template altogether. Thanks.

ilyapuchka:
@RuiAAPeres I don't think we should do that because PR template is just for description of the PR, but what we submit in the PR is the document that needs the template.

RuiAAPeres:
@ilyapuchka you are absolutely right. 👍

PR #41, opened 2019-02-15
How to Post and Test Snowplow Analytics Events.

zzcgumn:
This is guide does, hopefully, explain how to post events to our snow plow analytics back end.

### DIFF
--- a/TechnicalDocuments/SnowplowHowTo.md
+++ b/TechnicalDocuments/SnowplowHowTo.md
@@ -6,7 +6,7 @@ Snowplow has published a schema with two frequently used event types. The first 
 
 # Posting Screen View Events from the iOS Client.
 
-View controllers that are created with `BentoKit` and use the Babylon specialisation of `BoxViewController` will post screen view events if the view model conforms to `ScreenNaming`.
+View controllers that are created with [BentoKit](https://github.com/Babylonpartners/Bento) and use the Babylon specialisation of [BoxViewController](https://github.com/Babylonpartners/Bento/blob/master/BentoKit/BentoKit/Screen/BoxViewController.swift) will post screen view events if the view model conforms to `ScreenNaming`.
 
 ```swift
 final class AwesomeRenderer: BoxRenderer {
--- a/TechnicalDocuments/SnowplowHowTo.md
+++ b/TechnicalDocuments/SnowplowHowTo.md
@@ -1,6 +1,6 @@
 # Posting Analytics Events to Snowplow
 
-Snowplow is a data centric analytics platform that combines flexibility with reasonable type safety. This is achieved by verifying incoming events against a user defined schema before storing them. The same schema definitions are used to define database tables for storing analytics events. That events are stored in well defined database tables makes it possible to match front end analytics with backend log data. The downside is that becomes slightly more complicated to post front end analytics events.
+Snowplow is a data centric analytics platform that combines flexibility with reasonable type safety. This is achieved by verifying incoming events against a user defined schema before storing them. The same schema definitions are used to define database tables for storing analytics events. These events are stored in well defined database tables making it possible to match front end analytics with backend log data. The downside is that becomes slightly more complicated to post front end analytics events.
 
 Snowplow has published a schema with two frequently used event types. The first is for screen views and the other is based on the fields in a Google Analytics event. In Snowplow lingo they are referred to as screen view event (`SPScreenView`) and structured events (`SPStructered`). Events with a bespoke definition are referred as unstructured events (`SPUnstructured`)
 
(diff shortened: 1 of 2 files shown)


TechnicalDocuments/SnowplowHowTo.md
```diff
@@ -0,0 +1,163 @@
...
+    var schema: String { return "iglu:com.thegame/achievement/jsonschema/1-0-0" }
+}
+```
+
+The generic event dictionary may also contain `label`, `property` and `value` where `label` and `property` are strings and `value` is a double. The `specificEventDictionary` dictionary must be consistent with the JSON defined by `schema`. Finally, the event type needs to be added to the dispatch table.
+
+```swift
+fileprivate func makeDispatcher() -> AnalyticsEventDispatcher {
+    let dispatcher = AnalyticsEventDispatcher()
+        .forward(ActionEvents.AchievementUnlocked.self,to: SnowplowTracker.track)
```
ilyapuchka:
```suggestion
        .forward(ActionEvents.AchievementUnlocked.self, to: SnowplowTracker.track)
```

TechnicalDocuments/SnowplowHowTo.md
```diff
@@ -0,0 +1,163 @@
...
+```swift
+class AchievementUnlockedTests: BaseFeature {
+
+    override func setUp() {
+        super.setUp()
+        self.enableMockSnowplowServer()
+    }
+
+    override func tearDown() {
+        XCTAssert(snowplowServer.verifySnowplowEvents(), "Expected analytics events were not posted")
```
ilyapuchka: I'd like to see as a part of this proposal a more generic way of validating analytics events. For example this can be done in `tearDown` of `BaseFeature`.

zzcgumn: There are pros and cons with either approach. Maybe we can look at this together when you start work on adding events for `Appointments`? It would be good to leverage your experience with the mock server.

TechnicalDocuments/SnowplowHowTo.md
```diff
@@ -0,0 +1,163 @@
...
+        super.tearDown()
+    }
+
+    let achievementSchema = "iglu:com.thegame/achievement/jsonschema/1-0-0"
+
+    func test_honor_achievement_awarded() {
+        self.snowplowServer.expectedSnowplowEvents = [
```
ilyapuchka: Same here, I think we should add analytics events as part of steps definition, not explicitly define them in tests. For example step "I see some screen" should also add analytics event to the list of expected events.

zzcgumn: This is well worth exploring. This way it is obvious which events are being tested, but adding them in the step definitions might reduce the likelihood that there is a mistake in test definitions.

I did not want to interfere with the step definitions in what is essentially a MVP implementation of test automation for analytics events. I was worried that I might lock us in to a poor design decision by without sufficient knowledge about how this patterns out in practise.

Very happy to look a this idea together with you. I think that two people reasoning about this is more likely to produce a good strategy.

TechnicalDocuments/SnowplowHowTo.md
```diff
@@ -0,0 +1,163 @@
+# Posting Analytics Events to Snowplow
+
+Snowplow is a data centric analytics platform that combines flexibility with reasonable type safety. This is achieved by verifying incoming events against a user defined schema before storing them. The same schema definitions are used to define database tables for storing analytics events. That events are stored in well defined database tables makes it possible to match front end analytics with backend log data. The downside is that becomes slightly more complicated to post front end analytics events.
```
dmcrodrigues:
```suggestion
Snowplow is a data centric analytics platform that combines flexibility with reasonable type safety. This is achieved by verifying incoming events against a user defined schema before storing them. The same schema definitions are used to define database tables for storing analytics events. These events are stored in well defined database tables making it possible to match front end analytics with backend log data. The downside is that becomes slightly more complicated to post front end analytics events.
```

TechnicalDocuments/SnowplowHowTo.md
```diff
@@ -0,0 +1,163 @@
...
+
+View controllers that are not based on `BentoKit` should post a `ScreenEvent` in `viewWillAppear`.
+
+```swift
+class MassiveViewController: UIViewController {
+...
+    override func viewWillAppear(_ animated: Bool) {
+       super.viewWillAppear(animated)
+
+        Current.analyticsService.track(ScreenEvent.viewWillAppear(screen: ScreenNames.OldSchoolFeature.massive))
```
dmcrodrigues: I understand the exercise being done here but I still think it would be better to have a ViewModel and forward the `viewWillAppear` call to it and trigger the relevant event there over having this directly in the ViewController as we don't use MVC.

zzcgumn: Working with legacy code it is always a judgment call wether to do the minimal change or make an improvement to the modelling. I would not create a `ViewModel`, but if there is a legacy one I think it is okay to add a `viewWillAppear` method it.

PR #22, opened 2019-01-14
Screen Lifecycle Observation...

andersio:
Aims to provide a generalised mechanism that can supersede several custom BoxVC implementations currently in the codebase, and a potential mean to achieve the ultimate goal of #21 i.e. emitting analytics event only when the screen first appears.

PR #44, opened 2019-02-19
Feedback based navigation

sergdort:


TechnicalDocuments/Proposals/Feedback Navigation.md
```diff
@@ -0,0 +1,81 @@
...
+
+* Author(s): Sergey Shulga
+* Review Manager: TBD
+
+## Introduction
+
+As all of you probably know in our codebase we use `Route` based navigation, where VM has a signal of routes which emits a specific route and FlowController just listens to it and performs navigation.
+
+Although the current concept is fine and we've been happy with it for a while now, it does look a bit weird when it comes to integration with RAF.
+For example, we've recently agreed that we derive all routes from the `State`  of the VM. Also, sometimes we need to pass closures to the routes as we need to notify our VM when a user selects something on the child screen.
```
ilyapuchka: Not just selects, but also cancels action. In this case parent view model should be restored to the state that was before child view was triggered. Otherwise it will be in inconsistent state and it might not be possible to make selection again (depends on the reducer implementation).

TechnicalDocuments/Proposals/Feedback Navigation.md
```diff
@@ -0,0 +1,81 @@
...
+                return .showMap(mapResponsesObserver.send(value:)) // here we are passing a closure which will push value into our VM
+            case let .showingDirections(_, place):
+                return .showDirections(place)
+            }
+        }
+```
+
+## Motivation
+
+The idea is to deal with navigation in the same way we deal with networking. Where navigation can be treated as an async operation the result of which is going to impact the current state of the VM.
```
ilyapuchka: Restating my previous comment it's important to treat it not as asynchronous only, but also cancelable and repeatable (so we most likely should not complete observers passed along to the routers)

TechnicalDocuments/Proposals/Feedback Navigation.md
```diff
@@ -0,0 +1,81 @@
...
+
+final class PharmaciesFlowController {
+    func showMap() -> SignalProducer<MapResponse, NoError> {}
+    
+    // I case where navigation does not impact current screen
+    func showSomething() -> SignalProducer<Never, NoError> {}
+}
+```
+
+Unlike other Feedbacks, we may not want the navigation to be canceled when the state changes so we can not use `Feedback` overloads which are using `flatMapLatest` under the hood. The possible solution would be to use `flatMapFirst` strategy instead, so if navigation effect is in progress it won't be canceled until it's finished and a new one wont be started.
```
ilyapuchka: > new one wont be started

I think this deserves to mention what will happen if if there is some work in parallel to navigation effect that started earlier and finishes while navigation effect in progress, will it be simply ignored or postponed until navigation effect completes? what if we need behaviour that is different to what `flatMapFirst` does (i.e. if it cancels all parallel work results, is there a way to avoid that if it's needed)

TechnicalDocuments/Proposals/Feedback Navigation.md
```diff
@@ -0,0 +1,81 @@
...
+
+Other things that we can improve with this approach is we can finally unify the way we manage presentation of a screen. Meaning that an object that is responsible for presenting a screen would be responsible for dismissing it.
```
ilyapuchka: 💯

TechnicalDocuments/Proposals/Feedback Navigation.md
```diff
@@ -0,0 +1,81 @@
...
+
+```swift
+    func showMap() -> SignalProducer<MapResponse, NoError> {
+    
+// This is not a final  implementation, it's just a hint how it may look conceptualy.
+	return SignalProducer { observer, lifetime in
+		let vc = self.builders.makeMap(response: observer, modal: modal, presenting: self.presentationFlow)
```
ilyapuchka: the side effect of this change will be that most of the time we will not need to have `presenting` parameter in builders anymore, so I'd suggest to remove it and highlight it as a difference (like "before"/"after")

TechnicalDocuments/Proposals/Feedback Navigation.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Motivation
+
+The idea is to deal with navigation in the same way we deal with networking. Where navigation can be treated as an async operation the result of which is going to impact the current state of the VM.
+
+## Proposed solution
+
+```swift    
+static func whenSearching(flowController: PharmaciesFlowController) -> Feedback<State, Event> {
+        return Feedback { state -> SignalProducer<Event, NoError> in
```
ilyapuchka: I think it will worth mentioning that this will allow a better handling on situations when some state transitions shouldn't trigger routes, for example going back in navigation state and transitioning from `state2` back to `state1` should not trigger route that is usually triggered when transitioning to `state1` from `state0`. We should be able to isolate this logic into specific feedbacks instead of doing that when `route` signal is defined. We can even add some helper methods in future to make it more convenient to declare. Same with cancelation of navigation side effect.

Btw this constructor should be using proposed `flatMapFirst` constructor, I guess.

TechnicalDocuments/Proposals/Feedback Navigation.md
ilyapuchka: If navigation is handled by a feedback loop then by its nature it will be repeatable, but we should think as well how to make cancelation convenient when implementing this approach.

TechnicalDocuments/Proposals/Feedback Navigation.md
sergdort:
```suggestion
		let vc = self.builders.makeMap(response: observer, modal: modal, presenting: self.presentationFlow)
		vc |> self.presentationFlow.present
```

TechnicalDocuments/Proposals/Feedback Navigation.md
```diff
@@ -0,0 +1,82 @@
...
+// Here is example how we can rely on the `SignalProducer` lifecycle to dismiss the screen
+                self.presentationFlow.dismiss()
+            }
+        }
+    }
+```
+
+## Impact on an existing codebase
+
+This proposal is additive so it should not impact much on the rest of the codebase. Even the same flow controller potentially could have old (with routes) and new interface (with methods that return SignalProducer)
```
RuiAAPeres: There might not be an impact in our existing codebase, but it definitely impacts, in a meaningful way, how each member will create new features.

ilyapuchka:
I'm strongly in favour of this. My main concern so far is about tight coupling of view model with flow controller here:

```
static func whenSearching(flowController: PharmaciesFlowController) -> Feedback<State, Event>
```

Even though I don't think that having different flow controllers implementations is something we need on practice, this will inevitably impact testability, as now we will need to provide a flow controller to the view model, which always requires flows and often some builders. This can be solved in multiple ways, i.e. by providing only flow controller's `handle` function to the view model instead of the whole flow controller instance, or by other means, like protocols 🤷‍♂️ 

Here are the reasons (off the top of my head) why I'm in general in favour of this proposal:

1. As was mentioned in the proposal navigation is a side effect and is asynchronous. For this we use feedbacks. From this POV it makes more sense to treat navigation the same way as networking.
2. Using separate signal for routes seems to be just an artefact of our previous approach when we derived routes from actions.
3. Removing separate signal and using feedback unifies the pattern and makes it more consistent - there will be only state, feedbacks, input and output signals.
4. Both `handle` function of flow controller and `routes` signal definition in the view model usually aggressively violate SRP principle. Breaking up at least one of them (`routes`) is a good thing (we can address that without this proposal though). If we can break up `handle` function it will be even better, but this comes with more tradeoffs (will need to replace enum with protocol and create flow controller mocks)
5. There is an unfortunate side effect of deriving routes from state, not from actions - route can be triggered multiple times as a result of irrelevant state change, which require some special handling of state changes. It may be applicable only to specific routes and may be not a problem for all of them. In this case feedbacks allow better isolation of such special handling.
6. Using signals to model results of navigation actions is more aligned with our usage of RP for example for networking (we could use closures for networking too and it would still work)

ilyapuchka:
I don't know if this should be part of this proposal or not, but I was thinking today about how we currently model navigation with routes. The questions I was asking is why routes are even defined in the ViewModel? This makes it tightly couple to the particular flow, i.e. we make it aware that on particular action or state change we need to `showSomething`. Wouldn't it be better if we remove this knowledge from the view model completely by making flow controller observe state of the view model rather then `routes` as its derivative? Then flow controller can provide view model with navigation feedbacks for it to install and so view model will not be coupled with flow controller and only flow controller will be coupled with view model, but that's something we already have.

```swift
class FlowController {
  let navigationFeedbacks: [Feedback<ViewModel.State, Event>]

  init(...) {
    navigationFeedbacks = [
       FlowController.whenSelectingSomething(...),
       FlowController.whenConfirmedSomething(...)
    ]
  }
}

class ViewModel {
   init(
     ...
     navigationFeedbacks: [Feedback<ViewModel.State, Event>]
     ...
   ) {
     self.state = Property(..., feedbacks: [
       input.feedback, ...
     ] + navigationFeedbacks)
   }
}
```

TechnicalDocuments/Proposals/Feedback Navigation.md
sergdort: > if it cancels all parallel work results, is there a way to avoid that if it's needed

`Feedback`s can not influence other feedbacks. If one uses `flatMapLatest` like most of them than state changes would cancel work in them

sergdort:
> I would like to see further definition of `"it does look a bit weird"`. The example doesn't provide enough support for the previous claim.

By being weird I mean that we treat all effects that may change state in the VM via Feedback, but navigation is modeled differently, especially when it's navigation to a child screen to pick some data for a parent.

> we've recently agreed that we derive all routes from the `State` of the VM.

This is exactly how pretty much `Feedback` loops work, we observe the particular value of the state and tiggere navigation. Which is another argument to try model it via `Feedback`

> Which seems to be the ultimate problem and the raison d'etre of this proposal, but unfortunately is not fully explored. I would like to see further examples where passing closures around is not ideal, why it's not ideal and further examples on how this can be made cleaner.

This is not ideal because we can not say `state.filterMap(^\.route)`, as we need to pass closures there. Which lead that in some parts of the app we have a different pattern for this problem. Some introduce `Presentation` enums some call them `Intent` (like me🙈) and do something like, some make presentation a part of the state enum or status e.g `case .presentingAlert(Context)`
```swift
        enum Presentation: Equatable {
            case manualEntry
            case enableSynchronization(URL)
            case alert(AlertError)
        }

// actionsObserver is a ckosure that is being passed via route
       routes = state.signal.filterMap { [actionsObserver] state in
            guard let presentation = state.presentation else { return nil }

            switch presentation {
            case .manualEntry:
                return .showManualInput(
                    state.metric,
                    completed: Action.didFinishAdding >>> Event.ui >>> actionsObserver
                )
            case let .enableSynchronization(url):
                return .showSynchronizationWebPage(
                    url,
                    completed: ^.ui(.didFinishEnablingSync) >>> actionsObserver
                )
            case let .alert(error):
                return .showAlert(
                    error,
                    completed: ^.ui(.didDismissAlert) >>> actionsObserver
                )
            }
        }
```

Also one of the issues that I see in our current navigation stack is that navigation logic is managed by the VM, but VM does not know anything about `FlowController`, which makes it very easy to forget to observer routes in the builder. That makes builder do more than one thing. Providing FC to a VM as a dependency would also allow us to unit test navigation logic, which right now can be only tested manually or via integration tests

sergdort:
@ilyapuchka 
>
class FlowController {
  let navigationFeedbacks: [Feedback<ViewModel.State, Event>]
}

Yeah, I remember I was thinking something similar when I was sketching it couple months back  😁

ilyapuchka:
> Also one of the issues that I see in our current navigation stack is that navigation logic is managed by the VM, but VM does not know anything about `FlowController`, which makes it very easy to forget to observer routes in the builder. That makes builder do more than one thing. Providing FC to a VM as a dependency would also allow us to unit test navigation logic, which right now can be only tested manually or via integration tests

I actually think that providing flow controller to the view model will make VM harder to tests and that VM should not manage navigation logic - https://github.com/Babylonpartners/ios-playbook/pull/44#issuecomment-465376470. Moving this logic completely to flow controller will also allow us to test it separately from view model, as we only will need to create a state property with feedbacks defined by the flow controller.

> Yeah, I remember I was thinking something similar when I was sketching it couple months back 😁

Why have you dropped this idea?

RuiAAPeres:
> This is exactly how pretty much Feedback loops work, we observe the particular value of the state and triggered navigation. Which is another argument to try model it via Feedback

Actually, it's not. I am not even against this idea, conceptually, I am against how it's being framed. Another example of a weak argument:

> This is not ideal because we can not say state.filterMap(^\.route), as we need to pass closures there. Which lead that in some parts of the app we have a different pattern for this problem. Some introduce Presentation enums some call them Intent (like me🙈) and do something like, some make the presentation a part of the state enum or status e.g case .presentingAlert(Context)

This is not a problem about the architecture or information flow, it's a problem around how we review code and how we do things. 

> Also one of the issues that I see in our current navigation stack is that navigation logic is managed by the VM, but VM does not know anything about FlowController, which makes it very easy to forget to observer routes in the builder.

How is this a problem? As in, if navigation doesn't happen, it's most likely because the observation wasn't set it. Is there any evidence this is a recurrent issue? 

 A couple of questions not clearly answered:

1. How will testing happen going forward? Right now, if you want to test Navigation or flow, you don't need to create any FlowController, you simply observe a ViewModel's routing. 
2. A clear example of something in the codebase where our current approach is broken and how this one would solve it. And by example, I mean a real one, used in production.   
3. There is mention of no impact, but actually, that's not true, since how architecture layers/entities will change. Somehow the FlowController will be injected into the ViewModel.

andersio:
Commenting specifically on `flatMapFirst`:

`flatMapFirst` does provide a sufficient semantic for the identified scenarios. But as a general solution, the design does not allow the flexibility to _cancel_ the effects when necessary. That is, when writing `state.flatMapFirst { effects($0) }`, there is no mean for us to cancel the outstanding effect, should the state no longer favour it.

BabylonCore has already a feedback variant that supports these scenarios, and also is general enough to support cancellation. It is `Feedback.init(firstUntilNil:effect:)`.

andersio:
The proposal also advocates for a change of relationship between VM and FC, but it does not define how testing should change, or whether (or what) method of abstractions should be used.

> @ilyapuchka 
> The questions I was asking is why routes are even defined in the ViewModel?

View Model is the brain of the screen. Routes are just a narrowed representation of screen state for the View Model to tell the external world that it expects a certain scenario to happen with some contextual data, and optionally a callback is provided.

So in this sense, indeed we could say that FlowController should get the same direct access to the entirely of state as a Renderer, and that `Route` might not really be necessary.

andersio:
> @TheAdamBorek 
> how UINavigationController & back button is going to be handled

#22 gives certain set of hooks. But for a flow implementation agnostic perspective, we would always need to have the presented screen to callback its presenter upon dismissal. This is because _some_ presentation methods (e.g. `overFullScreen` modals) would not lead to the presenter view being removed from the view hierarchy, and hence the absence of `didAppear` message.

ilyapuchka:
I think how to deal with navigation and other types of presentation edge cases is implementation detail which can be figured out when implementing the proposal, but I agree it's important to consider such edge case from the start. Speaking of navigation I used this recently and it seems to work pretty fine, at least in the context where it is being used:

```swift
private extension UIViewController {
    func onPop(_ observer: Signal<Void, NoError>.Observer) -> UIViewController {
        self.reactive
            .signal(for: #selector(UIViewController.viewWillDisappear(_:)))
            .filter { [weak self] _ in self?.isMovingFromParent == true }
            .discardValues()
            .observeValues(observer.send)
        return self
    }
}
```

andersio:
> I think how to deal with navigation and other types of presentation edge cases is implementation detail which can be figured out when implementing the proposal

It has implications on the convention of how things should be built. So while it is implementation dependent, it is useful to establish recommendations.

ilyapuchka:
@andersio 

> the design does not have a leeway around cancelling the effects when necessary

I agree that this may be needed in some scenarios, but do you have an example in mind?

> 3D Touch peek & pop

This seems to be orthogonal to the proposal, how does it stay in the way of proposed approach?

inamiy:
After some consideration, I'm feeling a bit negative to current proposal as @RuiAAPeres and @ilyapuchka mentioned:

> - The state of the VM is also modified, with potential intermediate states, when routing is hapenning.
> - There is an unfortunate side effect of deriving routes from state, not from actions

i.e. another complexity inside `State`:

```swift
extension ViewModel {
    enum State {
        case ...

        /// - Warning: 
        /// Adding routing as extra state tends to become verbose 
        /// since we often need to switch back to previous state after transition completes
        /// which is a huge pain :(
        case routing(Route)
    }
    
    /// - Note: This should be preserved as a type for readability and testability (without having feedback mocks).
    enum Route { ... }
}
```

Ideally, `Route` can be handled as an addition to `Event` instead and not affecting state itself, i.e.:

```swift
extension ViewModel {
    enum Event {
        case ui(UserAction)
        case ...
        case route(Route) // NOTE: not necessarily derived from `UserAction`
    }
}
```

which in turn requires "Route (event) to Feedback" structure instead, i.e. `(Event, State) -> Signal<Event>` as previously mentioned in https://github.com/Babylonpartners/ReactiveFeedback/pull/32#pullrequestreview-218703551 a.k.a. Mealy machine.

Without this model, I find current Moore model exploding the states even worse and hard to control.
As a proof, every viewModel's `let routes: Signal<Route, NoError>` in the codebase consists of **both states and actions (event)**, so we should provide the same functionality in `ReactiveFeedback`.

inamiy:
P.S. 
`flatMapFirst` is currently proposed in https://github.com/ReactiveCocoa/ReactiveSwift/pull/713 .
But @andersio 's https://github.com/Babylonpartners/ios-playbook/pull/44#issuecomment-469877875 cancellable version might be more preferred.

RuiAAPeres:
An outcome is needed for this proposal:

1) Close it.
2) Schedule *another* sync meeting.

ilyapuchka:
@inamiy there is a misunderstanding - I'm conceptually in favour of this change, but some implementation details need to be clarified to move forward.

Sometime ago we agreed that we are deriving routes from states rather than actions. It does create some more intermediate states, but I'd say it's a good thing rather than bad (depending on how it is implemented ofc) as we have more information about what's the state of the system is right now. And presentation state is still an essential part of any feature state - stating that it is in some "loaded" state while it is actually presenting something is not necessary correct.
If common opinion changed with a time we should discuss that of course and agree if we change this approach, then the result of such discussion will make this proposal unneeded.

inamiy:
@ilyapuchka 

> Sometime ago we agreed that we are deriving routes from states rather than actions. It does create some more intermediate states, but I'd say it's a good thing rather than bad (depending on how it is implemented ofc) as we have more information about what's the state of the system is right now.

I understand some benefit of "route as intermediate states", but compared to current `viewModel.routes.observe(flowController.handle)` approach, this state-driven routing will add more complexity without doubt since `viewModel.routes` is mostly derived from `Signal.merge(state..., action...)` (two arguments are isolated from each other and merged into simple `Route` type) but our challenge here is to combine them into a single complex `State` type (through complex reducer) and then pass it to routing feedback.

So, my current guess is "route as intermediate states" works only negatively from this simulation.
But yes, there might be some positive result from real implementation, so I will wait for it and then discuss further.

sergdort:
I think @inamiy 's proposal of having feedback `(Event, State) -> Signal<Event>` could be very handy here.  I remember I was agains deriving `Routes` from state because it's rarely gives much value, and I still think that we need to be more pragmatic about it, consider that most of the time what we do is just performing a navigation after some user action.

@andersio 
> BabylonCore has already a feedback variant that supports these scenarios*, and also is general enough to support cancellation. It is Feedback.init(firstUntilNil:effect:)

Regarding this, the reason I think `flatMapFirst` fits naturally here, is the same we have a `flatMapsLatest` by default for most feedback loops.
We use `flatMapsLatest` for most async operation because we want them to be canceled when the state changes, as probably their effects are no longer relative, and sometimes we use different strategies if we need different behaviour.

Why I think `flatMapFirst` would be the perfect candidate for navigation:
1. most of the time we do not want for it to be canceled
2. We want to prevent any other navigation to be triggered if we already presenting something

Yes someone can say that `2` may only happen when you made a mistake in the `reducer`, but the consequences of such mistake could be that the app will crush, as `UIKit` sends an exception, when we for example trying to present a new screen modally if it's already presenting something. Or we could push screen twice in the navigation stack which would just look weird for for a user.

That's why i'm advocating for to use `flatMapFirst` most of the time to model navigation and if some would have different requirements they always can use different strategy. As this is the whole point of RAF. (to be able to use different side effect strategies within your `Feedback`s based on business requirements)

ilyapuchka:
> We want to prevent any other navigation to be triggered if we already presenting something

@sergdort  Will it fit when a view model manages a flow that pushes different screens into navigation stack? I mean when it pushes screen `A` the feedback that is being active is `whenShowingA`, triggered by some state `showingA`. Then it needs to push `B` when state changes to `showingB` with feedback `whenShowingB`. Now with `flatMapFirst` strategy will the `showingB` feedback ever performed?

>  so I will wait for it and then discuss further.

@inamiy that's how all latest features are implemented lately, so there should be plenty of examples already in our codebase, where `routes = state.signal.map(...)` rather than `routes = actions.map(...)`

sergdort:
>  Will it fit when a view model manages a flow that pushes different screens into navigation stack...

@ilyapuchka If we model it via `intent` or `presentation` where it's an optional property of the `State` `flatMapFirst` will prevent it

ilyapuchka:
>  If we model it via intent

What are other options?

The particular example from our code base is prescription view model that handles the flow with 2-3 (sometimes even 4) steps, each of them being represented by a view controller pushed to the navigation stack, or presented modally. We could of course consider making view models of each of these view controllers manage parts of the flow, but it may make or may not make sense for various reasons (i.e. these view controllers are reusable in different other flows and should be tightly coupled with other view controllers in the flow).
If using `flatMapFirst` would not allow such implementation that would impose a serious constraint on the design decisions. That would be a blocker for adoption it from my point of view. 😞

diegopetrucci:
What is the current status of this PR? Should we close it for the time being or are there plans to work on it?

andersio:
@ilyapuchka 

>> 3D Touch peek & pop
> This seems to be orthogonal to the proposal, how does it stay in the way of proposed approach?

The proposal pushes for a change in relationship between Flow Controller and View Model, which 3D Touch Peek & Pop _may_ warrant a change in a totally different direction.

https://developer.apple.com/documentation/uikit/uiviewcontrollerpreviewingdelegate/1621464-previewingcontext

RuiAAPeres:
Closing this for now.

PR #47, opened 2019-02-25
Remove new hires checklists from frontpage

diegopetrucci:
It's an useful doc but it's not that important that it needs to be displayed on the frontpage.


RuiAAPeres:
Can you solve the conflicts, so we can merge this @diegopetrucci.

PR #53, opened 2019-03-07
Adds @AliSoftware and @inamiy to the team

AliSoftware:
* Given that my role will be quite across multiple squads, not sure if I should really add myself into one of the squads in the table below as well. Maybe I could add myself both in "Native/Core" and "SDK" at first?
* Not sure I get what the "Availability" column is supposed to mean, does e.g. `1/3` means that there's only one person listed in that table because it only lists people in the iOS chapter, but there are 3 people in the squad in total, it's just that the other 2 are in another chapter?

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ At Babylon, we firmly believe that **transparency** is a core value that should 
 | Sergey Shulga           | [@SergDort](https://twitter.com/SergDort)                     |
 | Viorel Mihalache        | [@viorelMO](https://twitter.com/viorelMO)                     |
 | Witold Skibniewski      |                                                               |
+| Yasuhiro Inami          | [@inamiy](https://twitter.com/inamiy)                         |
 
 
 ## 2. Which squad each individual belongs to, and its availability.
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ At Babylon, we firmly believe that **transparency** is a core value that should 
 | João Pereira            | [@NSMyself](https://twitter.com/nsmyself)                     |
 | Martin Nygren           |                                                               |
 | Michael Brown           | [@mluisbrown](https://twitter.com/mluisbrown)                 |
+| Olivier Halligon        | [@AliSoftware](https://twitter.com/aligatr)                   |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -36,6 +36,7 @@ At Babylon, we firmly believe that **transparency** is a core value that should
 | João Pereira            | [@NSMyself](https://twitter.com/nsmyself)                     |
 | Martin Nygren           |                                                               |
 | Michael Brown           | [@mluisbrown](https://twitter.com/mluisbrown)                 |
+| Olivier Halligon        | [@AliSoftware](https://twitter.com/aligatr)                   |
```
diegopetrucci: Please add @inamiy as well while we're at it. Maybe he can confirm his squad?

diegopetrucci:
> Not sure I get what the "Availability" column is supposed to mean, does e.g. 1/3 means that there's only one person listed in that table because it only lists people in the iOS chapter, but there are 3 people in the squad in total, it's just that the other 2 are in another chapter?

My understanding is that the first number is the number of devs in the squad, the second one the total number of devs that the squad should have (or will). But @RuiAAPeres can clarify.
If so, given it's not really clear, let's please add it as an explanation.

README.md
AliSoftware: @inamiy I just added you as well, could you confirm your details, and add yourself in the proper squad in the table below too?

inamiy: My role is not fixed yet, so maybe after that will be fine 😀

RuiAAPeres:
> My understanding is that the first number is the number of devs in the squad, the second one the total number of devs that the squad should have (or will

☝️

RuiAAPeres:
The 1/3 is the typical `number of people` vs `how many people we need`. I thought that was clear.

RuiAAPeres:
Also the explanation is orthogonal do this PR's goal.

mluisbrown:
@diegopetrucci the 1/3 thing has been discussed before and we agreed to leave it as it is. I personally don't think it's that clear, but that's not a reason to hold up this PR. It should be a separate PR.

PR #62, opened 2019-03-12
13. Replace Builders with Free Functions

RuiAAPeres:


TechnicalDocuments/Proposals/BuildersToFunctions.md
```diff
@@ -0,0 +1,161 @@
...
+    enum ClinicalRecords {
+        func makeMedicalHistory(with session: Session, modal: Flow) -> UIViewController
+        func makeMedicalHistory(modal: Flow) -> UIViewController
+    }
+}
+```
+
+This would remove the intricated structure we have with builders and its children.
+
+A side effect of this approach, because we are dealing with functions (versus structs) we are able to remove local state. An example of such case is `enableClinicalDownloads` at `ClinicalRecordsChildBuilders`. You can see the code smell here:
```
ilyapuchka: what are you referring to as a "code smell" here and what will not be possible with free functions? what is "local state" in this example?

TechnicalDocuments/Proposals/BuildersToFunctions.md
```diff
@@ -0,0 +1,161 @@
...
+            |> navigation.present
+```
+
+## Impact on an existing codebase
+
+As Ilya pointed out it would make unit testing a bit more cumbersome. That's the case because you no longer inject builders. I would like to make the case for the following:
+
+1. We don't test `FlowControllers` with Unit tests anyway. This is not a reason to not do it, it's simply difficult to justify time doing so, when there are more important pieces to be tested and those tests would bring, in my opinion, little value. More value would be achieved by the next point (**2.**).
+2. We can test that a particular navigation/flow is done as expected independently of this proposal being approved or not. For that to happen we simply observe a ViewModel state and assert on it, since we derive routes from the state.
+3. If we "really" want to test FlowControllers, it's still possible by mocking `Current`, which is something trivial to achieve (`World.init()`). The difference here is that we can't say we are **Unit** testing FlowControllers, but instead these become **Integration** tests. I personally don't feel this is semantically significant.
```
ilyapuchka: By mocking `Current` we will be only able to mock the network session (assuming it is there which is not the case ATM). Builder methods provide view controllers as a black box and do not provide any access to underlying view models so there is no way to act on them in tests if builder is not injected or can't be overridden in any other way. So we wouldn't be able for example to write a test that would trigger a UI action in presented view model that is supposed to dismiss its screen and then assert that flow controller actually dismisses it by calling `dismiss` on a flow.

We currently rely heavily on automation tests for testing flows, but it does not mean IMO that we should not look into unit testing flow controllers. Unit tests have important advantages over automation tests (easier to write, faster to execute), which does not mean we should drop automation in favour of unit tests either. We should use any tools at our disposal (in a suitable way ofc) to ensure quality.

TechnicalDocuments/Proposals/BuildersToFunctions.md
```diff
@@ -0,0 +1,161 @@
...
+                                secondary: modal)
+            |> navigation.present
+```
+
+## Impact on an existing codebase
+
+As Ilya pointed out it would make unit testing a bit more cumbersome. That's the case because you no longer inject builders. I would like to make the case for the following:
+
+1. We don't test `FlowControllers` with Unit tests anyway. This is not a reason to not do it, it's simply difficult to justify time doing so, when there are more important pieces to be tested and those tests would bring, in my opinion, little value. More value would be achieved by the next point (**2.**).
+2. We can test that a particular navigation/flow is done as expected independently of this proposal being approved or not. For that to happen we simply observe a ViewModel state and assert on it, since we derive routes from the state.
```
ilyapuchka: That's what we already do in some view model tests, I think, but asserting routes is not the same as testing flows, it only tests view model part, but the bug can be (and we had such cases not that long time ago) in the flow controller.
Also considering proposal to use feedbacks for navigation we will not have routes any more.

TechnicalDocuments/Proposals/BuildersToFunctions.md
```diff
@@ -0,0 +1,161 @@
...
+
+Given all those points, it has no impact, since we don't test that layer and there are currently no plans in the short term to do so. 
+
+There is still the question about moving `UserSession`, or something in that shape, to `Current`. Based on an initial conversation with Anders, and quoting him, it would be a "fairly low" effort. This is for me a point that should be thought carefully and be a reason to reject this proposal. 
+
+As a proposal on its own, it wouldn't have any immediate impact. For new features, and assuming `UserSession` in `Current` is in place, those could start using this approach. Builders that don't rely on `UserSession` could be migrated immediately. 
+
+## Alternatives considered
+
+Leave as is.
```
ilyapuchka: I don't think we should leave things as it is, this issues was brought up multiple times in the past and by ignoring it we don't achieve anything, it's good to see things moving forward with this proposal. But we can maybe explore alternative solutions which will not compromise testability.

Here is a possible alternative to consider:

```swift
extension Builder {
  struct ClinicalRecords {
    var makeMedicalHistory: (_ modal: Flow) -> UIViewController = { /* implementation here */ }
  }
}
```

We don't have to provide any dependencies to this builder because it can access `Current` or just create what is needed in the builder method implementation (at most inject only session into it), but we will be able to override this in unit test and use a view model double instead of a real one. A downside of using closures here is that we can't use named parameters on the call site.

This is similar to pointfree approach, describe [here](https://www.pointfree.co/blog/posts/21-how-to-control-the-world):

```swift
struct API {
  var setToken = { APIClient.shared.token = $0 }
  var fetchCurrentUser = APIClient.shared.fetchCurrentUser
}
```

TechnicalDocuments/Proposals/BuildersToFunctions.md
mluisbrown: Code smell:
`enableClinicalDownloads` is state local to the builder. Having it like this obscures the fact that this is actually is a flag passed in from `dependencies.appConfiguration.enableClinicalDownloads`. If the builder methods that need it just referenced `Current` it would be clear. Right now you have to find out where the builder is initialised to know how that flag is set.

TechnicalDocuments/Proposals/BuildersToFunctions.md
```diff
@@ -0,0 +1,161 @@
...
+}
+```
+
+I will provide further examples in the proposal
+
+## Motivation
+
+There are a couple of reasons for this change:
+
+1. It was decided that DI via `Current` would happen at the builders level. This means that the requirement to inject dependencies via initializer, in builders, cease to exist. This would remove: protocols, clean FlowControllers APIs and dissolve the concept of child builders.
```
mluisbrown:
```suggestion
1. It was decided that DI via `Current` would happen at the builder level. This means that the requirement to inject dependencies via the initializer, in builders, ceases to exist. This would: remove protocols, clean FlowController's APIs and dissolve the concept of child builders. 
```

TechnicalDocuments/Proposals/BuildersToFunctions.md
```diff
@@ -0,0 +1,161 @@
...
+```swift
+func make(modal: Flow) -> UIViewController
+```
+
+Given the Module where it's located it could be migrated to:
+
+```swift
+extension Builder {
+    enum ClinicalRecords {
+        func makeMedicalHistory(modal: Flow) -> UIViewController
```
mluisbrown:
```suggestion
        static func makeMedicalHistory(modal: Flow) -> UIViewController
```

TechnicalDocuments/Proposals/BuildersToFunctions.md
```diff
@@ -0,0 +1,161 @@
...
+    }
+}
+```
+
+Particular medical histories (e.g. for minors) can also be created, but in this case, the minor session would be passed instead of relying on `Current`. This situation (with child builders) can be observed at `ClinicalRecordsChildBuilders.swift`:
+
+```swift
+extension Builder {
+    enum ClinicalRecords {
+        func makeMedicalHistory(with session: Session, modal: Flow) -> UIViewController
```
mluisbrown:
```suggestion
        static func makeMedicalHistory(with session: Session, modal: Flow) -> UIViewController
```

TechnicalDocuments/Proposals/BuildersToFunctions.md
```diff
@@ -0,0 +1,161 @@
...
+        func makeMedicalHistory(modal: Flow) -> UIViewController
```
mluisbrown:
```suggestion
        static func makeMedicalHistory(modal: Flow) -> UIViewController
```

TechnicalDocuments/Proposals/BuildersToFunctions.md
```diff
@@ -0,0 +1,161 @@
...
+    }
+}
+```
+
+This would remove the intricated structure we have with builders and its children.
```
mluisbrown:
```suggestion
This would remove the intricate structure we have with builders and their children.
```

TechnicalDocuments/Proposals/BuildersToFunctions.md
mluisbrown:
```suggestion
A side effect of this approach, because we are dealing with functions (versus structs) we are able to remove local state. An example of this is `enableClinicalDownloads` in `ClinicalRecordsChildBuilders`. You can see the code smell here:
```

TechnicalDocuments/Proposals/BuildersToFunctions.md
```diff
@@ -0,0 +1,161 @@
...
+                         enableClinicalDownloads: builders.enableClinicalDownloads,
+                         addressSwitches: builders.addressSwitches)
+```
+
+With the free functions approach, this would never be possible to do, since builders don't hold state:
+
+```swift
+extension Builder {
+    enum ClinicalRecords {
+        func makeAppointmentList(session: UserSession,
```
mluisbrown:
```suggestion
        static func makeAppointmentList(session: UserSession,
```

TechnicalDocuments/Proposals/BuildersToFunctions.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+With this:
+
+```swift
+case let .appointmentList(session):
+    Builder.ClinicalRecords
+            .makeAppointmentList(with: session,
+                                enableClinicalDownloads: enableClinicalDownloads,
+                                addressSwitches: addressSwitches,
```
mluisbrown: So would `addressSwitches` and `enableClinicalDownloads` be injected into the FlowController (from the builder that created it)? I think they both come from `AppConfiguration` or a similar "global" dependency.

TechnicalDocuments/Proposals/BuildersToFunctions.md
```diff
@@ -0,0 +1,161 @@
...
+2. We can test that a particular navigation/flow is done as expected independently of this proposal being approved or not. For that to happen we simply observe a ViewModel state and assert on it, since we derive routes from the state.
+3. If we "really" want to test FlowControllers, it's still possible by mocking `Current`, which is something trivial to achieve (`World.init()`). The difference here is that we can't say we are **Unit** testing FlowControllers, but instead these become **Integration** tests. I personally don't feel this is semantically significant. 
+
+Given all those points, it has no impact, since we don't test that layer and there are currently no plans in the short term to do so. 
+
+There is still the question about moving `UserSession`, or something in that shape, to `Current`. Based on an initial conversation with Anders, and quoting him, it would be a "fairly low" effort. This is for me a point that should be thought carefully and be a reason to reject this proposal.
```
mluisbrown:
```suggestion
There is still the question about moving `UserSession`, or something in that shape, to `Current`. Based on an initial conversation with Anders, and quoting him, it would be a "fairly low" effort. This is for me a point that should be thought about carefully and could be a reason to reject this proposal. 
```

TechnicalDocuments/Proposals/BuildersToFunctions.md
mluisbrown: In your suggested approach, where would these Builder structs be created? We need to ensure that the builder methods were immutable in production code.

ilyapuchka: I think they may be created at the same places where they are created now, we just remove constructor parameters from them. Or we may create them where we need them which is in flow controllers, as default values for their arguments:

```swift
class ClinicalRecords {
  init(builder: Builder.ClinicalRecords = Builder.ClinicalRecords())
}
```

RuiAAPeres: This is a different design decision/direction and as such, if you would like to propose something else, please feel free to do so in a different place, after there's an outcome for this proposal.

ilyapuchka: I don't see how this would be better than amending this proposal.

TechnicalDocuments/Proposals/BuildersToFunctions.md
RuiAAPeres: Based on Anders suggestion, we would mock at a higher level, not at the network level. As I previously said, one thing is to have an aspiration to do something which is what your critique aims at:

> So we wouldn't be able for example to write a test that would trigger a UI action in presented view model that is supposed to dismiss its screen and then assert that flow controller actually dismisses it by calling dismiss on a flow.

You can still write your tests, integrations tests in this case, and they would be as fast as any other unit test since they don't rely on UI as well. I am not going to spend more time debating this point, simply because I already provided an alternative for testing, that doesn't rely on UI Automation.

TechnicalDocuments/Proposals/BuildersToFunctions.md
RuiAAPeres: This proposal is independent of any other proposal.

TechnicalDocuments/Proposals/BuildersToFunctions.md
RuiAAPeres: If they come globally, they would be read from `Current`.

TechnicalDocuments/Proposals/BuildersToFunctions.md
ilyapuchka: Then please provide an example of such test.

RuiAAPeres: We are not doing any sort of those tests, so there isn't even an accepted way across the team on how to test, which is precisely my point. In any case:


```swift
world = World()
world.userSession = MockedUserSession() 

var isCalled = false
let mockedFlow = { isCalled = true }
let shouldntBeCalled = { fatalError() }

let flowController = AFlowController(navigation: mockedFlow, model: shouldntBeCalled)
flowController.handle(.aState)

XCTAssert(isCalled == true)
```

This would test that the right flow is presented.

TechnicalDocuments/Proposals/BuildersToFunctions.md
RuiAAPeres: I am not vouching for that change since it goes in a different direction: it still uses builders as entities that you pass around.

TechnicalDocuments/Proposals/BuildersToFunctions.md
ilyapuchka: My concern is not about a `handle` method that is being tested in this example, but about testing how flow controller reacts on UI actions from presented view models.

TechnicalDocuments/Proposals/BuildersToFunctions.md
```diff
@@ -0,0 +1,161 @@
...
+2. It would become easier to find factory methods to construct screens. One could just type `Builder.` and Xcode would autocomplete with the available factory methods. It would also free engineers from wasting too much time finding dependencies if they are immediately accessible via `Current` and the namespaced free functions. This is not a justification for engineers to not understand the codebase and how to do things "properly". It is simply an observation of how cumbersome passing dependencies via initializer is, the raison d'être for `Current` and the overall way of creating screens (VM + VC + FC). 
+
+The first point is about simplicity and clean-up of multiple ways of creating screens, while the second point focuses on discoverability and development ergonomics.   
+
+## Proposed solution
+
+**Note:** The proposed solution assumes that a way to authenticate HTTP requests is available at `Current` level (in this case something in the shape of `MainUserSession` abstracted via `UserSession`).
```
ilyapuchka: Is it a part of this proposal to move `MainUserSession` or any other network related instances into `Current`? How can we discuss an approach that is based on something that is not in place and that we previously explicitly agreed not to do? I'd suggest to first discuss requirements for this approach to be possible.

TechnicalDocuments/Proposals/BuildersToFunctions.md
RuiAAPeres: It feels that you are talking about integration tests, where you assert an entire flow from a ViewModel action down to a flowController "reaction". Assuming @sergdort 's proposal would be accepted (again this is a weak argument from my side, because it's a big assumption), you could just inject mocked FlowControllers Feedback into the VM and assert on those.

TechnicalDocuments/Proposals/BuildersToFunctions.md
RuiAAPeres: It's called an assumption.

TechnicalDocuments/Proposals/BuildersToFunctions.md
ilyapuchka: I'm talking about unit testing flow controller in isolation of their environment like real builders and other real view models. Injecting mocked feedback is something one would do in unit tests of view model, not flow controller which would define such feedback. This feedback is something that should be tested (as we test other feedbacks, why wouldn't we test some of them)

TechnicalDocuments/Proposals/BuildersToFunctions.md
```diff
@@ -0,0 +1,161 @@
...
+        static func makeMedicalHistory(with session: Session, modal: Flow) -> UIViewController
+        static func makeMedicalHistory(modal: Flow) -> UIViewController
+    }
+}
+```
+
+This would remove the intricate structure we have with builders and their children.
+
+A side effect of this approach, because we are dealing with functions (versus structs) we are able to remove local state. An example of this is `enableClinicalDownloads` in `ClinicalRecordsChildBuilders`. You can see the code smell here:
+
```
daniloaliberti: `enableClinicalDownloads` is passed through a long chain until appointment detail. 
A clear example of refactor to start using the convenience of Current. 
Also it seats in appConfiguration and that's the big rock what I'm moving to Current atm. 👍

RuiAAPeres:
> Proposed solution is based on the assumption that is not in place (having user session related things in the Current) and that was discussed before and team agreed not to do so.

If this proposal is accepted, nothing will change until the requirement is in place. 


> and that was discussed before and team agreed not to do so.

Discussions and decisions shouldn't be set in stone when new facts, different requirements and problems surface. Assuming that we have decided in the past is the best solution possible, is a hubris problem. 

> if consensus is that we don't care about unit testing flow controllers.

Consensus is not that that testing is important. The problem is:

1. Even if this proposal is accepted, testing FlowControllers is still possible. ✅ 
2. We are not currently testing any FlowControllers anyway, so there isn't any breaking change, or change in our ways of testing. ✅ 
3. Testing is very important for us as a team as usual. 👍

RuiAAPeres:
@ilyapuchka I will count your vote as a -1. If the majority of people are against this proposal I shall close it, otherwise, we will move forward. Thank you for your thorough review. 👍

TechnicalDocuments/Proposals/BuildersToFunctions.md
RuiAAPeres: Not sure what you want to achieve with testing FlowControllers in isolation. to make sure you get the right method being called in a builder? Is that it?

TechnicalDocuments/Proposals/BuildersToFunctions.md
ilyapuchka: > it still uses builders as entities that you pass around.
this is not needed with default initialiser parameters. Approach we already use in `Current`.

TechnicalDocuments/Proposals/BuildersToFunctions.md
RuiAAPeres: That can be tested by passing the dismiss state to the `handler` and asserting on the flow.

ilyapuchka: This would be a state of a presented view model, where and how it will be passed to the flow controller that presented it?  There will be no access to this view model from the test.

RuiAAPeres: You can pass via a handler as part of the state, that would be called once the presented screen is dismissed. For example `.showScreen(someData, dismissHandler)`

TechnicalDocuments/Proposals/BuildersToFunctions.md
```diff
@@ -0,0 +1,161 @@
...
+```swift
+func make(modal: Flow) -> UIViewController
+```
+
+Given the Module where it's located it could be migrated to:
+
+```swift
+extension Builder {
+    enum ClinicalRecords {
+        static func makeMedicalHistory(modal: Flow) -> UIViewController
```
sergdort: I wander if it make to have them as `var` closures? So you would be able stub them if we start to unit test `FlowControllers`.

TechnicalDocuments/Proposals/BuildersToFunctions.md
ilyapuchka: How this would help to test that `flow.dismiss` is called in the following implementation?

```swift
func handle(_ route: Route) {
  switch route {
    case let .showScreen(someData, dismissHandler):
       Builder.Something.makeSomething(someData, onDismiss: { result in
          flow.dismiss() { dismissHandler(result) }
       } |> flow.present
  }
}
```

TechnicalDocuments/Proposals/BuildersToFunctions.md
RuiAAPeres: How would you make the default value to use `Current`?

TechnicalDocuments/Proposals/BuildersToFunctions.md
RuiAAPeres: You assert that `dismissHandler` is called. 🤷‍♂️

ilyapuchka: the fact that `dismissHandler` has little to do with ensuring that `dismiss` (or any other method, it can be `popToRoot` or something) is called on the proper flow (flow most likely will be a mock implementation as well). Also one could write it as

```
flow.dismiss()
dismissHandler()
```

This may be completely fine implementation in some cases, but not all. So here `dismissHandler` will be caller regardless of `flow.dismiss` being called at all or being called on a correct flow instance.

sergdort: From `FC` perspective you usually what to unit test that providing a `route` it calls correct method on builder and correct `flow`. I not sure what else we would be interested in?

RuiAAPeres: The builder method, with the current proposal, wouldn't be possible in a straightforward way. The flow calling would be.

RuiAAPeres: @ilyapuchka quoting yourself:

> I'd be fine dismissing my concern about testability

I think we already agreed that:

1. We are not testing those particular scenarios, currently, independently of this proposal
2. We have no plans to do so in the short/medium term.
3. Critical paths of the App are already covered with UI Automation.

I am unsure why we keep going around this point.

dmcrodrigues: The example shared shows how to test FC if/when needed, the SUT needs to be atomic so we either test the FC or the VM but not both at the same time, that enters the territory of Acceptance/Integration Tests where UI Tests have clear advantages as you don't need to be aware of all the underlying implementation.

TechnicalDocuments/Proposals/BuildersToFunctions.md
sergdort:
```
enum Builder {
    static var buildVC: () -> UIViewController = {
        let analitics = Current.analyticsService
        return UIViewController()
    }
}
```

sergdort: But the issue now is that it's mutable 🙈

TechnicalDocuments/Proposals/BuildersToFunctions.md
ilyapuchka: That I agree with, but as I said this change requires a separate discussion. What will be an outcome of this proposal if the decision will be that we don't change our approach with `Current`?

TechnicalDocuments/Proposals/BuildersToFunctions.md
ilyapuchka: @sergdort It's fine and exactly the same how we agreed to use `Current` in builders. It's mutable now but we can just change it to `public internal(set) var Current: World = .production` to allow mutation only from tests using `@tesable import`

TechnicalDocuments/Proposals/BuildersToFunctions.md
RuiAAPeres: Based on a preliminary and informal assessment with a couple of members of the team, moving `MainUserSession` to `Current` seems to be desirable, especially by @daniloaliberti since it would make his work easier (with the `Current` migration). 

In any case, if we don't move `MainUserSession` to `Current`, we will park this.

dmcrodrigues:
I think there's a strong insistency around testing FlowControllers when:

1. we don't have any tests for them or expecting to have them anytime soon;
2. even if that changes they can be tested to make sure they handle everything correctly without involving the VM which is not a dependency of them and so they shouldn't be involved in any sort of FC's unit tests.

There's a big assumption related to this proposal, I agree with that but at the same time we are moving towards making the app available offline without requiring any network activity as part of @andersio's work which we already discussed and agreed on, so based on that I don't see a strong reason to keep user session out of `Current` as I think it aligns with the direction we are taking anyway.

ilyapuchka:
@dmcrodrigues 

> even if that changes they can be tested to make sure they handle everything correctly without involving the VM which is not a dependency of them and so they shouldn't be involved in any sort of FC's unit tests.

VM are not dependencies of flow controller indeed, but builders which are seating them are.

>  @andersio's work which we already discussed and agreed on

Is there a reference to this discussion and agreement that to achieve that goal `UserSession` should be moved to `Current`?

mluisbrown:
I really don't see why testing FlowControllers is even being mentioned in this proposal, much less why it is such a contentious discussion. I was going to say exactly what @dmcrodrigues said above.

This proposal is completely orthogonal to the testability of flow controllers, so this is a distraction to discussing the merits of the proposal itself.

ilyapuchka:
@mluisbrown there is an alternative (it's not even that much alternative, it's merely an adjustment) solution proposed that achieves the same results but does not have flaws of original solution. But it is being dismissed as orthogonal and going in a different direction. That makes me question the whole proposal process, the way this discussion is going on it is "my way or a highway" - we either do it the way proposed or we don't do anything.

mluisbrown:
> there is an alternative (it's not even that much alternative, it's merely an adjustment) solution proposed that achieves the same results

Are you referring to making the Builders pointfree structs as protocols? That might achieve some of the same things as this proposal (getting rid of protocols, at least), but it's fundamentally opposed to the core point of this proposal, which is to get rid of builders being an entity at all and just replace them with (namespaces) free functions.

> That makes me question the whole proposal process, the way this discussion is going on it is "my way or a highway" - we either do it the way proposed or we don't do anything.

Why? You are free to disagree with the proposal, as you already have. You are also free to create your own proposal.

RuiAAPeres:
@ilyapuchka it wasn't my intention to make you feel bad about the review process and If I did so, it was intentional and I am sorry. 

I would like to add that the same way you have the right to accept/reject my proposal, I have the same right to accept/reject your suggestions. I think this right goes both ways and I don't think I am obliged to accept your direction. 

Nevertheless, if there was a group of people raising the same concerns regarding testability, then I would assume that my thought process was flawed and I would have to re-adjust this. That's not the case so far. 

@ilyapuchka  I am not particularly against @sergdort suggestion regarding static vars. If this change would be added to the proposal, would it change your opinion?

RuiAAPeres:
@mr-v thank your comments. Is there a particular point that you would like to change on this proposal?

TechnicalDocuments/Proposals/BuildersToFunctions.md
daniloaliberti: To be more precise: to move `AppConfiguration` to Current, there's the `tabs` typed as `protocol TabContentBuilder` type, which declares a `make` method with a parameter of type `AuthenticatedTabContentBuildingMaterials`.
Considering such Material (and others) can be created from `Current`, then `tabs` construction would be much simpler.

viorel15: We have SDK's to do what MainUsesession does and if we go to this route we will never consume our SDK's. 
And if there is a big issue for us then maybe this is the same for our customers and we should rethink the SDK strategy.

RuiAAPeres: The name "`MainUserSession`" is used as an umbrella name to refer to an entity that is able to make authenticated and non-authenticated requests. That entity would use the SDK.

ilyapuchka:
I don't think creating rival proposals and discussing them for other two weeks is a way to go when we can reach an agreement discussing this one.

> I don't think I am obliged to accept your direction.

It was dismissed without any discussion and it was proposed to suggest it somewhere else. That's "my way or a highway" approach.

> Are you referring to making the Builders pointfree "structs as protocols"?

Yes, correct.

> but it's fundamentally opposed to the core point of this proposal, which is to get rid of builders being an entity at all and just replace them with (namespaces) free functions.

That's not the goal of the proposal, that's a proposed solution. The goal is to make working with builders less painful, which I can't agree with more.

> @ilyapuchka I am not particularly against @sergdort suggestion regarding static vars. If this change would be added to the proposal, would it change your opinion?

That would be very similar to what was proposed as alternative, so it's better than nothing, but I still don't see a strong reasoning to make them static.
What is the reason for such a strong preference of abstract static factories (static methods or static closures) over what is effectively just abstract factories (instance closure properties)? What are the pros and cons?

Abstract static factory with static methods:
Pros:
- ergonomic, does not have to pass any dependencies when using Current
Cons:
- not mockable (all given suggestions so far are about mocking other layers, not builders)

Abstract static factory with mutable static closure properties:
Pros:
- ergonomic, does not have to pass any dependencies when using Current
- mockable
Cons:
- mocking in one place affects all places (if closure is accessed in multiple places but need to be mocked differently)
- lost named arguments on call site

Abstract factory with instance method (current approach):
Pros:
- ergonomic, does not have to pass any dependencies when using Current, can create instance in place as it is stateless (`Builders.ClinicalRecords()`)
Cons:
- hard to mock, have to use protocols to mock factory methods

Abstract factory with instance closure properties:
Pros:
- ergonomic, does not have to pass any dependencies when using Current, so can create instance in place as it is stateless (`Builders.ClinicalRecords()`)
Cons:
- lost named arguments on call site

It will worth to mention that all these approaches also lead to spreading feature toggles all over the code base as they are accessed in place, what can be instead done in one place, AppDelegate, where the whole dependencies graph would be instantiated, for what we would need to inject builders in flow controllers rather than accessing them directly via type or creating instance in place. But this is renderer as a problem (I can agree with that) and we are fine with accessing feature toggle like that anyway and moving in this direction already, so it's fine.

TechnicalDocuments/Proposals/BuildersToFunctions.md
ilyapuchka: I think @viorel15 actually raised a good point. We may have two orthogonal goals here - one is consuming sdk, another is making user session accessible globally to enable this approach (or any other approach that would avoid passing builder instances around and injecting session into them). 
We can though make `Current` consume session object from SDK (I assume there is something like that) to solve it.

RuiAAPeres:
I agree with this:

> I don't think creating rival proposals and discussing them for other two weeks is a way to go when we can reach an agreement discussing this one.

It doesn't make sense to open a new proposal that goes against another proposal already open. It's just noise. 

--- 

> That's not the goal of the proposal, that's a proposed solution. The goal is to make working with builders less painful, which I can't agree with more.

Every proposal should be accompanied with a solution, so inherently you are attaching a solution to your end goal.

--- 

I still don't feel that keeping structs/classes/entities for Builder is the right way to go about this, especially with `Current` in place. Now your amendment is if I am not mistaken:

```
Abstract factory with instance closure properties:
Pros:
 - ergonomic, does not have to pass any dependencies when using Current, so can create instance in place as it is stateless (Builders.ClinicalRecords())
Cons:
 - lost named arguments on call site
```

This would pollute the FlowController APIs with default Builders. I mean, yes it gives you flexibility, but I am unsure we actually need it, given that testability is not an issue. 

It would boil down to this:

```swift
struct FlowControllerFooBar {
   private let primary: Flow
   private let builder: BuilderFooBar

   init(primary: Flow, builder: BuilderFooBar = Builder.FooBar()) {
      self.primary = primary
      self.builder = builder
   } 

   func handle(state: State) {
      switch (state) {
       case .bar: 
           builder.makeFooBar() |> primary.present
      }
   }
}
```

versus:

```swift
struct FlowControllerFooBar {
   private let primary: Flow

   init(primary: Flow)) {
      self.primary = primary
   } 

   func handle(state: State) {
      switch (state) {
       case .bar: 
            Builder.FooBar.makeFooBar() |> primary.present
      }
   }
}
```

Is this accurate @ilyapuchka ?

NSMyself:
I think it's a clear improvement on our current status, it'd remove a lot of boilerplate and it's conveniently useful being able to use autocomplete to get every builder. If we're not doing heavy FlowController testing then all the more reason to go with this proposal but I wouldn't oppose @sergdort's suggestion either.

Clear 👍  in my book, regardless how we look at this.

ilyapuchka:
@RuiAAPeres yes, that's accurate. I don't see it as a pollution of API though. If the need arrises to mock these builders or to pass different implementation of builder for any other reason we will be able to do that by passing an instance with a different implementation of a factory method (closure), this will be impossible with static methods and may be harder to do with static properties. This again can be solved by accessing feature switches in one or another place which as I mentioned we seem to be fine with.

RuiAAPeres:
@ilyapuchka given that your amendment is aligned with the motivations behind this proposal and addresses your concerns, I am ok with changing this proposal. Does that make sense?

TechnicalDocuments/Proposals/BuildersToFunctions.md
RuiAAPeres: @viorel15 Please let's stay on point. If you want to discuss the usage of `MainUserSession` and SDK, let's start the discussion somewhere else. Neither has anything to do with this proposal.

viorel15: It's fine. I understood that for this proposal it does not matter if it's main user session as it is now or refactored to include the SDK's.

ilyapuchka: @daniloaliberti this problem (dependency of app configuration on session) seems to be a real monster here, is it agreed that moving user session in the `Current` is the solution we going forward? Because there is also a parallel discussion of refactoring MainUserSession, which may solve it in a different way (I'm not up to date with it).
This problem also makes me wonder about other requirements of `AuthenticatedTabContentBuildingMaterials`, what happens with them? And another question is do we really need that in `BabylonDependencies` and `Current` specifically? It would be better to discuss that in a different channel not to diverge this discussion.

zzcgumn:
My day disappeared quickly and I will make some more detailed comments tomorrow. I think the disadvantages can be mitigated by passing and instance of `World` to each factory method.

```swift
extension Builder {
    enum ClinicalRecords {
              static func makeMedicalHistory(modal: Flow, current: World = Current) -> UIViewController {
                 ....
               }
      }
}
```

This would allow us to use the builders in unit tests or playground driven development by passing in a test version of `Current`.

```swift
let mockedRecords = Builder.ClinicalRecords. makeMedicalHistory(modal: UIViewController.modal, current: .mock)
```

or perhaps a more specialised version 

```swift
let dreamRecords = Builder.ClinicalRecords. makeMedicalHistory(modal: UIViewController.modal, current: .inMartinsDreamWorld)
```

I am also wondering if changing the return type can improve things further.

```swift
extension Builder {
     enum ClinicalRecords {
           static func makeMedicalHistory(modal: Flow, _ current: World = Current) -> Result<UIViewController, AuthenticationError> {
                 ....
            }
      }
 }
```

zzcgumn:
> The proposed solution assumes that a way to authenticate HTTP requests is available at `Current` level

I feel this requires some careful analysis. We are about to start refactoring MainUserSession from a monolith to a composite. I can see several advantages in serving an authenticated context and clinical authentication from Current. I am, however, not sure if we need to refactor MainUserSession before serving an authenticated context from Current.

I try my best to expand a bit on this topic tomorrow.

gtsiap:
I am in favour of this proposal, since it will remove a lot of our boilerplate and it will improve discoverability.

Also I liked @zzcgumn's idea about passing `World` as a default value.

RuiAAPeres:
@zzcgumn @gtsiap I followed Ilya's suggestion. Can you please review the approach suggested?

TechnicalDocuments/Proposals/BuildersToFunctions.md
```diff
@@ -74,88 +78,74 @@ Given the Module where it's located it could be migrated to:
 
 ```swift
 extension Builder {
-    enum ClinicalRecords {
-        static func makeMedicalHistory(modal: Flow) -> UIViewController
+    struct ClinicalRecords {
+        internal(set) var makeMedicalHistory: (Flow) -> UIViewController = { _ in UIViewController() }
```
ilyapuchka: Unfortunate negative effect of that, as I was mentioning as cons, is that now on call site we wouldn't be able to use named arguments. Static functions don't have it. We could solve it, if needed, with adding a method wrapping such closure:

```swift
struct ClinicalRecords {
    internal(set) var makeMedicalHistory: (Flow) -> UIViewController = { _ in 
        return UIViewController() 
    }
    func makeMedicalHistory(flow: Flow) -> UIViewController {
        return self.makeMedicalHistory(flow)
    }
}
```
I haven't compiled this code but I don't think it will result in ambiguity or something. This is a bit of boilerplate so maybe not worth it.

zzcgumn:
I would just like to add some comments about a pattern for initialising builders that I think will be ergonomic.


```swift
	extension Builder {
	    struct ClinicalRecords {
	        let makeMedicalHistory: (Session, Flow) -> UIViewController 
	        let makeMedicalHistory: (Flow) -> UIViewController

                init(makeMedicalHistory: (Session, Flow) -> UIViewController,
                     makeMedicalHistory: (Flow) -> UIViewController) {
                      ....
                }

                static functional make(current: World = Current) -> Result<ClinicalRecords, AuthenticationError> {
                    let configuration = ClinicalRecords.Configuration(LocalizationClinicalRecords....)
                    let clinicalRecordsSDK = BabylonSDK.ClinicalRecords(with: current.session.authenticatedContext, configuration: configuration)
                   ...
                }           
	    }
	}
```

What I am discussing here are, obviously, implementation details and there is no need to resolve them before accepting this proposal. The important points that I am trying to illustrate are:

1. We need to think about error handling if the `session` inside `Current` is not in the correct state.

2. `World` would not handle any state, it would just hold a default `session` (or whatever we decide to call it).

3. There is no impact at the SDK level. These builders will create required SDK instances from the contents of `Current`.

4. Test environments can be injected at two levels, either as factory methods or by passing a `Current.mock`. 

5. With well designed initialisers, this version of the proposal becomes very similar to the first.

RuiAAPeres:
@zzcgumn this proposal is orthogonal to `MainUserSession` and should be reviews on its own merits:

> The goal of this proposal is to simplify the usage of builders across the codebase, improve discoverability and development ergonomics.

At no point `MainUserSession` is mentioned as part of said goals.

gtsiap:
Here is an example based on this comment

```swift
struct FooFlowController {
   let barViewController: () -> UIViewController
   let zViewController: () -> UIViewController
   let primary: Flow
   let secondary: Flow

   .......
}


enum Builder {
    enum Foo {
        static func make(userSession: UserSession, primary: Flow, primary: Flow) -> UIViewController {
            Builder.Bar.make(userSession: userSession)
            Builder.Z.make(userSession: userSession)

            let fooViewModel = FooViewModel()
            let fooViewController = FooViewController()
            fooViewController.viewModel = fooViewModel

            let flowController = FlowController(
                barViewController: {},
                zViewController:, 
                primary: primary, 
                secondary: secondary
            )

            fooViewModel.routes.observeValue(flowController.handle)

            return fooViewController
        }
    }
}
```

ilyapuchka:
I like what @gtsiap is suggesting, That's similar to approach demonstrated [here](https://www.youtube.com/watch?v=uTLG_LgjWGA) but in application to view models.

But:
- it still suffers from unnamed parameters issue (the local names in closures are still not enforced).
- another issue is that it will force us to define a closure in flow controller for each builder method, rather than declaring just a single builder property, which also may negatively affect ergonomics

gtsiap:
> * another issue is that it will force us to define a closure in flow controller for each builder method, rather than declaring just a single builder property, which also may negatively affect ergonomics

Yes, only if we decide to unit test flow controllers which isn't the case right now and even if we decide to unit test flow controllers in the future, it's less complicated to mock a closure rather than a builder.

ilyapuchka:
@gtsiap there is no problem with mocking builder instance as well, one can create an instance of it and replace only a closure that needs to be mocked, there is only a small level of indirection introduced by defining a struct type that at the same time provides name spacing to group multiple methods and effectively inject them all together rather than one by one.

I'd amend your suggestion with flow controllers having default values for closures:

```swift
struct FooFlowController {
   let barViewController: @escaping (...) -> UIViewController = Builder.Bar.make
   let zViewController: @escaping (...) -> UIViewController = Builder.Z.make
   let primary: Flow
   let secondary: Flow

   .......
}
```

This will effectively make it the same as the latest version in the proposal with exception of defining a structure for builder.

RuiAAPeres:
> Transitive dependencies that do not go through Current would now have to be held by Flow Controllers after this.

Please provide an example of that.

> In any case, if we say all dependencies are to be passed via Current, stateless or stateful, we would reach a similar place as this proposal dictates anyway.

We don't say that we simply state that certain dependencies (even the stateful ones) make sense to be in `Current`. We sacrifice purity for development experience/ergonomics.

TechnicalDocuments/Proposals/BuildersToFunctions.md
```diff
@@ -0,0 +1,158 @@
...
+    struct ClinicalRecords {
+        static func makeMedicalHistory(with session: Session, flow: Flow) -> UIViewController 
+        static func makeMedicalHistory(flow: Flow) -> UIViewController
+    }
+}
+```
+
+This would remove the intricate structure we have with builders and their children and protocols.
+
+## Impact on an existing codebas
```
daniloaliberti:
```suggestion
## Impact on an existing codebase
```

TechnicalDocuments/Proposals/BuildersToFunctions.md
```diff
@@ -0,0 +1,158 @@
...
+    }
+```
+
+Particular medical histories (e.g. for minors) can also be created, but in this case, the minor session would be passed instead of relying on `Current`. This situation (with child builders) can be observed at `ClinicalRecordsChildBuilders.swift`:
+
+```swift
+extension Builder {
+    struct ClinicalRecords {
+        static func makeMedicalHistory(with session: Session, flow: Flow) -> UIViewController 
+        static func makeMedicalHistory(flow: Flow) -> UIViewController
```
daniloaliberti: If I understand correctly, the minor's is the one with the `session` param, right? 
I'd change the method name to avoid confusion defining better the difference.

TechnicalDocuments/Proposals/BuildersToFunctions.md
```diff
@@ -0,0 +1,158 @@
...
+Using Ilya's suggestion about structs with functions as properties, versus static free functions. The overall goals of the proposal are preserved and it provides flexibility for testing. 
+
+---
+
+Decision made during the call:
+
+1. Popular Vote: Use/Move UserSession (not the entity, but what it represents), to `Current`. **Majority has approved.**
+2. Popular Vote: In favour of the original proposal to use static builder methods. **Majority has approved.**
+   1. Amendment to the original proposal: **by default** `FlowControllers` will only be injected with `Flows`, but if testing is deemed, then further dependencies can be passed.
+
```
daniloaliberti: I see `analytics` being often triggered in flowControllers. 
And along with this, (if it's still an assumption) just `Builder`s can access `Current`, so then analytics can be injected to a flowController.

RuiAAPeres: analytics should be in `ViewModels` not FlowControllers.

TechnicalDocuments/Proposals/BuildersToFunctions.md
RuiAAPeres: yeah you are correct.

PR #72, opened 2019-03-27
add Component Composition Documentation

gtsiap:
Document how we deal with Component Composition.


TechnicalDocuments/CompomentComposition.md
```diff
@@ -0,0 +1,63 @@
...
+        text: "Some Details",
+        styleSheet: LabelStyleSheet()
+            .compose(\.textColor, tokens.colors.grey700)
+            .compose(\.font, font(fontAttributes: tokens.typography.subhead))
+    ).asAnyRenderable()
+]
+.stack(
+    axis: .vertical,
+    spacing: 2
+)
```
sergdort: Would be nice also to have an image of how it's going to look like. We could make a table e.g

| Code | Screenshot |
| --- | --- |
| some code | screenshot |

TechnicalDocuments/CompomentComposition.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+Currently `Renderable` provides some util methods like `card(..)`, `stack(...)`, `width(...)` and `height(...)`.
```
sergdort: Would be nice to mention that someone can make their own extensions. For example I just needed `minHeight` behaviour. I guess that we'll provide documentation how to create them later

TechnicalDocuments/CompomentComposition.md
```diff
@@ -0,0 +1,63 @@
...
+
+With `UIStackView` based composition we can have truly self sizing cells but this means that its incompatible with `BentoKit`, because `BentoKit` doesn't really provides self sizing cells. `BentoKit` is using Auto Layout but it also provides the size of the component via `HeightCustomizing`.
+In order to make a `BentoKit` component compatible with `UIStackView` based composition you either have to drop `HeightCustomzing` or you should use `UIView.systemLayoutSizeFitting` in your `HeightCustomzing` implementation.
```
sergdort: I do not think this is correct, `UIView.systemLayoutSizeFitting` won't save you there, right?

gtsiap: Why not? If you can create a `Renderer.View` and you configure it and you call `UIView.systemLayoutSizeFitting` it will return the correct size, but you will have to call it correctly.

Of course if your `BentoKit` component is too complicated, Auto Layout may produce warnings and stuff like that. For instance if you try to compose `Component.TitledDescription` with `UIStackView` even if you use `UIView.systemLayoutSizeFitting` probably it won't work well.

Am I missing something?

Are there other implications of using `HeightCustomzing` that I am missing?

TechnicalDocuments/CompomentComposition.md
gtsiap: How can we have something like this?

`enter` creates a new line we would have to create new lines with `<br>` which isn't neither convenient nor scalable.

I believe that the `GalleryApp` that we have is good, its real working code after all :)

TechnicalDocuments/CompomentComposition.md
```diff
@@ -0,0 +1,64 @@
+# Bento Component Composition Using `UIStackView`
```
diegopetrucci: There's a typo in the file name `compoNent`

TechnicalDocuments/CompomentComposition.md
```diff
@@ -0,0 +1,64 @@
...
+#### Updated at: 23-3-2019
+
+## The scope of this document 
+Component Composition with `Bento` is a very wide topic due to its flexible design.
+`Bento` components can be composed in many ways.
+
+The purpose of this document is to provide documentation for __composing Bento components using `UIStackView`__.
+Other ways of composition are out of the scope of this document.
+
+## How to use component composition in your framework
```
diegopetrucci: Which framework?

TechnicalDocuments/CompomentComposition.md
```diff
@@ -0,0 +1,64 @@
...
+| <ul><li> Step 1: Compose your components like in the code sample. This step means that you use Auto Layout. </li><li> Step 2: render your `Box` </li> <li>Step 3: `UITableView`/`UICollectionView` use Auto Layout in order to calculate the size of the cell.</li></ul> | Step 1: Implement your components using Auto Layout. </li> <li> Step2: Implement `HeightCustomizing` and calculate the size of the component. <li> Step 3: render your `Box` </li> <li>Step 4: `UITableView`/`UICollectionView` use Auto Layout in order to calculate the size of the cell __if the component doesn't comfort to `HeightCustomizing` othewise it uses the provided value__.</li></ul>  |  <li>Step 1: Implement your components *and* use Auto Layout. </li><li> Step 2: render your `Box` </li> <li>Step 3: `UITableView`/`UICollectionView` use Auto Layout in order to calculate the size of the cell.</li></ul>  |
+
+With `UIStackView` based composition we can have truly self sizing cells but this means that its incompatible with `BentoKit`, because `BentoKit` doesn't really provides self sizing cells. `BentoKit` is using Auto Layout but it also provides the size of the component via `HeightCustomizing`.
```
diegopetrucci:
```suggestion
`UIStackView`-based composition allows us to have true self-sizing cells, but that makes incompatible with `BentoKit`, as the latter does not support self-sizing cells (it uses Auto Layout and it also provides the size of the component via `HeightCustomizing` directly).
```

TechnicalDocuments/CompomentComposition.md
```diff
@@ -0,0 +1,64 @@
...
+## How to use component composition in your framework
+
+Well, the answer to this question is that your can't.
+The reasoning for this decision is that
+* we must reuse components via the `DesignLibrary`
+* If you need something custom and customizable then using `UIStackView`s probably won't be good enough for you
+* We need to be flexible and to depend on abstractions.
+
+#### Where and how to use component composition
+In `BabylonUI` there is a new namespace called `UI` which is the place for `Atomic Components`.
```
diegopetrucci:
```suggestion
`BabylonUI` now has a new namespace called `UI`, which contains a few `atomic` components.
```

TechnicalDocuments/CompomentComposition.md
```diff
@@ -0,0 +1,64 @@
...
+
+- `Atomic Components` are the `Bento` Components version of `UIKit`'s components so the nomenclature is similar  like `UILabel` and `UI.Label`, `UIButton` and `UI.Button` etc
```
diegopetrucci:
```suggestion
- `Atomic Components` are the `Bento` equivalent of standard `UIKit` components — this is why they retain a similar nomenclature (e.g: `UILabel` and `UI.Label`, `UIButton` and `UI.Button`);
```

TechnicalDocuments/CompomentComposition.md
```diff
@@ -0,0 +1,64 @@
...
+The reasoning for this decision is that
+* we must reuse components via the `DesignLibrary`
+* If you need something custom and customizable then using `UIStackView`s probably won't be good enough for you
+* We need to be flexible and to depend on abstractions.
+
+#### Where and how to use component composition
+In `BabylonUI` there is a new namespace called `UI` which is the place for `Atomic Components`.
+
+- `Atomic Components` are the `Bento` Components version of `UIKit`'s components so the nomenclature is similar  like `UILabel` and `UI.Label`, `UIButton` and `UI.Button` etc 
+- `Atomic Components` use `Stylesheets` for styling
```
diegopetrucci:
```suggestion
- `Atomic Components` use `Stylesheets` for styling;
```

TechnicalDocuments/CompomentComposition.md
RuiAAPeres: > I believe that the GalleryApp that we have is a good solution for now, its real working code after all :)
 
Not really, if it forces people to dig through code every time they want to do something. In a way it defeats the purpose of documentation if you defer with "check the code."

gtsiap: The `GalleryApp` is written in such a way which is similar with Xcode playgrounds.

Every component is in a different group/folder like 

|--- Rows
|--- ItemPicker
|--- Toggle

and within these folders there is only *one* file with content similar to this

```swift
struct ItemPickerBuilder {
    func make() -> UIViewController {
        return ComponentPreviewerBuilder().make { library in
            [
                library.componentsBuilder()
                    .itemPicker(
                        title: "Country",
                        subtitle: "UK"
                    )
            ]
        }
    }
}
```

So yes you "check the code" but its not really that you have to dig in order to find it.

sergdort: > Would be nice also to have an image of how it's going to look like. We could make a table e.g

Here I mean that when we provide a composition example in the documentation to have a screenshot an the side to show how it's going to look like

TechnicalDocuments/CompomentComposition.md
gtsiap: What do you mean?

It means how to use `Component Composition` in your application/framework/screen.

TechnicalDocuments/CompomentComposition.md
```diff
@@ -0,0 +1,64 @@
...
+|:-------:|:------:|:------:|
+| <ul><li> Step 1: Compose your components like in the code sample. This step means that you use Auto Layout. </li><li> Step 2: render your `Box` </li> <li>Step 3: `UITableView`/`UICollectionView` use Auto Layout in order to calculate the size of the cell.</li></ul> | Step 1: Implement your components using Auto Layout. </li> <li> Step2: Implement `HeightCustomizing` and calculate the size of the component. <li> Step 3: render your `Box` </li> <li>Step 4: `UITableView`/`UICollectionView` use Auto Layout in order to calculate the size of the cell __if the component doesn't comfort to `HeightCustomizing` othewise it uses the provided value__.</li></ul>  |  <li>Step 1: Implement your components *and* use Auto Layout. </li><li> Step 2: render your `Box` </li> <li>Step 3: `UITableView`/`UICollectionView` use Auto Layout in order to calculate the size of the cell.</li></ul>  |
+
+`UIStackView`-based composition allows us to have true self-sizing cells, but that makes incompatible with `BentoKit`, as the latter does not support self-sizing cells (it uses Auto Layout and it also provides the size of the component via `HeightCustomizing` directly).
```
ilyapuchka: That's not entirely correct because only some BentoKit components use `HeightCustomizing` and it's not a requirement of implementing Bento component. In the end Bento Component per se is just a namespace for view, stylesheet and component description, no one stops from using stack views or pure auto layout for that.

TechnicalDocuments/CompomentComposition.md
```diff
@@ -0,0 +1,64 @@
...
+
+When you compose `Bento` components using `UIStackView` __there is an implicit behaviour change on how you create your UI__.
+
+So let's start with how it works
+
+| Atomic Components using UIStackView | BentoKit | Vanilla Bento |
+|:-------:|:------:|:------:|
+| <ul><li> Step 1: Compose your components like in the code sample. This step means that you use Auto Layout. </li><li> Step 2: render your `Box` </li> <li>Step 3: `UITableView`/`UICollectionView` use Auto Layout in order to calculate the size of the cell.</li></ul> | Step 1: Implement your components using Auto Layout. </li> <li> Step2: Implement `HeightCustomizing` and calculate the size of the component. <li> Step 3: render your `Box` </li> <li>Step 4: `UITableView`/`UICollectionView` use Auto Layout in order to calculate the size of the cell __if the component doesn't comfort to `HeightCustomizing` othewise it uses the provided value__.</li></ul>  |  <li>Step 1: Implement your components *and* use Auto Layout. </li><li> Step 2: render your `Box` </li> <li>Step 3: `UITableView`/`UICollectionView` use Auto Layout in order to calculate the size of the cell.</li></ul>  |
```
ilyapuchka: Table format is not very convenient here, there separate booklet lists would be easier to read. It also have almost the same content for each column so probably what really needs to be explained is only why `HeightCustomising` thing is/was needed in the first place and how it affects the component, i.e. that the calculated height will be always preferred to the height set by other means, like `Height` component from Design Library.

ilyapuchka: We could even maybe have a runtime warning when `height` is applied on a component that implements `HeightCustomising` protocol, if that's even possible to check. cc @sergdort

RuiAAPeres: 👀

TechnicalDocuments/CompomentComposition.md
gtsiap: >  I thought that if you need to implement a custom UI that can't use any of existing components - go on and compose atomic components in your renderer. The explanation though makes the emphasis on that it's not even possible. If that's really the case then what's the solution?

Atomic components are only available in `BabylonUI`.
New components can be added, but they have to be in `DesignLibrary`.
`Renderer`'s shouldn't compose components.

PR #74, opened 2019-03-28
[Proposal] Top-priority member-variable and initializer declarations

inamiy:
TLDR:
Moving member variables (states) and initializers (constructors) to the top of the scope.
Inspired by Rust and Haskell.

Twitter opinions: https://twitter.com/inamiy/status/1111103087693844480

----

### UPDATED 2019/04/26

My current preference is the following (but this doesn't have to be strict for this proposal):

```
Data type declaration (required in declaration scope):
1. instance stored properties (struct), or enum cases
2. designated initializers
3. deinit

In extension (if possible):
4. convenience initializers
5. instance computed properties
6. instance methods 

In extension (if possible):
7. type stored properties (NOTE: this rarely appears in practice because singleton is discouraged)
8. type computed properties
9. type methods

In extension (if possible):
10. nested types
```

The basic concept here is: **whenever you can move code to `extension`, that's not the important code**.
So, concentrating on **states (stored properties) and constructors first** is the whole idea of this proposal.

### DIFF
--- a/TechnicalDocuments/Proposals/TopPriorityMemberVariables.md
+++ b/TechnicalDocuments/Proposals/TopPriorityMemberVariables.md
@@ -119,7 +119,7 @@ Requires many efforts for manual refactoring and resolving conflicts.
 ## Appendix
 
 See [#74](https://github.com/Babylonpartners/ios-playbook/pull/74) for more strict declaration order rule, but it is NOT a part of this proposal.
-(This proposal only contains section 0 to 3)
+(This proposal only contains section 1 to 3)
 
 ```
 Data type declaration (required in declaration scope):
--- a/TechnicalDocuments/Proposals/TopPriorityMemberVariables.md
+++ b/TechnicalDocuments/Proposals/TopPriorityMemberVariables.md
@@ -115,3 +115,30 @@ Requires many efforts for manual refactoring and resolving conflicts.
 ## Alternatives considered
 
 - No alternatives. (Reject this proposal and stay as is)
+
+## Appendix
+
+See [#74](https://github.com/Babylonpartners/ios-playbook/pull/74) for more strict declaration order rule, but it is NOT a part of this proposal.
+(This proposal only contains section 0 to 3)
+
+```
+Data type declaration (required in declaration scope):
+0. type stored properties (NOTE: this rarely appears in practice because singleton is discouraged)
+1. instance stored properties (struct), or enum cases
+2. designated initializers
+3. deinit
+
+(Below order is optional)
+
+In extension (if possible):
+4. convenience initializers
+5. instance computed properties
+6. instance methods
+
+In extension (if possible):
+7. type computed properties
+8. type methods
+
+In extension (if possible):
+9. nested types
+```
--- a/TechnicalDocuments/Proposals/TopPriorityMemberVariables.md
+++ b/TechnicalDocuments/Proposals/TopPriorityMemberVariables.md
@@ -1,0 +1,117 @@
+# Top-priority member-variable and initializer declarations
+
+* Author: Yasuhiro Inami
+* Review Manager: Rui Peres
+
+## Introduction
+
+This proposal is to improve code readability for every declared type by **moving member variables (states) and initializers (constructors) to the top of the scope**.
+
+## Motivation
+
+The code becomes hard to read when types are heavily nested, especially when inner type's member variables appear before outer type's variables in the file.
+
+For example, below code shows that we can't immediately figure out which member variables belong to which type, and what kind of constructors are available:
+
+```swift
+enum Outer {
+    struct Inner {
+        func foo() {
+            ...
+            ...
+            ...
+        }
+
+        let a: A
+        let b: B
+
+        init(a: A, b: B) {
+            ...
+        }
+    }
+
+    func foo() {
+        ...
+    }
+
+    case c
+
+    var bar: Bar {
+        ...
+    }
+
+    init?(c: C) {
+        ...
+    }
+
+    case d(D)
+}
+```
+
+Since understanding the data structure i.e. product type (member variables a.k.a. states) and sum type (enum cases a.k.a. constructors) are the key to understand the types, architecture, and whole business logic, it is important to organize the code as proposed in the next section.
+
+## Proposed solution
+
+**Move member variables (states) and initializers (constructors) to the top of the scope**, at least manually, and in future, automatically by formatter tool e.g. SwiftSyntax.
+
+For example:
+
+```swift
+// Group constructors (enum cases).
+enum Outer {
+    case c
+    case d(D)
+
+    // NOTE: This can be implemented as `extension Outer` as well
+    init?(c: C) {
+        ...
+    }
+}
+
+// Methods (non-constructors) should move to extension scope to separate from constructors.
+// Or, they can be in the same scope as `enum Outer` but MUST NOT be at the top of its scope.
+extension Outer {
+    func foo() {
+        ...
+    }
+
+    var bar: Bar {
+        ...
+    }
+}
+
+// Declare inner types using outer type's extension
+// (so that entire scope can be easy to migrate to other file if needed)
+extension Outer {
+
+    // Group member variables (struct stored property).
+    struct Inner {
+        let a: A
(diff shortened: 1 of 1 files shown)


inamiy:
### Caveat

There are some edge cases when we want to add immutable `Property` on top of private `ReactiveSwift.MutableProperty` or `RxSwift.BehaviorRelay` but put these 2 code together in same place:

```swift
class ViewModel {
    // Edge-case: computed property is declared before member variable!
    var property: Property<Int> { return Property(_behaviorRelay) }
    private let _behaviorRelay: BehaviorRelay<Int>
}

// NOTE: This doesn't work exactly the same in ReactiveSwift because of onCompleted in MutableProperty
```

In above case, I personally recommend to code-gen `var property` computed property via Sourcery so that we can avoid this edge-case.
(Or, we can just allow this by checking certain type signatures...)

AliSoftware (commented):
Should this proposal also address:
- position of type members (`static var/let`)
- position of private methods (at end of type or even in a `private extension`)

I'm in favor of the general idea of that proposal 👍. But I think it could additionally:
- benefit from just listing the order in which we declare every kind of declaration (e.g. first type members then instance members then init then functions then private functions?)
- should address the case of memberwise inits in struct (especially the fact that it's common practice to declare convenience inits in an `extension` for `structs` when you don't want to lose the auto-synthesised memberwise init, a case not considered here)
- could also define our rules/guidelines for grouping: should we order the instance members in arbitrary order, in alphabetical order, in semantic order by grouping related ones together, something else, we don't care?
- suggest the use of `// MARK:` to separate sections between member declarations and init and the rest

inamiy:
@AliSoftware 

- For `static var/let` position, I added my preference at the top comment, which is to [prioritize instance methods over class methods](https://twitter.com/inamiy/status/1111153312068300800). But this is just my opinion yet that doesn't have to be strict, and we can only apply the rule for "instance" variables / inits / methods for now.
- "Define our rules/guidelines for grouping" is difficult in general, which is out of syntactic analysis, so I won't give further proposal here.
- `// MARK:` is definitely +1 for better documentation 👍 
- "Order state properties vs computed properties" is sometimes pain to think e.g. https://github.com/Babylonpartners/ios-playbook/pull/74#issuecomment-477431751 , but I prefer to just follow the basic rule: state property comes first
- `didSet` will not affect the order since rule should be simple.

AliSoftware:
Got you. I still think the proposal would be more readable if presented as an ordered list, like [harmoniclattice](https://twitter.com/harmoniclattice/status/1111142267752976384?s=21) did on Twitter (not saying we should use the order they mention, just the presentation of it)

If I understand correctly what you're suggesting, probably something like this for us then?
- instance stored properties
- type stored properties
- initializers
- nested types
- type computed properties
- type methods
- instance computed properties
- instance methods

And if so, where do private vs public go?
Should we declare private instance properties next to public/internal ones?

Personally I think such a proposal should answer those questions — at least give the general guideline we want to try and follow for the common case — as just saying "member variables at the top" is too ambiguous otherwise (e.g. what about private and computed vars). A clear ordered list would help clarify the ambiguity.

inamiy:
Updated the top comment.
In this proposal, I want to only **focus from 0. to 3. declarations aligned in order** so that rule becomes simple and easy to adopt:

```
0. type stored properties (singleton is HIGHLY DISCOURAGED)
1. instance stored properties (struct), or enum cases
2. initializers
3. deinit
4. (Anything else will appear here in whatever order you want)
```

@AliSoftware 
For nested types, I'm confident that it can move to the bottommost of the file, or in another file.

Also, any "type XXX" is (mostly) not really a part of the type's structure, so I think we can move it to bottom as well.
The important exception here is **`static func make(...) -> Self` which can be treated as "generalized" initializer** like GADTs (Generalized Algebraic Data Types).
It's a bit annoying where to place `static func make`, though.
I highly prefer it to be in "2. initializers", but by definition it is "7. type methods".
So, we might need a bit more documentation for the definition of "initializers" if everyone agrees on this exception.

For access modifiers e.g. `public` or `private`, I don't have strong preference for the order, since the best grouping strategy may differ depending on situations.

And same thing can be said for each ambiguous part as well.
I don't think we need a perfect unified rule right now.
We can embrace some uncertainty and brush up the rules in the future 😉

TechnicalDocuments/Proposals/TopPriorityMemberVariables.md
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// Methods (non-constructors) should move to extension scope to separate from constructors.
+// Or, they can be in the same scope as `enum Outer` but MUST NOT be at the top of its scope.
+extension Outer {
+    func foo() {
+        ...
+    }
+
+    var bar: Bar {
```
ilyapuchka: As per pr description properties should go before methods or it's not enforced?

ilyapuchka (commented):
I first was confused by suggested structure, but after some thinking it actually make sense to me.
One small piece is still missing though, regarding convenience initialisers and member wise init that @AliSoftware mentioned? Should they be defined in extension as a rule like this?

```
In extension (if possible):
4. instance computed properties
5. convenience initialisers
6. instance methods 
```

TechnicalDocuments/Proposals/TopPriorityMemberVariables.md
inamiy: Computed properties are not enforced.

inamiy:
@ilyapuchka https://github.com/Babylonpartners/ios-playbook/pull/74#pullrequestreview-223592486

> 5. convenience initialisers

Good catch! I personally prefer them placed close to designated initializer, so I would go as follows (updated PR body):

```
In extension (if possible):
4. convenience initialisers
5. instance computed properties
6. instance methods 
```

(But just a reminder that this is not a mandatory for this PR)

inamiy:
With regard to "generalized initializer" https://github.com/Babylonpartners/ios-playbook/pull/74#issuecomment-477892087 , current Swift does have ability to pattern-match like GADTs, and it's not even required to put in the same scope as type declaration.
Thus, `static func make(...) -> Foo<Bar>` should probably be treated as a normal "type methods" rather than "designated initializers".

RuiAAPeres:
@inamiy did you scheduled a meeting regarding this proposal?

inamiy:
@RuiAAPeres 
I will first update proposal to the latest version, then schedule a meeting.

inamiy:
728e3f7 Updated preferred (not mandatory) declaration order since `type stored properties` can actually move to `extension`.

inamiy:
cf. https://github.com/realm/SwiftLint/blob/master/Rules.md#type-contents-order

SwiftLint's `type-contents-order` may support or conflict with this proposal.

ilyapuchka:
@inamiy I looked at this rule and thought there is a config object I can't see if it's possible to configure it differently in yaml because the hardcoded order is not following your suggestion =/

P.S. this rule's code is weird...

```
let article = ["a", "e", "i", "o", "u"].contains(content.substring(from: 0, length: 1)) ? "An" : "A"
```

TechnicalDocuments/Proposals/TopPriorityMemberVariables.md
```diff
@@ -0,0 +1,144 @@
...
+    case d(D)
+
+    // NOTE: This can be implemented as `extension Outer` as well
+    init?(c: C) {
+        ...
+    }
+}
+
+// Methods (non-constructors) should move to extension scope to separate from constructors.
+// Or, they can be in the same scope as `enum Outer` but MUST NOT be at the top of its scope.
```
mr-v: > Methods (non-constructors) should move to extension scope to separate from constructors.

I'd stick with having them in the same scope unless there's a reason to group them together, e.g. implementing a protocol.

mr-v (approved):
Points 1-3 are a good start. 👍 


> In extension (if possible):
> 10. nested types

👍

daniloaliberti (approved):
👍

ilyapuchka (approved):

inamiy:
Closing as style guideline is merged in https://github.com/Babylonpartners/ios-playbook/pull/126 🎉

inamiy:
Oops, this was pull request 😓

PR #73, opened 2019-03-27
Update new hires checklist

mluisbrown:
As I realised whilst setting up my new Mac, the checklist is woefully inadequate and misses a number of important steps. Also, as I found out, it's important to have Git LFS installed *before* you clone the repo. That way you don't need to `git lfs pull` which can mess things up completely.

PR #75, opened 2019-04-01
Update New hire checklist

TheAdamBorek:
Updated the new hire checklist after I've setup my new machine.
I've included Oliver's notes from the last PR https://github.com/Babylonpartners/ios-playbook/pull/73

### DIFF
--- a/TechnicalDocuments/NewHiresCheckList.md
+++ b/TechnicalDocuments/NewHiresCheckList.md
@@ -41,8 +41,6 @@ Here's how to get the iOS project up and running.
 1. Globally configure Git to use SSH instead of HTTPS:
      ```
      git config --global url."git@github.com:".insteadOf "https://github.com/"
-     git config --global url."git@github.com:".pushinsteadof "git://github.com/"
-     git config --global url."git@github.com:".pushinsteadof "https://github.com/"
      ```
 1. Install Homebrew if you don't already have it installed:    
      ```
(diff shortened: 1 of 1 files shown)


TechnicalDocuments/NewHiresCheckList.md
```diff
@@ -41,8 +41,6 @@ Here's how to get the iOS project up and running.
 1. Globally configure Git to use SSH instead of HTTPS:
      ```
      git config --global url."git@github.com:".insteadOf "https://github.com/"
-     git config --global url."git@github.com:".pushinsteadof "git://github.com/"
-     git config --global url."git@github.com:".pushinsteadof "https://github.com/"
```
TheAdamBorek: Those 2 are not needed. I only had to setup the first line and everything works just fine.

AliSoftware:
Also I just saw that there seems to be an extra `\` in the beginning of the code-block in step 6 ("Install RVM and Ruby version 2.4"), if you could take the occasion of this PR to remove it that would be great

TheAdamBorek:
@AliSoftware done

TechnicalDocuments/NewHiresCheckList.md
```diff
@@ -54,15 +52,17 @@ Here's how to get the iOS project up and running.
...
+1. Clone the iOS repository: https://github.com/Babylonpartners/babylon-ios  
+*Make sure you've already done the `Setup Git LFS` step before cloning the project.*
      ```
      git clone git@github.com:Babylonpartners/babylon-ios.git
      ```
 
 1. Install RVM and Ruby version 2.4
      ```
-     \curl -sSL https://get.rvm.io | bash -s stable
+     curl -sSL https://get.rvm.io | bash -s stable
```
mluisbrown: Why did you remove the leading `\` ? This is means it will really run `curl` and ignore any alias you might have.

TheAdamBorek: I didn't know that as I thought it's there by mistake. I believe @AliSoftware also thought so :D 
bringing it back

AliSoftware: Oh, TIL indeed.

mluisbrown: It was also TIL for me when I first saw this on the RVM site 😄

AliSoftware: @TheAdamBorek If we put it back, I thonk it would then be worth mentioning the trick using a `# The backslash ensures this command will run the real curl and not an alias` or something? as many would otherwise assume it is a typo

PR #78, opened 2019-04-03
Improve discoverability of elements in UI automation tests

andersio:


TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
```diff
@@ -0,0 +1,174 @@
+# Improve discoverability of elements in UI automation tests
+
+* Author(s): Anders Ha
+* Review Manager: David Rodrigues, Sergey Shulga
+
+## Introduction
+At the moment, there is no systematic approach to apply accessibility identifiers in Bento. Some components accept a custom accessibility identifier as a temporary measure, e.g. Chatbot UI components.
+
+## Motivation
+`String` accessibility identifier is fragile, and there were already a few breakages observed due to accessibility identifier mismatch. This is because the main codebase does not publish accessibility identifiers used, and therefore automation tests have no mean to ensure themselves always pick up the same identifier as defined in the main codebase. In other words, automation tests are prone to breakage when changes in accessibility identifiers happen.
```
ilyapuchka:
```suggestion
Using textual content of UI components rather than accessibility identifier is fragile, and there were already a few breakages observed due to strings mismatch. This is because the main codebase does not publish accessibility identifiers used, and therefore automation tests have no mean to ensure themselves always pick up the same identifier as defined in the main codebase. In other words, automation tests are prone to breakage when changes in UI happen.
```

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
```diff
@@ -0,0 +1,174 @@
...
+
+    func updateAccessibilityIdentifiers(usingPrefix prefix: String)
+}
+```
+
+When Bento is built with a Debug configuration, it should invoke `updateAccessibilityIdentifiers(usingPrefix:)` on the component root view, whenever it is bound to a given component at a given ID path for the first time.
+
+Components are free to set accessibility identifiers in whatever way they see fit, but they *must* append the given prefix to any accessibility identifier it set.
+
+Bento shall generates a prefix based on the section ID and the item ID. All variables are suffixed by `/`. So for example, the prefix for item `card` of section `activity` should be `activity/card/`.
```
ilyapuchka: I'd suggest using `_`, camel case is more common practice for naming identifiers. Unless there is a specific reason to prefer something else to `_`

andersio: `_` doesn't quite signify it being a path across three levels of entities (sections, items and "subcomponents").

ilyapuchka: do we really need that as a path? I don't mind `/`, it's just `_` would be more obvious imo

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
```diff
@@ -0,0 +1,203 @@
...
```
ilyapuchka: > Bento shall generates a prefix based on the section ID and the item ID

A lot of times these ids contain unique identifiers of the DTOs they are rendering. Considering that this can be UUIDs (and even with simpler Ints) this will make such generated ids hard to use in practice

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
```diff
@@ -0,0 +1,203 @@
...
+
+### Working with Atomic Components and Design Library
+
+Atomic components present a challenge, since composition is allowed and thus multiple components may present at the same ID path.
+
+The proposal suggests that all atomic components allow a custom _automation name_ to be specified, and implements `ViewLifecycleAware.updateAccessibilityIdentifiers(usingPrefix:)`. Composite components, that have multiple opaque child components, should propagate this call to its children if appropriate.
+
+The Design Library shall introduce a new umbrella namespace `DesignLibrary.AutomationName` for all defined automation names:
```
ilyapuchka: why calling "automation name" what is commonly called "accessibility identifier"?

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
```diff
@@ -0,0 +1,203 @@
...
+}
+```
+
+and this particular composite component called `prettyCard`:
+
+```swift
+extension DesignLibrary.ComponentsLibrary {
+    public func prettyCard() -> AnyRenderable {
+        return [
+            UI.Label(/* ... */, automationName: String(AutomationName.prettyCardNotice)),
```
ilyapuchka:
```suggestion
            UI.Label(/* ... */, automationName: AutomationName.prettyCardNotice.rawValue),
```

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
```diff
@@ -0,0 +1,203 @@
...
+
+The utility would be in the form of a `AutoPrettyEnumDescription` protocol which refines `CustomStringConvertible` with a default implementation of `description`. It would use `Mirror` internally.
+
+A caveat: There is a scenario, filed as [SR-10272](https://bugs.swift.org/browse/SR-10272), where `Mirror` does not reflect an enum correctly as expected. But it should be rare in practice.
+
+## Impact on existing codebase
+Additive changes.
+
+## Alternatives considered
+Status quo: Fragile accessibility identifier; ad hoc definition of accessibility identifier.
```
ilyapuchka: I think we can do better in finding alternative solutions, like in Swift evolution proposal, rather than just suggesting to do things in one particular way or leave it as it is.

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
```diff
@@ -0,0 +1,203 @@
...
+Printing `.item(.bento)` results in `item(SwiftPlayground.SpecialItem.bento)`, which includes the fully qualified type name of the enum as part of the associated payload. On the other hand, printing `.generic("wasabi")` results in `generic("wasabi")`, which preserves the double quote.
+
+The issue can be resolved by implementing `CustomStringConvertible` on the ID enums. But to improve the developer experience, the proposal sugggests also providing a utility printing enums in a consistent way that works for us. Specifically, these criteria should be met:
+
+* enum without associated values should be printed as `caseName`; and
+* enum with associated values should be printed as `caseName(value)`; and
+* integer and strings as associated values should be printed as is without any punctuation added; and
+* Repeated printing should always lead to the same outcome.
+
+The utility would be in the form of a `AutoPrettyEnumDescription` protocol which refines `CustomStringConvertible` with a default implementation of `description`. It would use `Mirror` internally.
```
ilyapuchka: please provide an example of the results that such utility will yield

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
```diff
@@ -0,0 +1,203 @@
...
+import BabylonHealthManagementUI
+
+extension BaseScreen {
+    // NOTE: alias in the UI testing target to help reduce verbosity.
+    typealias Name = DesignLibrary.AutomationName
+}
+
+class FeedScreen: BaseScreen {
+    func tapSubmit() {
+        let button = app[FeedRenderer.self, .activity, .card, Name.PrettyCard.submit]
```
ilyapuchka: the subscript is defined on BaseScreen, but applied here in the `app`, which is `XCUIElement`. We shouldn't be forced to make queries in the app object, as it takes more time to perform query. At least it should be still possible to do `app.buttons[...]`, ideally there shouldn't be limitation in how elements are accessed.

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
```diff
@@ -0,0 +1,203 @@
...
+
+Now that we have a systematic approach of defining and publishing information about accessibility identifiers, we can use this to reliably generate accessibility identifiers in UI automation tests!
+
+The proposal suggests that the UI automation test target should link with all UI frameworks. This allows the automation code to have access to `Renderer`s and any published `AutomationName` enums, and can therefore construct accessibility identifiers. Changes made to any section ID, item ID or automation name enums would either be automatically picked up by the automation code, or lead to compile-time breakage depending on the circusmtances.
+
+`BaseScreen` shall be extended to provide a conveinence:
+```swift
+extension BaseScreen {
+    subscript<Renderer: BoxRenderer>(
+        _ renderer: Renderer.Type,
```
ilyapuchka: why `renderer` parameter is needed? `Renderer.SectionID` is not enough for compiler?

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
ilyapuchka: I'd suggest to try implement API like this:

```swift
// root table view should have id based on type of renderer
let table = app.tables[FeedRenderer.self]

//most specific query, search for button in all cells, use section and node id as part of cell identifier
screen.cells[.activity, .card].buttons[.submit]

// same as 
table.cells["activity_card"].buttons["submit"]

// or least specific query, search among all buttons using all ids to make final identifier
screen.buttons[.activity, .card, .submit]

// same as
table.buttons["activity_card_submit"]
```

I suppose we wouldn't generally have table views in table view (and in this case inner table view should be queried in two steps any way), so subscript with maximum 3 parameters (section, node, component) should be enough.

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
ilyapuchka: One of such solutions can be instead of relying on automatically generated accessibility identifiers to require components to have accessibility identifier parameter (even if just per convention) and to have these identifiers defined manually in a name space in a renderer, so that tests code can access it easily. This way they wouldn't depend on unique identifiers as for automation it's most of the time enough to query elements by their index, otherwise they can be still queried by their textual content if specific element need to be accessed.

```swift
class FeedRenderer {
  func render(...) -> Box {
    return .empty
             |-+ Section(id: .activity)
             |---+ Node(
                     id: .card, 
                     component: prettyCard(
                       accessibilityIdentifier: AccessibilityIDs.activityCard(1)
                     )
             )
             |---+ Node(
                     id: .submit,
                     component: button(
                       accessibilityIdentifier: AccessibilityIDs.submitButton
                     )
             )
  }
}

extension FeedRenderer {
  enum AccessibilityIDs {
    static let activityCard: (Int) -> String = { "activity_card_\($0)" }
    static let submitButton = "submit"
  }
}
```

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
JasonDobo: I think both are valid, but I tend to agree with @ilyapuchka that _ would be more obvious

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
JasonDobo: If possible we should not use UUIDs are part of identifiers

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
JasonDobo: Even though the "identifier" is only developer facing and primarily used in the context of UI automation, I think calling them "accessibility identifier" is better

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
```diff
@@ -0,0 +1,203 @@
...
+
+Atomic components present a challenge, since composition is allowed and thus multiple components may present at the same ID path.
+
+The proposal suggests that all atomic components allow a custom _automation name_ to be specified, and implements `ViewLifecycleAware.updateAccessibilityIdentifiers(usingPrefix:)`. Composite components, that have multiple opaque child components, should propagate this call to its children if appropriate.
+
+The Design Library shall introduce a new umbrella namespace `DesignLibrary.AutomationName` for all defined automation names:
+
+```swift
+extension DesignLibrary {
+    public enum AutomationName {}
```
JasonDobo: Even though the "identifier" is only developer facing and primarily used in the context of UI automation, I think calling them "accessibility identifier" is better

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
```diff
@@ -0,0 +1,203 @@
...
+                )
+            ])
+        ]
+    )
+)
+```
+
+It should result in two accessibility identifiers:
+
+* `activity/card/notice` applied to the label; and
```
JasonDobo: I could be wrong but by label you mean "accessibilityLabel", we need to be careful here as the label is visible to the end user and used by VoiceOver, or did I misunderstand?

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
JasonDobo: Agreed we should not be doing app queries in the app

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
```diff
@@ -0,0 +1,203 @@
...
+```swift
+import BabylonHealthManagementUI
+
+extension BaseScreen {
+    // NOTE: alias in the UI testing target to help reduce verbosity.
+    typealias Name = DesignLibrary.AutomationName
+}
+
+class FeedScreen: BaseScreen {
+    func tapSubmit() {
```
JasonDobo: I'm probably missing something obvious, but were is FeedRenderer being defined? and how does the screen object know which renderer.self to invoke

andersio: `FeedRenderer` is defined in the UI frameworks or the main app target, which is the actual entity who maps screen states into components for rendering.

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
andersio: I mean the `UILabel`.

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
andersio: It doesn't constitute the entire accessibility identifier, only a path component of the final accessibility identifier.

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
andersio: Swift compiler does not work out the type backwards AFAIK. Even if you write:

```swift
self[FeedRenderer.SectionID.header, FeedRenderer.ItemID.headerTitle]
```

From the type system's perspective, the ID types can be used by multiple types for the same associated type location.

Also it is less verbose with the proposed subscript, which enables contextual lookup at the 2nd and 3rd arguments. 
```swift
self[FeedRenderer.self, .header, .headerTitle]
```

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
andersio: We can move the extension to `XCUIElementQuery`.

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
ilyapuchka: > it is less verbose/noisy

it's still more verbose and noisy than `self[.header, .headerTitle]`. I still feel there is a room for improvement here by associating Screen object with Renderer type (which is 1-to-1 relationship anyway in most cases, if it's not - it should be) rather than requiring to specify it in each subscript, and having a subscript like `screen[.header, .headerTitle]` where `screen` would be a helper wrapper generic over renderer type. Roughly this:

```
class FeedScreen {
  let screen = ScreenLookup<FeedRenderer>(app)
}

class ScreenLookup<Renderer: BoxRenderer> {
  subscript(_ sectionId: Renderer.SectionID, _ nodeId: Renderer.NodeId) -> XCUIElement {
    return app[...]
  }
}
```

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
andersio: Re: Two-level queries
Proposal has been updated to state why this doesn't work at all with the current Accessibility mechanism.

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
andersio: We can't go simpler than that unless we have that type information on the screen object or some kind of proxy object, like your snippet.

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
andersio: https://github.com/Babylonpartners/ios-playbook/pull/78#discussion_r276171525

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
andersio: If we have to be free from these payloads for uniqueness purpose, we would need sections and nodes to define an automation name/ID that is free of these matters.

andersio: However, if we go along this way, we might as well just have components accepting accessibility IDs.

TechnicalDocuments/Proposals/IdentifiersForUIAutomation.md
```diff
@@ -0,0 +1,219 @@
...
+
+Notice that automation names for the relevant subcomponents of `prettyCard` has been published in the `DesignLibrary.AutomationName` umbrella. This facilitates the next suggestion of this proposal...
+
+### Working with UI automation tests
+
+Now that we have a systematic approach of defining and publishing information about accessibility identifiers, we can use this to reliably generate accessibility identifiers in UI automation tests!
+
+The proposal suggests that the UI automation test target should link with all UI frameworks. This allows the automation code to have access to `Renderer`s and any published `AutomationName` enums, and can therefore construct accessibility identifiers. Changes made to any section ID, item ID or automation name enums would either be automatically picked up by the automation code, or lead to compile-time breakage depending on the circusmtances.
+
+`BaseScreen` shall be extended to provide a conveinence:
```
daniloaliberti:
```suggestion
`BaseScreen` shall be extended to provide a convenience:
```

RuiAAPeres:
Closing this for now. Please re-open when you can continue. 👍

PR #86, opened 2019-04-10
Created `Current-guide.md`

daniloaliberti:


Cookbook/Current.md
```diff
@@ -0,0 +1,22 @@
+# How to use `Current` in Babylon iOS Project
+
```
diegopetrucci: I would explain what Current is and what its purpose is first.

daniloaliberti: https://github.com/Babylonpartners/ios-playbook/blob/master/Cookbook/Proposals/ControlTheWorld.md

daniloaliberti: also this is a TBD.

diegopetrucci: Yes, but that is a proposal. This is the documentation for it -- I think it should serve as an independent and self-sufficient document explaining what Current is, what it does, how to use it, future directions, etc.

diegopetrucci: > also this is a TBD contents.
Then maybe we should have this as WIP?

daniloaliberti: I agree. But it's the beginning. It'll grow.

diegopetrucci: I feel like at the current state, and as it is public, this document is a bit too barebones to be merged.

diegopetrucci (changes requested):
I feel like at the current state, and as it is public, this document is a bit too barebones to be merged 🤔 

Maybe we could have this PR as a draft/WIP and slowly populate it?

Cookbook/Current.md
```diff
@@ -0,0 +1,22 @@
...
+### Where is that? 
+- `Current` is the global implementation of the `World` file which is located in the `BabylonDependencies.framework`.
+
+### Properties with default values
```
gtsiap: What is the value of documenting these here?

Cookbook/Current.md
```diff
@@ -0,0 +1,22 @@
...
+TBD
+
+### Properties without default values..
```
gtsiap: What is the value of documenting these here?

Cookbook/Current.md
```diff
@@ -0,0 +1,22 @@
...
+..which depends on asyncronous events to have its value
+
+TBD
+
+### Runtime dynamic changes as `DesignLibrary`
```
gtsiap: What is the value of documenting these here?
If the variable is something Reactive like `Signal`/`Property`/etc you get a strong hint 
that it will change over time.

Cookbook/Current.md
daniloaliberti: I see no problem: 
- there's no link to this document;
- it's a documentation and there's nothing wrong on the content and publishing an in progress document;
- life is asynchronous 😛

daniloaliberti: @diegopetrucci We need to speed up documentations.

Cookbook/Current.md
daniloaliberti: you will see when you need to create one here.

daniloaliberti: it's a guide

daniloaliberti:
@gtsiap basically you're asking what's the value of creating this guide, is that correct?

Cookbook/Current.md
diegopetrucci: But you still have to go through PR approval whenever you want to make a change, so why not beefing it up a bit more for the first draft?

daniloaliberti: the first idea was to create *just* the title.. and I did a lot more.

Cookbook/Current.md
daniloaliberti: > strong hint 

is not documentation.

If you feel that it's overlapping too much, I can remove it from here. 
But I feel that it's not that clear how to use stuff (Current, DesignLibrary, NVL, components). Imagine for a new joiner. 

I thought you'd feel happy with it. 

Again, don't you see value on this kind of documentation? @gtsiap

Cookbook/Current.md
daniloaliberti: ditto

Cookbook/Current.md
daniloaliberti: > I would explain what Current is and what its purpose is first.

there's a reference to the proposal at the bottom.

mluisbrown:
TBH I agree with @gtsiap and @diegopetrucci. I don't see the value in merging this document until it has at least some useful content.

The Cookbook already has a link to the `Current` proposal, which explains the concept so this is not (currently - sorry, no pun intended) adding anything useful 🤷‍♂️

daniloaliberti:
> TBH I agree with @gtsiap and @diegopetrucci. I don't see the value in merging this document until it has at least some useful content.
> 
> The Cookbook already has a link to the `Current` proposal, which explains the concept so this is not currently (sorry, no pun intended) adding anything useful 🤷‍♂️

https://github.com/Babylonpartners/ios-playbook/pull/86/files#r273911104

daniloaliberti:
I understood @diegopetrucci and @mluisbrown don't want to merge an empty document. 
But @gtsiap could you please clarify if you don't see value on this document even with the content? 
thanks

gtsiap:
A document for current is useful.
IMO we should have a document covering our architecture explaining what a viewModel, flow controller etc is. So the same applies to current. 

The only sections of this document which I don't think that provide a lot of value are `Properties with default values`, `Properties without default values..`, `Runtime dynamic changes as DesignLibrary`.

Getting this info is a matter of search,  its trivial. If we add them here we will just increase the maintenance effort that is required in order to keep this document up to date.
On there hand, providing `Dos and don't` is really beneficial because these are conventions which
a developer can't figure out just by looking the code.

for documenting API a tool like [jazzy](https://github.com/realm/jazzy) would be easier to maintain and use

Cookbook/Current.md
```diff
@@ -0,0 +1,22 @@
+# How to use `Current` in Babylon iOS Project
+
+### Where is that? 
+- `Current` is the global implementation of the `World` file which is located in the `BabylonDependencies.framework`.
```
ilyapuchka:
```suggestion
- `Current` is the global instance of the `World` type which is located in the `BabylonDependencies.framework`. It's purpose is to provide convenient access to the instances that should be shared between different parts of the app (cross-cutting concerns)
```

Cookbook/Current.md
ilyapuchka: I think it will make sense at least to mention that these properties should be observed rather than be accessed via `.value`. But for that may the next section will be already a good fit.

inamiy:
- [ ] Please mention about `Current` can only be used in `Builder` layer.

daniloaliberti:
> [ ] Please mention about `Current` can only be used in `Builder` layer.

yeah, that's the main point for this document. 👍

diegopetrucci (changes requested):
Not really sure why this is coming back as pending my review, but my considerations still stand.

RuiAAPeres:
There has been no meaningful update on this for some time. cc @daniloaliberti

Cookbook/Current.md
```diff
@@ -0,0 +1,18 @@
+# How to use `Current` in Babylon iOS Project
+
+### Where is that? 
+- `Current` is the global instance of the `World` type which is located in the `BabylonDependencies.framework`. 
+It's purpose is to provide convenient access to the instances that should be shared between different parts of the app (cross-cutting concerns). For more information please read the [How Control the World](/Cookbook/Proposals/ControlTheWorld.md) document.
```
diegopetrucci:
```suggestion
Its purpose is to provide convenient access to the instances that should be shared between different parts of the app (cross-cutting concerns). For more information please read the [How Control the World](/Cookbook/Proposals/ControlTheWorld.md) document.
```

Could we add a few examples? Just to get an idea, because this is very abstract talk and I don't think I really understand it 😅

Cookbook/Current.md
```diff
@@ -0,0 +1,18 @@
...
+
+## `Do`s and `Don't`s
+
+#### do
```
diegopetrucci:
```suggestion
#### Dos
```

Cookbook/Current.md
```diff
@@ -0,0 +1,18 @@
...
+- access `Current` **only** from `Builder`s layer;
+- pass the required dependencies down the chain;
+- contribute to it;
+
+#### don't
```
diegopetrucci:
```suggestion
#### Don't
```

Cookbook/Current.md
```diff
@@ -0,0 +1,18 @@
...
+- access `Current` from any other layer other than `Builder`. I.E all the other layers (ViewModel, Flowcontroller, ViewController, Model, BusinessController) continue to receive its dependencies throught injection, preferably via `init`;
```
diegopetrucci:
```suggestion
- access `Current` from any other layer other than `Builder`. All the other layers (`ViewModel`, `FlowController`, `ViewController`, `Model`, `BusinessController`) continue to receive their dependencies through injection when created by the builder, preferably at initialisation time;
```

Cookbook/Current.md
```diff
@@ -0,0 +1,18 @@
...
+- inject `current` as a dependency;
```
diegopetrucci:
```suggestion
- inject `Current` as a dependency;
```

Cookbook/Current.md
```diff
@@ -0,0 +1,18 @@
...
+- try to mutate `Current`. In `Release` configuration it's a constant `let`;
```
diegopetrucci: Could this be further explained? For example I didn't really follow the release issue on slack, so I have no idea why the Release config has to be a constant property.

Cookbook/Current.md
```diff
@@ -0,0 +1,18 @@
...
+### Where is that? 
+- `Current` is the global instance of the `World` type which is located in the `BabylonDependencies.framework`. 
+It's purpose is to provide convenient access to the instances that should be shared between different parts of the app (cross-cutting concerns). For more information please read the [How Control the World](/Cookbook/Proposals/ControlTheWorld.md) document.
+
+## `Do`s and `Don't`s
+
+#### do
+- access `Current` **only** from `Builder`s layer;
+- pass the required dependencies down the chain;
+- contribute to it;
```
diegopetrucci: How? What can we add to it and what should we not?

Cookbook/Current.md
```diff
@@ -0,0 +1,18 @@
...
+
+### Where is that? 
+- `Current` is the global instance of the `World` type which is located in the `BabylonDependencies.framework`. 
+It's purpose is to provide convenient access to the instances that should be shared between different parts of the app (cross-cutting concerns). For more information please read the [How Control the World](/Cookbook/Proposals/ControlTheWorld.md) document.
+
+## `Do`s and `Don't`s
+
+#### do
+- access `Current` **only** from `Builder`s layer;
+- pass the required dependencies down the chain;
```
diegopetrucci: What does this mean? Using current to access some stuff that is then injected to a VM?

diegopetrucci (commented):
Very nice progress

Cookbook/Current.md
daniloaliberti: the link explains it with loads of examples. it'll be redundant.

Cookbook/Current.md
daniloaliberti: it's intentional to be `current` because I already saw it appearing in the codebase as lowerCase.

Cookbook/Current.md
daniloaliberti: it's basically this bit: 
```swift
#if DEBUG
public var Current: World = .production
#else
public let Current: World = .production
#endif
```

We can mutate for Debug and specially for Tests. But production not at all. 
But it mutates via `setPropertyName` which changes the value of `MutableProperties` which maintains the immutability of `Current`. 
Semantics ✨ 

I'll think how that can be explained in there.
👍

Cookbook/Current.md
daniloaliberti: pass the dependencies to VM and any other layer that needs any dependency. 
Isn't that clear?

Cookbook/Current.md
daniloaliberti: just to highlight

`public var`
`public let` 

I took time to realise the difference of these 2 lines. lol

diegopetrucci (approved):
I still feel like this document is a bit barebones, or at least the title could be changed. It says "how to use current", but there are only dos and don'ts. So maybe renaming it to Current DOsDONTs could be an improvement. Then, in a later moment, when it gets expanded (are there plans to do so?) the title can be changed again.

So yes, fine to go, but I would change the title as at the moment it does not reflect the content.

Cookbook/Current.md
daniloaliberti: added the snippet

Cookbook/Current.md
daniloaliberti: I think it doesn't make sense.. once all `appDependencies` are moved into Current, no need to contribute to it. As Giorgos said, it's a matter to follow the pattern os other properties: with default value or set asynchronously.

I changed to the mutation when it's allowed. 😉

Cookbook/Technical-Documents/Current-guide.md
```diff
@@ -0,0 +1,29 @@
+# `Current` Guide - how to use it in Babylon iOS Project
+
+### Where is that? 
+- `Current` is the global instance of the `World` type which is located in the `BabylonDependencies.framework`. 
+Its purpose is to provide convenient access to the instances that should be shared between different parts of the app (cross-cutting concerns). For more information please read the [How Control the World](/Cookbook/Proposals/ControlTheWorld.md) document.
+
+## `Do`s and `Don't`s
+
+#### Dos
```
inamiy:
```suggestion
#### Do
```

inamiy (approved):
Can you also mention about when to `setupCurrent` ?
cf. https://github.com/Babylonpartners/babylon-ios/pull/7511

Or, I can add the doc later.

Cookbook/Technical-Documents/Current-guide.md
```diff
@@ -0,0 +1,29 @@
...
+- access `Current` **only** from `Builder`s layer;
+- pass the required dependencies down the chain;
+- create a customised version to fit your specific needs **for tests**;
+```swift
+let locale = Locale(identifier: "haw_US")
+Current = World(autoupdatingLocale: locale)
+```
+
+#### Don't
+- access `Current` from any other layer other than `Builder`. All the other layers (`ViewModel`, `FlowController`, `ViewController`, `Model`, `BusinessController`) continue to receive their dependencies through injection when created by the builder, preferably at initialisation time;
```
Ben-Henshall: What about decodable/encodable objects? Was there a consensus on this?

Cookbook/Technical-Documents/Current-guide.md
```diff
@@ -0,0 +1,29 @@
...
+
+## `Do`s and `Don't`s
+
+#### Dos
+- access `Current` **only** from `Builder`s layer;
+- pass the required dependencies down the chain;
+- create a customised version to fit your specific needs **for tests**;
+```swift
+let locale = Locale(identifier: "haw_US")
+Current = World(autoupdatingLocale: locale)
```
Ben-Henshall: I'm not sure about this one though. Mutating the global `Current` would cause issues in tests if running in parallel - we've seen this issue with snapshots when changing the design library theme in Current and I don't see why the same wouldn't happen when doing it this way

Ben-Henshall: (If this is correct, it should be added to the don'ts list)

Cookbook/Technical-Documents/Current-guide.md
```diff
@@ -0,0 +1,29 @@
...
+```
+
+#### Don't
+- access `Current` from any other layer other than `Builder`. All the other layers (`ViewModel`, `FlowController`, `ViewController`, `Model`, `BusinessController`) continue to receive their dependencies through injection when created by the builder, preferably at initialisation time;
+- inject `current` as a dependency;
```
Ben-Henshall:
```suggestion
- inject `Current` as a dependency. Individual parts of `Current` should be passed instead.
```

Cookbook/Technical-Documents/Current-guide.md
daniloaliberti: The consensus is: just use it on Builders. The first *don’t* mentions `Model` ... should receive its dependencies through injection.

Cookbook/Technical-Documents/Current-guide.md
daniloaliberti: Isn’t it repeating the previous? 
> .. continue to receive their dependencies through injection when created by the builder, preferably at initialisation time;

Cookbook/Technical-Documents/Current-guide.md
daniloaliberti: If it can’t be mutated then the 
```#if DEBUG
var Current```
doesn’t make sense anymore. 
It has to be fixed or a `restore` mechanism put in place..
Good point tho 👍

File: Cookbook/Technical-Documents/Current-guide.md (hunk 17 above)
Comment (Ben-Henshall) [reply Cookbook/Technical-Documents/Current-guide.md]: Yep!

File: Cookbook/Technical-Documents/Current-guide.md (hunk 16 above)
Comment (Ben-Henshall) [reply Cookbook/Technical-Documents/Current-guide.md]: Hmmm, you're right. If we make it immutable, then we're essentially back where we were but with normal DI, but without the long initialisers.

File: Cookbook/Technical-Documents/Current-guide.md [hunk 18]
```diff
@@ -0,0 +1,29 @@
...
+
+#### Dos
+- access `Current` **only** from `Builder`s layer;
+- pass the required dependencies down the chain;
+- create a customised version to fit your specific needs **for tests**;
+```swift
+let locale = Locale(identifier: "haw_US")
+Current = World(autoupdatingLocale: locale)
+```
+
```
ilyapuchka:
```suggestion
- reset the instance of `Current` to the default in the `tearDown` method
```

Cookbook/Technical-Documents/Current-guide.md
```diff
@@ -0,0 +1,29 @@
...
+### Where is that? 
+- `Current` is the global instance of the `World` type which is located in the `BabylonDependencies.framework`. 
+Its purpose is to provide convenient access to the instances that should be shared between different parts of the app (cross-cutting concerns). For more information please read the [How Control the World](/Cookbook/Proposals/ControlTheWorld.md) document.
+
+## `Do`s and `Don't`s
+
+#### Dos
+- access `Current` **only** from `Builder`s layer;
+- pass the required dependencies down the chain;
+- create a customised version to fit your specific needs **for tests**;
```
ilyapuchka:
```suggestion
- create a customized version to fit your specific needs **for tests** and set it in the test `setUp` method
```

Cookbook/Technical-Documents/Current-guide.md
daniloaliberti: @Ben-Henshall I think @ilyapuchka 's suggestion here addresses your concerns about the `Current` mutations..
```- reset the instance of `Current` to the default in the `tearDown` method```

Given that, I think we can resolve this now right?

Comment (ilyapuchka) [reply Cookbook/Technical-Documents/Current-guide.md]: Yes, I think it can be one of the reasons we can have instability in tests. I'm more convinced in that than that it is mutating `Current` in the first place - tests that are running in parallel should be running in separate processes thus not sharing a memory, and tests running in one process should be running sequentially, otherwise, I don't know how parallelization works in general 🤷‍♂

File: Cookbook/Technical-Documents/Current-guide.md [hunk 20]
```diff
@@ -0,0 +1,29 @@
+# `Current` Guide - how to use it in Babylon iOS Project
+
+### Where is that? 
+- `Current` is the global instance of the `World` type which is located in the `BabylonDependencies.framework`. 
+Its purpose is to provide convenient access to the instances that should be shared between different parts of the app (cross-cutting concerns). For more information please read the [How Control the World](/Cookbook/Proposals/ControlTheWorld.md) document.
```
diegopetrucci: I was re reading this and I think we could improve this by saying what goes into Current and what the client can expect from it. Right now there is a link to the proposal, but the proposal might be outdated and not up to our current standards. So if this is the document that explains how to use Current it should also provide a tiny window into its contents.

Cookbook/Technical-Documents/Current-guide.md
```diff
@@ -0,0 +1,29 @@
+# `Current` Guide - how to use it in Babylon iOS Project
+
+### Where is that?
```
diegopetrucci: What is `that`? Could this be rephrased?

Cookbook/Technical-Documents/Current-guide.md
```diff
@@ -0,0 +1,29 @@
+# `Current` Guide - how to use it in Babylon iOS Project
+
+### Where is that? 
+- `Current` is the global instance of the `World` type which is located in the `BabylonDependencies.framework`. 
+Its purpose is to provide convenient access to the instances that should be shared between different parts of the app (cross-cutting concerns). For more information please read the [How Control the World](/Cookbook/Proposals/ControlTheWorld.md) document.
+
+## `Do`s and `Don't`s
+
+#### Do
+- access `Current` **only** from topmost `AppDelegate` until `Builder`s layer
```
diegopetrucci:
```suggestion
- You should only access `Current` in builders, not in ViewModels or Renderers. The AppDelegate is included
```

Cookbook/Technical-Documents/Current-guide.md
```diff
@@ -0,0 +1,29 @@
...
+- pass the required dependencies down the chain;
```
diegopetrucci:
```suggestion
- Pass the required dependencies down the chain. The builder accesses Current and it injects the specific dependencies down the chain
```

Cookbook/Technical-Documents/Current-guide.md
```diff
@@ -0,0 +1,29 @@
...
+- create a customized version to fit your specific needs **for tests** and set it in the test `setUp` method
```
diegopetrucci:
```suggestion
- In your tests, you are free to create a customized version of Current that fits your specific needs. Set it up in the `setUp` method. After you're done, reset it in `tearDown`. Eg
```

Cookbook/Technical-Documents/Current-guide.md
```diff
@@ -0,0 +1,29 @@
...
+```swift
+let locale = Locale(identifier: "haw_US")
+Current = World(autoupdatingLocale: locale)
+```
+- reset the instance of `Current` to the default in the `tearDown` method
```
diegopetrucci:
```suggestion
```

Cookbook/Technical-Documents/Current-guide.md
```diff
@@ -0,0 +1,29 @@
...
+#### Don't
+- access `Current` from any other layer other than `Builder`. All the other layers (`ViewModel`, `FlowController`, `ViewController`, `Model`, `BusinessController`) continue to receive their dependencies through injection when created by the builder, preferably at initialisation time;
+- inject `current` as a dependency;
```
diegopetrucci:
```suggestion
```

Cookbook/Technical-Documents/Current-guide.md
```diff
@@ -0,0 +1,29 @@
...
+- try to mutate `Current`. In `Release` configuration it's a constant `let`;
```
diegopetrucci:
```suggestion
- try to mutate `Current` (excluding Tests). In `Release` configuration it's a constant `let`;
```

diegopetrucci (dismissed):

Cookbook/Technical-Documents/Current-guide.md
daniloaliberti: `that` is `Current`, the subject of the previous line, which I prefer to avoid to repeat.

diegopetrucci:
```suggestion
### Where can I find it?
```

diegopetrucci: Also this section feels more like an overview that just a where to find the file. Maybe something like 

```suggestion
### Overview
```

daniloaliberti: I accepted this but I don't agree using first or second person. 
I've checked and I don't see any of the Cookbook using this way to explain. (expect for the Olivier's about framework).

Cookbook/Technical-Documents/Current-guide.md
daniloaliberti: It's repeating "down the chain" twice.

daniloaliberti: also as items I prefer to have them as 
- lowercase

Cookbook/Technical-Documents/Current-guide.md
daniloaliberti: this is very important. It shouldn't be removed.

diegopetrucci: But it’s already stated in the previous point

Cookbook/Technical-Documents/Current-guide.md
diegopetrucci: Items in a list are usually uppercase https://www.quickanddirtytips.com/education/grammar/formatting-vertical-lists

Cookbook/Technical-Documents/Current-guide.md
daniloaliberti: 1.
> what goes into Current

Isn't this part clear? 
> convenient access to the instances that should be shared between different parts of the app

2.
> what the client can expect from it
I don't understand it. 
Again:
> ...access to the instances that should be shared...

3.
> might be outdated and 
the part that is outdated has been pointed by @inamiy  in [this comment](https://github.com/Babylonpartners/ios-playbook/pull/86#pullrequestreview-243068580) . I'm thinking how to cover that.

> not up to our current standards

I don't understand it. 

4. 
> it should also provide a tiny window into its contents
IMO pointing where to find is already enough to the Engineer find it's contents. But considering the tiny window, there is one already:
```swift 
let locale = Locale(identifier: "haw_US")
Current = World(autoupdatingLocale: locale)
```
I'm assuming (by the sequence of the Cookbook) that who's reading it, has already read the Proposal already mentioned. So no need to repeat.

Cookbook/Technical-Documents/Current-guide.md
daniloaliberti: - lowercase, and not in first person
- the **not in** is part of the `Don't` section

Cookbook/Technical-Documents/Current-guide.md
RuiAAPeres: List items AFAIK are uppercase. 🤔

diegopetrucci:
I have dropped my request for changes because I think we have fundamental fundamentally different approaches on how we should structure this document, and there is very little to do unless a third party intervenes. So my comments are left but as suggestions, not requests.

Cookbook/Technical-Documents/Current-guide.md
daniloaliberti: @diegopetrucci from the link you provided I see a capitalised list 👍 :

Screen Shot 2019-06-11 at 14 50 59

But giving a quick read... I clicked the link Chicago Manual Style... https://www.quickanddirtytips.com/education/grammar/grammar-style-issues

> One book discusses capitalization, another partially discusses punctuation, and so on. Only the [Chicago Manual of Style](https://www.quickanddirtytips.com/education/grammar/grammar-style-issues) gives the topic the respect it deserves. 

... I see a lowercased list 

Screen Shot 2019-06-11 at 14 52 53

I change it as long it adds value to the document and improve the understanding.

Cookbook/Technical-Documents/Current-guide.md
daniloaliberti: it shouldn't be removed. I prefer the clarity of the separation **do** and **don't**

Cookbook/Technical-Documents/Current-guide.md
daniloaliberti: it shouldn't be removed. I prefer the clarity of the separation do and don't

Cookbook/Technical-Documents/Current-guide.md
diegopetrucci: My problem here is that it still feels part of the previous step, as we are talking about how to use current in tests.

daniloaliberti: You're correct. It is indeed part of the previous step. 
This commit (https://github.com/Babylonpartners/ios-playbook/pull/86/commits/27b1f6b10e55e9504ba3d68bb3ae95912dda7b56) I addressed that specific `setUp` and `tearDown` methods changes. 
Thanks.

Cookbook/Technical-Documents/Current-guide.md
```diff
@@ -0,0 +1,36 @@
...
+### Where can I find it?
+- `Current` is the global instance of the `World` type which is located in the `BabylonDependencies.framework`. 
+Its purpose is to provide convenient access to the instances that should be shared between different parts of the app (cross-cutting concerns). For more information please read the [How Control the World](/Cookbook/Proposals/ControlTheWorld.md) document.
+
+## `Do`s and `Don't`s
+
+#### Do
+- access `Current` **only** from topmost `AppDelegate` until `Builder`s layer
+- pass the required dependencies down the chain;
+- create a customized version to fit your specific needs **for tests** and set it in the test `setUp` method. Please reset it on the `tearDown` method
```
TheAdamBorek: I think this is no longer the right approach since we have started to run unit tests in parallel. Am I right?

RuiAAPeres:
Ping @daniloaliberti

Cookbook/Technical-Documents/Current-guide.md
daniloaliberti: We have parallel **and** serial tests running. Imagine the tests `a`, `b`, and `c` in a process and `d`, `e` and `f` in another.. 

When `a` finishes, it's important to reset `Current` in `tearDown`. 
@TheAdamBorek ^
@AliSoftware 
@ilyapuchka

daniloaliberti:
> Can you also mention about when to `setupCurrent` ?
> cf. [Babylonpartners/babylon-ios#7511](https://github.com/Babylonpartners/babylon-ios/pull/7511)
> 
> Or, I can add the doc later.

Done here https://github.com/Babylonpartners/ios-playbook/pull/86/commits/5fea05aa4866965b5fb52b0fbc9227811620a62f

daniloaliberti:
@gtsiap @ilyapuchka @TheAdamBorek Could you guys please have a second read on this please? 
I believe I've addressed/responded you. Thanks.

Ben-Henshall (approved):

gtsiap (approved):

Cookbook/Technical-Documents/Current-guide.md
AliSoftware: I confirm what @daniloaliberti said, that's also my understanding of how they are distributed across sims. Parallel testing doesn't mean you shouldn't reset the state.

Cookbook/Technical-Documents/Current-guide.md
AliSoftware: @AliSoftware Reminder to self: make a Danger rule for this.

Cookbook/Technical-Documents/Current-guide.md
```diff
@@ -0,0 +1,40 @@
...
+}
+
+override func tearDown() {
+    Current = .production // which returns a new `World()`
+}
+```
+#### Don't
+- access `Current` from any other layer other than `Builder`. All the other layers (`ViewModel`, `FlowController`, `ViewController`, `Model`, `BusinessController`) continue to receive their dependencies through injection when created by the builder, preferably at initialisation time;
+- inject `current` as a dependency;
+- try to mutate `Current` (excluding Tests). In `Release` configuration it's a constant `let`;
```
AliSoftware: Should we remove the `Current.changeTheme` method from `World` then?

Is there a difference in that sentence between "mutating the content of `Current`" and "replace the `Current` instance with another one"? iow, are we both prohibited from doing `Current = someNewValue` but also `Current.changeWhatever` – assuming `changeWhatever` mutates some properties of the `Current` instance?

I think this point deserves clarification in this sentence to disambiguate between the two cases.

AliSoftware:
Super small nits, just if you have the occasion (otherwise don't bother it's not that important): 

 - some markdown editors require that you leave a blank line after a `#`/`##`/... title to consider them a title (GFM is lenient about that but not all editors are) so I like to have them so that the doc is compatible with all markdown editors
- Also, could be nice to have a link to that new document at the end of the `ControlTheWorld.md` doc

Cookbook/Technical-Documents/Current-guide.md
daniloaliberti: Perfect point @AliSoftware 

There is a bunch of `set` methods:
`setBiometricAuth`, `setNetworkDependencies`, `setEnvironment`, `setAppConfiguration` and also as you said the `changeTheme`. 

These methods change **values** inside Current's private mutable properties that are consumed by its public immutable computed ones. Basically value semantics over mutability.

Answering your question: both are prohibited. The only entity allowed to run these `set` methods is the SharedAppDelegate.

I wonder if changing the:
> (excluding Tests)

to 

> (excluding Test targets and `SharedAppDelegate.setupCurrent` methods)

Would it disambiguate that point?

> - try to mutate `Current` (excluding Test targets and `SharedAppDelegate.setupCurrent` methods). In `Release` configuration it's a constant `let`;

AliSoftware: I think it would be useful to specify `excluding test targets and setupCurrent` indeed, but I don't think that would be enough to clarify that both "Replacing the Current instance altogether" and "mutating the content of Current" are prohibited elsewhere.

I mean if you look at the code you can eventually see that `World` is a `struct` so has value semantics and thus when it's `let` (in Release) you won't be able to mutate even its content, so you can _guess_ and _conclude_ that indeed both should not be allowed anyway, but making it explicit in such a document dedicated to `Current` still feels useful to me.

```suggestion
- Mutate `Current`'s properties – or re-affect it to another instance – after `SharedAppDelegate.setupCurrent()`. You are only allowed to mutate `Current` in tests, not in app. In `Release` configuration it's a constant `let` anyway to prevent mutation;
```

daniloaliberti:
> * some markdown editors require that you leave a blank line after a `#`/`##`/... title to consider them a title...

All titles in all levels have ` ` after them. 

Regarding the proposal link, it's already linked when mentioned in the first section. But I've added in the end of the file https://github.com/Babylonpartners/ios-playbook/pull/86/commits/7705fa8292a061b825acf93cc205244b0ff7b2b9

AliSoftware:
> All titles in all levels have ` ` after them.

@daniloaliberti Sorry for not being clear, I was talking about a blank newline after the titles, so between the line containing the title and the first paragraph of the section – and not a whitespace between `###` and the title text 😛 (Not to worry, that was a nit anyway)

> Regarding the proposal link, it's already linked when mentioned in the first section. But I've added in the end of the file 7705fa8

I was talking about the other way around, adding a link in `ControlTheWorld.md` to link to your new `Current-Guide.md` document 😅  to that both reference each other and people only reading `ControlTheWorld.md` know the new follow-up document you just added exists too 😉

daniloaliberti:
@AliSoftware ah got them all now. I'll take care of that in the next change. 👍 thanks man.

PR #87, opened 2019-04-11
Add QA-specific question to the Cookbook

NSMyself:
This PR stems from the fact that at the moment I am uncomfortable reviewing QA PRs since I am not really familiarized with their syntax. It's not the "When/Given/And/etc" format itself, but its content instead. 
Right now I do not know how to evaluate its correctness and would love know how to.

As such, I added a question to the Cookbook (no link, since there's no answer) and am hoping someone can provide this answer. This is in line with what I believe was mentioned in previous meetings.


RuiAAPeres:
Can you add a small emoji (✏️ ) to show that it needs to be written? Like:

```
* [✏️ ] What is the When / Given / Then/ And formatting used by QA and how can i determine it's correctness
```

NSMyself:
Of course @RuiAAPeres. Updating...

JasonDobo:
@NSMyself The syntax your asking about is called "Gherkin" it was originally used for 'Selenium" and has become the default sync for writing test scenarios in many companies. on iOS we used a POD called "XCTest-Gherkin" to enable support for Gherkin. 
https://docs.cucumber.io/gherkin/reference/

NSMyself:
Thanks @JasonDobo, wasn't aware of that 👍 i'll have a look at the documentation. In any case, how can I confirm that the PR changes are good to go?

JasonDobo:
On a connected note, I defined a while ago how step definitions should be written and the way they describe actions in our tests. @RuiAAPeres should I add this to the playbook?

NSMyself:
All done @RuiAAPeres

RuiAAPeres:
@JasonDobo yes. For now this PR can be merged as is tho.

PR #93, opened 2019-04-11
Automate most of the new-hire onboarding checklist

pr-babylon:
- Pull most commands into a single shell script that only requires one
    argument (a target directory to clone the repo into).

  - Update the instructions to avoid duplicating installation
    instructions for third-party tools -- instead, link directly to
    their documentation so we don't have to worry about keeping the
    specific installation instructions up-to-date.

### DIFF
--- a/Cookbook/Technical-Documents/NewHiresCheckList.md
+++ b/Cookbook/Technical-Documents/NewHiresCheckList.md
@@ -1,6 +1,6 @@
 # New Hires Checklist
 
-## GitHub access
+## GitHub Access
 
 Prior to starting, make sure you have a Babylon GitHub account and that you have access to the following repositories:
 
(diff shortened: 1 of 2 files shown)


Cookbook/Technical-Documents/NewHiresCheckList.md
```diff
@@ -42,52 +42,23 @@ As an iOS Engineer, you should be in the following Slack channels:
...
-
-1. Clone the iOS repository: https://github.com/Babylonpartners/babylon-ios  
-*Make sure you've already done the `Setup Git LFS` step before cloning the project.*
-     ```
-     git clone git@github.com:Babylonpartners/babylon-ios.git
-     ```
-
-1. Get our latest Xcode templates by running `./Templates/install_xcode_templates.sh`
-
-1. Install RVM and Ruby version 2.4
```
inamiy: We can preserve RVM + Ruby 2.4 install manual for better tutorial, though RVM is not always necessary.

Cookbook/Technical-Documents/NewHiresCheckList.md
```diff
@@ -42,52 +42,23 @@ As an iOS Engineer, you should be in the following Slack channels:
 
 Make sure you also join your Tribe/Squad's Slack channels.
 
-## Setup your development environment
+## Install prerequisites.
```
inamiy:
```suggestion
## Install prerequisites
```

Cookbook/Technical-Documents/NewHiresCheckList.md
```diff
@@ -42,52 +42,23 @@ As an iOS Engineer, you should be in the following Slack channels:
...
 
-Here's how to get the iOS project up and running.
+  - Xcode + its command-line tools
+  - [Homebrew](https://brew.sh/)
+  - Ruby 2.4 (you can use [RVM](https://rvm.io/) to install it and manage your Ruby versions)
+
+## Set up your development environment.
```
inamiy:
```suggestion
## Setup Guide
```

Just for simple header :)

scripts/ios-onboarding
```diff
@@ -0,0 +1,52 @@
...
+
+# Install project dependencies (Ruby gems and Cocoapods pods).
+bundle install
+bundle exec pod install
+
+set +x
+echo
+echo "The babylon-ios project is now ready for local development on your system."
+echo "It has been set up in: ${target_directory}"
+echo "Run 'xed .' inside that directory to get started!"
```
inamiy: Nice script 👏

scripts/ios-onboarding
```diff
@@ -0,0 +1,52 @@
...
+#
+#   $ CODE_DIRECTORY=/Users/me/code scripts/ios-onboarding
+#
+# See the "New Hires Checklist" in the ios-playbook repo for more details.
+
+set -eo pipefail  # exit immediately when any command fails
+set -o nounset    # treat unset variables as an error
+set -x            # echo every command before running it
+
+target_directory="${CODE_DIRECTORY}/babylon-ios"
```
inamiy: Please take care if `CODE_DIRECTORY` is empty.

inamiy: Never mind, `set -o nounset` will solve :)

Cookbook/Technical-Documents/NewHiresCheckList.md
mluisbrown: I agree with @inamiy. IMO we should keep the Ruby + RVM installation instructions here because it's not immediately obvious from the RVM site what you should do.

Also, https://rvm.io is currently down for me, but the `get.rvm.io` script is not.

pr-babylon: Good point. I added back the instructions.

pr-babylon:
Thanks for the feedback. Updated in response to comments and rebased. Also added `gem install bundler` to the script based on the recent edit to the `NewHiresChecklist` yesterday.

scripts/ios-onboarding
```diff
@@ -0,0 +1,53 @@
...
+if [[ ! -d "${target_directory}" ]]; then
+  git clone git@github.com:Babylonpartners/babylon-ios.git "${target_directory}"
+fi
+cd "${target_directory}"
+
+# Get the latest Xcode templates.
+./Templates/install_xcode_templates.sh
+
+# Install project dependencies (Ruby gems and Cocoapods pods).
+gem install bundler
```
mattbrictson-babylon: This will install the latest version of bundler which is 2.x, but the project currently requires 1.16.4.

Perhaps this is over-engineered, but to install the appropriate version you could do:

```
gem install bundler -v `ruby -e 'puts IO.read("Gemfile.lock")[/BUNDLED WITH\n   (\S+)$/, 1]'`
```

mattbrictson-babylon (commented):
Nice improvements! ✨

scripts/ios-onboarding
pr-babylon: Good catch. It worked on my machine since I ended up with both Bundler 1 and Bundler 2 installed, but it wouldn't work on someone's machine if they didn't have Ruby/Bundler before.

I set it to just install the latest Bundler 1.

inamiy (approved):

Cookbook/Technical-Documents/NewHiresCheckList.md
```diff
@@ -42,53 +42,32 @@ As an iOS Engineer, you should be in the following Slack channels:
 
 Make sure you also join your Tribe/Squad's Slack channels.
 
-## Setup your development environment
+## Install prerequisites
 
-Here's how to get the iOS project up and running.
+  - Xcode + its command-line tools
```
chitrakotwani:
```suggestion
  - Xcode 10.1 + it's command-line tools
```

Ben-Henshall (approved):

Cookbook/Technical-Documents/NewHiresCheckList.md
diegopetrucci: I think the original was correct.

diegopetrucci (approved):

Cookbook/Technical-Documents/NewHiresCheckList.md
mluisbrown: I agree. First `its` should be without an apostrophe, regardless. Also, I don't think we should mention a specific Xcode version number. The version number issue is only a temporary one when a new Xcode version is released.

What *might* be a good thing to mention is to install Xcode from the Apple Developer Downloads page and **not** from the Mac App Store, to avoid unexpected automatic updates.

pr-babylon: @mluisbrown Agreed all around. I hit merge just as you were posting this; I'll open a PR to add that note.

PR #98, opened 2019-04-15
[Proposal] Mealy-Machine ReactiveFeedback

inamiy:
This proposal focuses on following improvements in `ReactiveFeedback`.

1. Allow `Feedback` to also handle "incoming event" (not just from "current state" only)
2. Distinguish between "external event" and "feedback"
3. Remove duplicated `state` caching in both `SignalProducer` and `Property`

This is a companion proposal of https://github.com/Babylonpartners/ios-playbook/pull/102 .

### References

- https://github.com/Babylonpartners/ReactiveFeedback/pull/32
- https://github.com/Babylonpartners/ReactiveFeedback/pull/38
- https://github.com/Babylonpartners/ReactiveFeedback/pull/41 (previous rough implementation)
- https://github.com/Babylonpartners/ios-playbook/pull/44

### DIFF
--- a/Cookbook/Proposals/MealyReactiveFeedback.md
+++ b/Cookbook/Proposals/MealyReactiveFeedback.md
@@ -104,6 +104,19 @@ Furthermore, by separating "external event" nicely, we can send "initial externa
          ...
      }
  }
+
++// NOTE: We won't have `SignalProducer.system` using `scan` anymore,
++// but instead we use `Property.producer`.
+-extension SignalProducer where Error == NoError {
+-    public static func system<Event>(
+-        initial: Value,
+-        scheduler: Scheduler = QueueScheduler.main,
+-        reduce: @escaping (Value, Event) -> Value,
+-        feedbacks: [Feedback<Value, Event>]
+-    ) -> SignalProducer<Value, NoError> {
+-        ...
+-    }
+-}
 ```
 
 - `Feedback` preserves the streaming format, i.e. `Signal<???> -> Signal<Event>` rather than `??? -> Signal<Event>` so that it can have flattening capability e.g. `flatMap(.latest)` or `flatMap(.first)`.
@@ -112,7 +125,9 @@ Furthermore, by separating "external event" nicely, we can send "initial externa
     - Resolves "idle -> routing -> idle" navigation state transition dance.
 - **Backward compatible** with current Moore machine (we can still add as many states for precise control as we want)
 
-### Usage
+### Usage Examples
+
+#### Example 1: Simple (counter, logging, analytics event)
 
 ```swift
 let externalEvent = Signal.merge(
@@ -120,17 +135,134 @@ let externalEvent = Signal.merge(
     decrement.map { Event.decrement }
 )
 
+let loggingFeedback = Feedback { signal -> Signal<Event> in
+    return signal
+        .on(value: { event, state in // side-effect
+            // NOTE: Now we can also retrieve `event` as well as `state`!
+            print("receiving event = \(event) with state = \(state)")
+        })
+        .ignoreValues() // no feedback loop
+        .promoteValue() // for adjusting type
+}
+
+let analyticsFeedback = Feedback { signal -> Signal<Event> in
+    return signal
+        .filter { event, _ in event == .increment }
+        .on(value: { event, _ in // side-effect
+            // NOTE: Analytics is one of the example that doesn't need state changes at all.
+            analytics.send(event: Analytics.Event.didTapIncrement)
+        })
+        .ignoreValues() // no feedback loop
+        .promoteValue() // for adjusting type
+}
+
 let system = Property(
     initial: 0,
     reduce: counterReducer,
     events: externalEvent,
     feedbacks: [loggingFeedback, analyticsFeedback]
 )
-
-/// We can simply create SignalProducer from this system,
-/// without creating producer first then bind to property.
-let p = system.producer
 ```
+
+To simplify the duplicated code above, see `Feedback.noFeedback` in [ios-playbook#102](https://github.com/Babylonpartners/ios-playbook/pull/102).
+
+#### Example 2: Navigation
+
+```swift
+let externalEvent = Signal.merge(
+    didTapA.map { Event.route(Route.pushA) },
+    didTapB.map { Event.route(Route.modalB) }
+)
+
+let navigationFeedback = Feedback { signal -> Signal<Event> in
+    return signal
+        // NOTE: Using `flatMap(.first)` prevents from routing to multiple screens simultaneously
+        .flatMap(.first) { event, _ in
+            guard case let route = event.route else { return .empty }
+            return flowController.handle(route)
+        }
+        .ignoreValues() // no feedback loop
+        .promoteValue() // for adjusting type
+}
+
+let system = Property(
+    initial: 0,
+    reduce: reducer,
+    events: externalEvent,
+    feedbacks: [navigationFeedback]
+)
+```
+
+For navigation feedback rule and `flatMap(.first)`, see also [Babylonpartners/ios-playbook#44](https://github.com/Babylonpartners/ios-playbook/pull/44).
+This is a handy FRP way of managing "isTransitioning" state outside of `State`.
+
+Because this routing is now event-driven rather than state-driven, **we don't need to transit the state back to the original position after completed, which will reduce the intermediate `case`s e.g. `State.isPushingA` and `Event.didFinishPushingA`**.
+(But note that we might still want to have these cases if we want to precisely handle tem to avoid interfering with other events/states/feedbacks)
+
+#### Example 3: Image Loading
+
+```swift
+enum Event {
+    case callAPI(Request)
+    case fetchImages(Response)
+    case reloadCell(image: UIImage, index: Int)
+}
+
+struct State {
+    var images: [/* cellIndex */ Int: UIImage?] = [:]
+}
+
+let externalEvent = Signal.merge(
+    viewDidLoad.map { Event.callAPI(Request(apiURL)) }
+}
+
+let apiFeedback = Feedback { signal -> Signal<Event> in
+    return signal
+        .flatMap(.first) { event, _ in
+            guard case let .callAPI(request) = event else { return .empty }
+            return apiSession.send(request)
+        }
+        .map(Event.fetchImages)
+}
+
+let imageFeedback = Feedback { signal -> Signal<Event> in
+    return signal
+        .flatMap(.merge) { event, _ -> Signal<(UIImage?, index: Int)> in
+            guard case let .fetchImages(response) = event else { return .empty }
+            return imageLoader.fetchAllImages(response.imageURLs)
+        }
+        .map { image, index in
+            Event.reloadCell(image: image, index: index)
+        }
+}
+
+let reducer = { state, event in
+    guard case let .reloadCell(image, index) = event else { return state }
+
+    var state = state
+    state.images[index] = image
+    return state
+}
+
+let system = Property(
+    initial: State(),
+    reduce: reducer,
+    events: externalEvent,
+    feedbacks: [apiFeedback, imageFeedback]
+)
+
+system.signal
+    .combinePrevious(initial: system.value)
+    .startWithValues { oldState, newState in
+        // NOTE: diff & patch algorithm must be efficient enough to update images only
+        let patch = diff(oldState, newState)
+        realView.apply(patch: patch)
+    }
+```
+
+This is a typical image loading example that changes state for every cell's image being ready.
+
+**Caveat:** Diff & patch algorithm must be efficient enough to update images only. Otherwise, we need to rely on primitive FRP e.g. `ReactiveSwift.Property` which will become tightly coupled with virtual view (See [Babylonpartners/Bento#144](https://github.com/Babylonpartners/Bento/issues/144) and [Slack](https://babylonhealth.slack.com/archives/G9S9L0TEK/p1554731131006300))
 
 ## Impact on existing codebase
 
--- a/Cookbook/Proposals/MealyReactiveFeedback.md
+++ b/Cookbook/Proposals/MealyReactiveFeedback.md
@@ -1,0 +1,141 @@
+# Mealy-Machine ReactiveFeedback
+
+* Author: Yasuhiro Inami
+* Review Manager: TBD
+
+## Introduction
+
+This proposal focuses on following improvements in `ReactiveFeedback`.
+
+1. Allow `Feedback` to also handle "incoming event" (not just from "current state" only)
+2. Distinguish between "external event" and "feedback"
+3. Remove duplicated `state` caching in both `SignalProducer` and `Property`
+
+## Motivation
+
+### 1. Allow `Feedback` to also handle "incoming event"
+
+Current `ReactiveFeedback` (as of ver 0.5.0) has `Feedback` type which represents side-effect and event feedback loop for a particular or entire system:
+
+```swift
+public struct Feedback<State, Event> {
+    let events: (Scheduler, Signal<State, NoError>) -> Signal<Event, NoError>
+}
+```
+
+(For simplicity, we will ignore `Scheduler` for this entire discussion)
+
+This kind of structure is so-called [Moore machine](https://en.wikipedia.org/wiki/Moore_machine) where feedback (output function) only takes `State` (or stream of `State`s) as argument and produces side-effects with accompanying new `Event` (or stream of `Event`s).
+
+Compared to this approach, current frontend architecture normally uses [Mealy machine](https://en.wikipedia.org/wiki/Mealy_machine), or [Queue automaton](https://en.wikipedia.org/wiki/Queue_automaton) (feedback model) instead, where reducer (transition function) takes both `State` and `Event`, then produces a new `State` and `Output` (with feedback).
+
+<details>
+<summary>Examples of Mealy machine (Click to open)</summary>
+
+For example, [Elm](https://package.elm-lang.org/packages/elm-lang/core/latest/Platform) uses `update: Event -> State -> (State, Cmd<Event>)` as reducer.
+Note that this single function can be splitted into two functions:
+
+1. `Event -> State -> State`, and
+2. `Event -> State -> Cmd<Event>`
+
+which we often call 1. reducer and 2. feedback.
+
+[Redux](https://redux.js.org) also provides similar mechanism called `Middleware`, where `middleware: (store: Store) -> (next: Dispatch) -> Action -> Any`.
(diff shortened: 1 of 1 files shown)


RuiAAPeres:
There's a strong theoretical focus in this proposal, which is nice, but it lacks the necessary examples to make it through. Can you please add "real" examples, instead of the `increment`/`decrement` snipet? 

Regarding the points you mentioned in the beginning of this proposal:

> Allow `Feedback` to also handle "incoming event" (not just from "current state" only). 

✅ 


> Distinguish between "external event" and "feedback"

✅ 

> Remove duplicated state caching in both SignalProducer and Property

❓ 

I see you have a 3rd section mentioning this point, but I couldn't see from the explanation how you remove it. 😞 

Given that this proposal attemps to superseed #44, it would be good for it to provide examples of such situation.

inamiy:
> Remove duplicated state caching in both SignalProducer and Property

This means we won't have `SignalProducer.system` using `scan` anymore, but instead we use `systemProperty.producer`.

> it would be good for it to provide examples of such situation.

Yes, I will add more examples and update proposal :)

Cookbook/Proposals/MealyReactiveFeedback.md
```diff
@@ -0,0 +1,141 @@
...
+
+Please note that **both Moore and Mealy machines are theoretically convertible to each other** , so it is still reasonable to keep using current Moore-based machine.
+
+However, from practical point of view, Mealy machine has several advantages over Moore machine:
+
+- Requires less states
+    - This is a general conclusion of Mealy machine compared to Moore because we will add more graph-edges (transition patterns) than graph-nodes (states).
+- `Event`-recognizable `Feedback`
+    - For simple side-effects such as **logging**, **sending analytics events**, or somtimes **routing** or **image-loading** can be handled just from incoming events without precisely handling their states inside `ReactiveFeedback`
+    - This will remove verbose "idle -> routing -> idle" navigation state transition dance, and improves the feedback-based-navigation idea in [Babylonpartners/ios-playbook#44](https://github.com/Babylonpartners/ios-playbook/pull/44)
```
Ben-Henshall: >This will remove verbose "idle -> routing -> idle" navigation state transition dance

❤️ ❤️ ❤️

Ben-Henshall (commented):
This seems very promising, but it almost exclusively affects the ReactiveFeedback library. Shouldn't the proposal be posted there instead?

inamiy:
@Ben-Henshall 
This affects the whole team as discussed in https://github.com/Babylonpartners/ReactiveFeedback/pull/41#pullrequestreview-224973268 , so I think we can start discussion here first.

Ben-Henshall (approved):

Cookbook/Proposals/MealyReactiveFeedback.md
```diff
@@ -0,0 +1,141 @@
...
+)
+
+/// We can simply create SignalProducer from this system,
+/// without creating producer first then bind to property.
+let p = system.producer
+```
+
+## Impact on existing codebase
+
+This proposal will be a **breaking change**, but modification will be relatively **small** because current state structures can be fully preserved (though it can be minimized).
```
sergdort: Not sure why is it a breaking change?

Cookbook/Proposals/MealyReactiveFeedback.md
```diff
@@ -0,0 +1,141 @@
...
+```swift
+let externalEvent = Signal.merge(
+    increment.map { Event.increment },
+    decrement.map { Event.decrement }
+)
+
+let system = Property(
+    initial: 0,
+    reduce: counterReducer,
+    events: externalEvent,
```
sergdort: This is good 👍  Now we want need to use `Feedback.input`

Cookbook/Proposals/MealyReactiveFeedback.md
inamiy: It's because `Property.init` type signature changes.

inamiy:
e3ff332 Updated examples.

Cookbook/Proposals/MealyReactiveFeedback.md
```diff
@@ -0,0 +1,273 @@
...
+
+#### Example 1: Simple (counter, logging, analytics event)
+
+```swift
+let externalEvent = Signal.merge(
+    increment.map { Event.increment },
+    decrement.map { Event.decrement }
+)
+
+let loggingFeedback = Feedback { signal -> Signal<Event> in
```
ilyapuchka: can you please add some type annotations to make it clear what `signal` is? I assume it is `Signal<(Event, State), NoError>`?

Cookbook/Proposals/MealyReactiveFeedback.md
ilyapuchka: I'd personally prefer `input`, or `eventInput` as a name for this parameter, to disambiguate it with events that are being produced inside the system itself.

I can see some similarities with disturbance (`Gd(s)`) in the control theory, if we assume the disturbance is the signal of events which is summed up with a signal of events in the system itself (Gc(s))

Screenshot 2019-04-22 at 17 10 09
https://ocw.mit.edu/courses/mechanical-engineering/2-017j-design-of-electromechanical-robotic-systems-fall-2009/lecture-notes/MIT2_017JF09_control.pdf

Cookbook/Proposals/MealyReactiveFeedback.md
```diff
@@ -0,0 +1,273 @@
...
+    reduce: reducer,
+    events: externalEvent,
+    feedbacks: [navigationFeedback]
+)
+```
+
+For navigation feedback rule and `flatMap(.first)`, see also [Babylonpartners/ios-playbook#44](https://github.com/Babylonpartners/ios-playbook/pull/44).
+This is a handy FRP way of managing "isTransitioning" state outside of `State`.
+
+Because this routing is now event-driven rather than state-driven, **we don't need to transit the state back to the original position after completed, which will reduce the intermediate `case`s e.g. `State.isPushingA` and `Event.didFinishPushingA`**.
```
ilyapuchka: will we then still have states like `loading`?

Cookbook/Proposals/MealyReactiveFeedback.md
```diff
@@ -0,0 +1,273 @@
...
+(But note that we might still want to have these cases if we want to precisely handle tem to avoid interfering with other events/states/feedbacks)
+
+#### Example 3: Image Loading
```
ilyapuchka: I'd put this example as first as this describes the most common type of feedbacks we use.

ilyapuchka: How about an example where we have to call api to display some data and while this network call is being called we need to display a loading indicator? Without a state change how the reducer will know what it needs to render?

ilyapuchka (commented):
I agree with @Ben-Henshall that this should be discussed in the scope of ReactiveFeedback project rather than here (as well as probably other RAF related proposal)

The proposal also mentions that suggested type of state machine is convertible with what we have now and that it is reasonable to still use exiting approach. Is it a suggestion to allow having both? How they will coexist? How we will decide what to use when?

Cookbook/Proposals/MealyReactiveFeedback.md
ilyapuchka: IMO this is not just about getting rid of these intermediate states, it's about conceptually different way of reasoning about the state. IMO with this we will give up having the whole picture of state.

Comparing with approach that we used before, where routes (navigation signal) were derived from actions (events) rather than state and was kind of out of the feedback system, this change will effectively make it feel like it is _in_ the feedback system because it is installed as a feedback in a feedback loop rather than uses a separate subscription. But one could say that even with a separate subscription routes were still _in_ the feedback system because events were only produced by the system itself. So effectively to me it feels like it is going back to the previous way of defining routes but using a different syntax.

TheAdamBorek:
This proposal looks very nice as It would solve few issues we have currently with RAF (especially where to put analytics and how to handle routing).

However, what I lack is the context of why we haven't used the `Mealy machine` from the beginning. Is there any reason behind that why we prefered `(State) -> Event` instead `(State, Event) -> Event`? @sergdort ?

Cookbook/Proposals/MealyReactiveFeedback.md
```diff
@@ -0,0 +1,273 @@
...
+        .ignoreValues() // no feedback loop
+        .promoteValue() // for adjusting type
+}
+
+let analyticsFeedback = Feedback { signal -> Signal<Event> in
+    return signal
+        .filter { event, _ in event == .increment }
+        .on(value: { event, _ in // side-effect
+            // NOTE: Analytics is one of the example that doesn't need state changes at all.
+            analytics.send(event: Analytics.Event.didTapIncrement)
```
TheAdamBorek: ❤️ 
as you have already mentioned, a helper would be nice to have here :)

Cookbook/Proposals/MealyReactiveFeedback.md
TheAdamBorek: I think `loading` is a different case as this very often represents the state actually (the fact that you need to show a loading indicator)

ilyapuchka: In the same manner "showing" something is a state IMO, the difference is that it does not change how the screen looks. State does not necessary needs to map 1-to-1 to specific renderer change.

Cookbook/Proposals/MealyReactiveFeedback.md
inamiy: > it feels like it is going back to the previous way of defining routes but using a different syntax.

This is right, though I'm not sure about the history of previous approach (how it didn't work well) since I wasn't there at that time.

My understanding is that #44 will introduce FRP's `flatMap(.first)` to also manage navigation state _indirectly_.
Since its FRP-based state can filter simultaneous multiple navigations, having extra navigating-states will mostly be redundant.

ilyapuchka: > how it didn't work well

It worked fine, the problem was more theoretical than technical wether it's correct to derive routes from actions or from state.

Cookbook/Proposals/MealyReactiveFeedback.md
inamiy: @ilyapuchka https://github.com/Babylonpartners/ios-playbook/pull/98#discussion_r277337817

I'm not familiar with control theory (which I believe is an interesting topic with regard to FRP and RAF), but as far as glancing its basics, I think:

- `R(s)` is the input (we are currently calling `externalEvent`)
- `H(s)` is the feedback
- `Gd(s)` is a noise (we probably don't need to care)

In contrast to FRP, external events and feedback events can't be combined together with `+`/`-`.

Cookbook/Proposals/MealyReactiveFeedback.md
```diff
@@ -0,0 +1,273 @@
...
+
+```swift
+let externalEvent = Signal.merge(
+    increment.map { Event.increment },
+    decrement.map { Event.decrement }
+)
+
+let loggingFeedback = Feedback { signal -> Signal<Event> in
+    return signal
+        .on(value: { event, state in // side-effect
```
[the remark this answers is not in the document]
inamiy: > `event` and `state` will be exactly the same which were previously passed into reducer?

For backward compatibility, **after passing `reducer`**,`Feedback`'s signal argument will emit:

1. incoming event
2. **`reducer`-modified** state
    - Ideally, this should be **the state before passing `reducer`** for simplicity

- [ ] TODO: Add documentation for above discussion

Cookbook/Proposals/MealyReactiveFeedback.md
inamiy: Any states are worth adding if we don't want to replay the same flow with multiple same events given.

Cookbook/Proposals/MealyReactiveFeedback.md
inamiy: > I'd put this example as first as this describes the most common type of feedbacks we use.

I want to demonstrate simple case first, so IMO analytics example fits in first place.

> How about an example where we have to call api to display some data and while this network call is being called we need to display a loading indicator?

This will be the same "state-driven" approach as we are currently doing now.

- [ ] TODO: Add documentation

Cookbook/Proposals/MealyReactiveFeedback.md
inamiy: > Any states are worth adding if we don't want to replay the same flow with multiple same events given.

Exception comes when `Feedback` uses `flatMap(.first)`.
This will mostly behave the same as "if we don't want to replay the same flow with multiple same events given" without adding extra state in `RAF` side.

Cookbook/Proposals/MealyReactiveFeedback.md
inamiy: I'm also +1 with `input` over `events`, so I will update the code :)

Cookbook/Proposals/MealyReactiveFeedback.md
```diff
@@ -0,0 +1,273 @@
...
+    initial: State(),
+    reduce: reducer,
+    events: externalEvent,
+    feedbacks: [apiFeedback, imageFeedback]
+)
+
+system.signal
+    .combinePrevious(initial: system.value)
+    .startWithValues { oldState, newState in
+        // NOTE: diff & patch algorithm must be efficient enough to update images only
```
andersio: A remark on efficiency.

In our world of pure `virtual component_tree = f(state)`,  the virtual component tree would contain the image, and the diff-patching is done by Bento.

Assuming there is an arbitrarily large number of items, my educated assumption is that the diff-patching Bento is unlikely our most prominent bottleneck. The linear cost of the virtual component tree recomputation, against many state change occurrences cased by small updates, is more likely to first be the issue. Most importantly, this is inherently the cons of the pure model we have, and perfect diff-patching does not help alleviate this in any manner.

Cookbook/Proposals/MealyReactiveFeedback.md
```diff
@@ -0,0 +1,273 @@
...
+
+### Changes
+
+```diff
+ public struct Feedback<State, Event> {
+-    let events: (Scheduler, Signal<State, NoError>) -> Signal<Event, NoError>
++    let events: (Scheduler, Signal<(Event, State), NoError>) -> Signal<Event, NoError>
+ }
+
+ extension Property {
```
andersio: Not sure if it is valuable to have `Property<(State, Event)>`.

andersio (changes requested):
I am in favour of this. A few comments on specific matters though.

Cookbook/Proposals/MealyReactiveFeedback.md
andersio: It might also be better to refactor this example closer to what would actually happen in our codebase, if it would still be relevant afterwards to the topic at hand.

e.g. `Action.itemWillAppear` reports item appearance to VM, where the VM initiates the network request. Similarly, VM responds to `Action.itemDidDisappear` by cancelling requests or purging stuff as needed.

mr-v (commented):
Looks very promising.

It would be nice to have a bigger example, e.g. whole `ViewModel` implementation to give a better idea how this would affect our code and highlight differences with current approach.
I realise we're about to have a meeting about this, so it's just for future consideration. Full example/guide would also help when working with this approach day-to-day.

Cookbook/Proposals/MealyReactiveFeedback.md
inamiy: `Property<State>` stays as is, so it won't be `Property<(State, Event)>`.

Cookbook/Proposals/MealyReactiveFeedback.md
```diff
@@ -0,0 +1,273 @@
...
+     public convenience init<Event>(
+         initial: Value,
+         scheduler: Scheduler = QueueScheduler.main,
+         reduce: @escaping (Value, Event) -> Value,
++        events: Signal<Event>,   // external events
```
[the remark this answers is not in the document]
inamiy: This is an input argument that will replace `Feedback.input` which uses hot `Signal.pipe()`.

For initial dispatch from cold signal, we can perhaps adopt the same idea as Elm that provides [`init : ( model, Cmd msg )`](https://package.elm-lang.org/packages/elm-lang/html/1.1.0/Html-App#program) with initial side-effect being given as yet another extra parameter (but not as feedback).

For example:

```swift
let initialProducer: SignalProducer<Event> = ...
let externalEvents: Signal<Event> = ...

self.state = Property(initial: (initialValue, initialProducer), events: externalEvents, ...)
```

But I omitted this idea because we can actually rewrite it using ReactiveSwift as follows (which is a bit cumbersome though):

```swift
let pipe = Signal.pipe()
self.state = Property(initial: initialValue, events: Signal.merge(externalEvents, pipe.output), ...)

pipe.input <~ initialProducer
```

Cookbook/Proposals/MealyReactiveFeedback.md
```diff
@@ -0,0 +1,273 @@
...
+    didTapA.map { Event.route(Route.pushA) },
+    didTapB.map { Event.route(Route.modalB) }
+)
+
+let navigationFeedback = Feedback { signal -> Signal<Event> in
+    return signal
+        // NOTE: Using `flatMap(.first)` prevents from routing to multiple screens simultaneously
+        .flatMap(.first) { event, _ in
+            guard case let route = event.route else { return .empty }
+            return flowController.handle(route)
```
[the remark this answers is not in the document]
inamiy: Yes, #44 is more responsible for navigation discussion.

Cookbook/Proposals/MealyReactiveFeedback.md
inamiy: Thanks for detailed explanation here.
I have to admit that this example is rather my thought experiment (for Bento without `Reactive.Property` issue) that might not work well in practice.
Though I believe handling view lifecycles e.g. `itemWillAppear` are still possible, managing every remote image rendering states may be boring and not really interested, as Elm or React are not even doing this kind of micro-management AFAIK.

I think I should remove this example.

RuiAAPeres:

inamiy:
This PR is a supporting proposal of [#44 Feedback-based Navigation](https://github.com/Babylonpartners/ios-playbook/pull/44) to extend the idea of handling any kinds of side-effect inside ReactiveFeedback, and ease its state-driven-only difficulties by also having `Event` as an argument.

So, if #44 won't be accepted and we decide to implement side-effects in plain `ReactiveSwift.Signal` as is (hybrid style), that will also be an architectural choice that this PR will no longer have a strong point to adopt. /cc @sergdort

inamiy:
FYI Check out this new Moore-Mealy-hybrid machine improvements 🤖

[Refactor code by adding \`Effect\`, \`EffectQueue\`, \`Feedback\` by inamiy · Pull Request \#12 · inamiy/ReactiveAutomaton](https://github.com/inamiy/ReactiveAutomaton/pull/12)

This also supports `Feedback` system seamlessly.

PR #101, opened 2019-04-17
Add trailing commas section to style guide

diegopetrucci:


zzcgumn (approved):

AliSoftware (approved):
Approved since looking at our existing codebase, that's indeed the rule we have applied so far in most places.

RuiAAPeres (approved):
Can we extend this to other things besides Arrays (e.g. dictionaries)?

diegopetrucci:
@RuiAAPeres the swift lint rule fixes dictionaries too, so might as well https://github.com/realm/SwiftLint/blob/master/Rules.md#trailing-comma

Cookbook/Style-guide/README.md
```diff
@@ -958,6 +959,28 @@ In larger expressions, optional parentheses can sometimes make code read more cl
 let playerMark = (player == current ? "X" : "O")
 ```
 
+## Commas
+
+Trailing commas in arrays and dictionaries should be avoided.
```
mluisbrown: I fundamentally disagree with this.

mluisbrown (changes requested):
I fundamentally disagree with removing trailing commas. They are a very useful feature to allow easy refactoring.
I'm not suggesting we should enforce them everywhere, but I see no reason at all to force their removal.

Cookbook/Style-guide/README.md
RuiAAPeres: Thanks for your opinion. 👍 :trollface:

ilyapuchka: I double @mluisbrown on that

ilyapuchka (changes requested):
I don't see a strong motivation to forbid (considering our style guide gets translated into SwiftLint rules that will treat all warnings as errors) trailing commas, which are very useful when editing arrays.

ilyapuchka:
I'd like to see this proposal with motivation for it. The fact that we are not putting trailing commas in every array or dictionary we define does not mean that doing so should be a violation of a style guide, and the other way around.
Trailing commas has a huge impact on refactoring experience, for example when rearranging elements or adding new one. It's very easy to forget to add comma and very annoying to remove them in one place and add them in another.

TheAdamBorek:
I don't see any benefits in having this rule. If it is in the code I rather ignore it when I'm reading the code.

Imo this is one of rules where we don't need to listen to majority and if someone is against then ... well ¯\_(ツ)_/¯ maybe it's not worth it.

I'll say again, I don't think it is going to change readablity of our codebase

diegopetrucci:
I will close this for now, and if we have strong feelings about it in the future we can re-open it.

andersio:
Trailing commas do not add any noise that is already there, and enforcing not having it does not appear to have any substantial benefit.

Cookbook/Style-guide/README.md
sergdort: I love trailing comas, I forget the comas every time I add new stuff and off course it fails to build

PR #106, opened 2019-04-25
[Proposal] Standardise Setup of Models in Tests

mr-v:
Goal of this proposal is to standardise and simplify test setup when it comes to models. At the moment we already have a few patterns in place to do this and it's not always clear which one to use.

Another goal related to this is for setup code to contain only values relevant to the tested logic. This would improve experience of setting up state in tests and maintenance later on.

More details in the [Standardise Setup of Models in Tests](https://github.com/Babylonpartners/ios-playbook/blob/witold/model_fixtures/Cookbook/Proposals/ModelFixtures.md) proposal.

* [x] I will send a meeting invitation, using this [template](Template_Proposal_Meeting_Invitation.MD), scheduled for 2 weeks after this proposal is made, so an agreement can be reached.
* [x] **By creating this proposal, I understand that it might not be accepted**. I also agree that, if it's accepted, depending on its complexity, I might be requested to give a workshop to the rest of the team. 🚀

### DIFF
--- a/Cookbook/Proposals/ModelFixtures.md
+++ b/Cookbook/Proposals/ModelFixtures.md
@@ -41,7 +41,7 @@ In our codebase there are multiple implementations of this pattern with variatio
 
 On top of that some tests create value type by directly calling the initializer.
 
-Those inconsistencies cause friction when trying to setup state in tests. It's hard to discover existing test helpers. Usage of unparametrized factory methods (e.g. `MockAddressDTO.make()`) makes it hard to understand on what data the test depends.
+These inconsistencies cause friction when trying to setup state in tests. It's hard to discover existing test helpers. Usage of unparametrized factory methods (e.g. `MockAddressDTO.make()`) makes it hard to understand on what data the test depends.
 
 ## Proposed solution
 
--- a/Cookbook/Proposals/ModelFixtures.md
+++ b/Cookbook/Proposals/ModelFixtures.md
@@ -1,0 +1,249 @@
+# Standardise Setup of Models in Tests
+
+* Author: Witold Skibniewski
+* Review Manager: TBD
+
+## Introduction
+
+Goal of this proposal is to standardise and simplify test setup when it comes to models. At the moment we already have a few patterns in place to do this and it's not always clear which one to use.
+
+Another goal related to this is for setup code to contain only values relevant to the tested logic. This would improve experience of setting up state in tests and maintenance later on.
+
+## Motivation
+
+<!-- // This section should the answer the `why?`. -->
+
+<!-- Describe the problems that this proposal seeks to address. If the
+problem is that some common pattern is currently hard to express, show
+how one can currently get a similar effect and describe its
+drawbacks. If it's completely new functionality that cannot be
+emulated, motivate why this new functionality would help Swift
+developers create better Swift code. -->
+
+Well-structured test code has a few distinct sections. One way to split them is to use Arrange, Act, Assert pattern. This proposal focuses on improving the initial Arrange stage of the test. First improvment is to make writing setup part as easy as possible. Second is to make this code easy to understand to help with readability and maintanance.
+
+So far the most used pattern for setting up values in tests is to add enum for namespacing and introduce `make` test helper method. Enumeration's name follows "`MockSomeNameDTO`" pattern, so for example tests  call `MockPatientDetailsDTO.make()`. There are a few issues with this approach:
+
+- Values passed to model's initializer are hidden inside the helper method and you don’t have any knowledge and control over them.
+- `Mock` prefix doesn’t really fit models. Most often models are just data with some limited behaviour. Naming of the namespace is misleading and to work with value types you don't really need mocks.
+> The term "mock" has a precise definition as well, as a specific sub-type of test double. A mock object asserts that certain invocations are made on itself, and will raise exception as soon as any unexpected interactions take place. [[source](https://github.com/testdouble/contributing-tests/wiki/Mock#precise-use)]
+
+- Naming of namespacing enums is inconsistent e.g. `MockPatientDetailsDTO` for `PatientDTO`.
+- Sometimes there are multiple `make` helpers to maintain. See more details about issues with this in "Object Mother" section in "Altenratives considered".
+
+In our codebase there are multiple implementations of this pattern with variations in naming and functionality, e.g.
+
+- `MessageTestData.makeMessages`,
+- `MessageTestData.makeMessageResponse`,
+- `MockPatientDetailsDTO.makeWithPartner`,
+- `MockAddressDTO.make`,
+- `CoreSessionDTO.mock`.
+
+On top of that some tests create value type by directly calling the initializer.
+
+Those inconsistencies cause friction when trying to setup state in tests. It's hard to discover existing test helpers. Usage of unparametrized factory methods (e.g. `MockAddressDTO.make()`) makes it hard to understand on what data the test depends.
+
+## Proposed solution
+
+<!-- // This section should the answer the `how?`. -->
+<!-- Describe your solution to the problem. Provide examples and describe
+how they work. Show how your solution is better than current
+workarounds: is it cleaner, safer, or more efficient? -->
+
+Proposed approach is similar to what we often do by the convention with `MockSomeNameDTO.make` pattern. It aims to improve current situation by introducing a few important rules.
+
+1. Helpers for models should be defined as an **extension of the model type** in a test target or framework with test utilities.
+- Helper method should be named `fixture` to signify that it's part of the tests.
+- Helpers should use parameters that mirror model's initializer.
+- All of helper method parameters should have default value. This will make relevant data stand out and irrelevant data will be provided by defaults.
+
+Let's see sample implementation. First define extension on an existing type, instead of introducing extra type for namespacing.
+
+```swift
+// Example for points 1. and .2
+
+struct Foo {
+    let a: Int
+    let b: String
+}
+
+// add extension in test target or test utilities framework:
+
+extension Foo {
+    static func fixture() -> Foo {
+        return Foo(a: 1, b: "xyz")
+    }
+}
+```
+
+Building upon this, use the same arguments in `fixture` factory method as in the models's initializer. Provide a default value for each argument. Those values should be "reasonable defaults", e.g. for identifier it can be `"1"` value. Example of adjusted `Foo.fixture` method:
+
+```swift
+// Example for points 3. and .4
+
+extension Foo {
+    static func fixture(a: Int = 1, x: String = "xyz") -> Foo {
+        return Foo(a: a, x: x)
+    }
+}
+```
+
+This provides necessary customization point and allows to skip setup of irrelevant data. The values used to put test subject in the initial state stand out more thanks to this and make test easier to understand.
+
+If the type of argument is another model, you can use its `.fixture()` helper to provide default value.
+
+```swift
+struct Bar {
+    let foo: Foo
+    let c: String
+}
+
+extension Bar {
+    static fixture(foo: Foo = .fixture(), c: String = "any") -> Bar {
+        return Bar(foo: foo, j: c)
+    }
+}
+```
+
+Following example illustrates how those helpers can nicely compose and how only single value of `x` argument is customized (and thus relevant to the test code that would follow).
+
+```swift
+let bar = Bar.fixture(foo: .fixture(x: "value relevant to test"))
+```
+
+---
+
+Benefits of proposed solution:
+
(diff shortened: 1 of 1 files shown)


gtsiap:
Can you also include in your proposal how we should deal with situations like these?
```
public enum MockPatientDetailsDTO {
    public static func makeWithPartner() -> PatientDTO { ... }
    public static func makeUpdated() -> PatientDTO { ... }
    public static func makeMinor() -> PatientDTO { ... }
    public static func makeEphemeralMinor() -> PatientDTO { ... }
}
```

There are cases like in subscriptions where we need multiple mocks for monthly, yearly etc.

ilyapuchka (approved):
Great call!
I'd say that the important part of this standardisation will be to have all the models, at least which are available in the core framework, to start with, to have fixtures defined in the test utilities framework. The question stays for types from feature frameworks that should be also accessible in the tests of other targets (i.e. we can have appointment type fixture in the appointments framework, but how do we use it in the app target unit tests if we need it)

diegopetrucci (approved):
I love this. The only thing that we could add is how to deal in cases where an object might have multiple fixtures — maybe an enumeration a-la `.fixture(yourCase: Enum)`?

Cookbook/Proposals/ModelFixtures.md
```diff
@@ -0,0 +1,249 @@
...
+    public static func makeWithPartner() -> PatientDTO { ... }
+    public static func makeUpdated() -> PatientDTO { ... }
+    public static func makeMinor() -> PatientDTO { ... }
+    public static func makeEphemeralMinor() -> PatientDTO { ... }
+}
+```
+
+This kind of setup obscures what's going on and it's very hard to tell which data is changed to achieve end result. It's also impractical when we'd like to use similar data to one returned from existing helper, but with slight variation. We can't adjust the returned value, since our models have immutable properties. We could tweak data by adding parameters to those methods, but then we would end up with approach similar to proposed `fixture` helper. Model creation for tests would still be scattered across multiple places. Other solution would be to add another factory method on Object Mother, but this would get even less maintenable with time.
+
+If we encounter case where there are many instances of passing the same parameters to the `fixture` method, then we could consider adding methods like this on case by case basis, rather than making it a default.
```
ilyapuchka: @gtsiap @diegopetrucci I guess this should answer your questions.

So we would have something like this:

```
extension SomeDTO {
    func makeSomeFixture() -> SomeDTO
    func makeAnotherFixture() -> SomeDTO
}
```

inamiy (approved):
+1 👍 

IMO I would like to see this proposal elevated into use of Sourcery to codegen all fixtures.

Cookbook/Proposals/ModelFixtures.md
```diff
@@ -0,0 +1,249 @@
...
+}
+```
+
+Building upon this, use the same arguments in `fixture` factory method as in the models's initializer. Provide a default value for each argument. Those values should be "reasonable defaults", e.g. for identifier it can be `"1"` value. Example of adjusted `Foo.fixture` method:
+
+```swift
+// Example for points 3. and .4
+
+extension Foo {
+    static func fixture(a: Int = 1, x: String = "xyz") -> Foo {
```
TheAdamBorek: 👍 
just btw. I used to call it `static func fake(...)`.
Just to consider if we like the name more.
However, I'm fine with `fixture`

TheAdamBorek (approved):
This is the way I used to set up fake data previously :)
"You have my sword"!

Cookbook/Proposals/ModelFixtures.md
chitrakotwani: As much as I love the approach and proposal, am not a huge fan of a fancy name `fixture`.  Could we make it something more obvious like `Foo.testObject`?

chitrakotwani: (Sorry for unsolicited review! Don't know what the system is around that) 😬

ilyapuchka: @TheAdamBorek  @chitrakotwani I'd be personally fine with either `fake` or `testObject`, but I guess it's a matter of habit. Strictly speaking though it is not necessary a `fake` double, because it can be actually used by tested code, if we use Fouler's blog post as a reference for classification. And fixtures he mentioned [here](https://www.martinfowler.com/articles/mocksArentStubs.html#FixtureSetup) but he does not mention this as a type of test doubles (he does mention Object Mother pattern). It's something that I used to see in other languages like Ruby and JavaScript so seems like a common term for this kind of stuff. I'd even suggest to use the same name instead of our `canned responses`, which is another weird name in fact.

ilyapuchka:
If we do step 2 of that and use Sourcery I'd be happy to assist 🙇  It's sometimes an effort to update all the "stubs" and "mocks" we have when we add new properties and Xcode is usually not very helpful with error messages.

Cookbook/Proposals/ModelFixtures.md
```diff
@@ -0,0 +1,249 @@
...
+
+<!-- Describe the problems that this proposal seeks to address. If the
+problem is that some common pattern is currently hard to express, show
+how one can currently get a similar effect and describe its
+drawbacks. If it's completely new functionality that cannot be
+emulated, motivate why this new functionality would help Swift
+developers create better Swift code. -->
+
+Well-structured test code has a few distinct sections. One way to split them is to use Arrange, Act, Assert pattern. This proposal focuses on improving the initial Arrange stage of the test. First improvment is to make writing setup part as easy as possible. Second is to make this code easy to understand to help with readability and maintanance.
+
```
zzcgumn: Typo: `improvment` -> `improvement`.

Cookbook/Proposals/ModelFixtures.md
```diff
@@ -0,0 +1,249 @@
...
+In our codebase there are multiple implementations of this pattern with variations in naming and functionality, e.g.
+
+- `MessageTestData.makeMessages`,
+- `MessageTestData.makeMessageResponse`,
+- `MockPatientDetailsDTO.makeWithPartner`,
+- `MockAddressDTO.make`,
+- `CoreSessionDTO.mock`.
+
+On top of that some tests create value type by directly calling the initializer.
```
zzcgumn:
```
Further inconsistencies are added by tests that initialise value types directly.
```

reads more easily in my opinion.

Cookbook/Proposals/ModelFixtures.md
```diff
@@ -0,0 +1,249 @@
...
+
+Those inconsistencies cause friction when trying to setup state in tests. It's hard to discover existing test helpers. Usage of unparametrized factory methods (e.g. `MockAddressDTO.make()`) makes it hard to understand on what data the test depends.
```
zzcgumn: `Those` -> `These`.

zzcgumn:
I am in favour of adding this pattern and naming convention for creating test values. I am not sure how feasible it would be to try rename test double automatically. We do have fair few `Mock...` test doubles that actually are mocks and should not be renamed.

sergdort (approved):
tenor-211133427

Cookbook/Proposals/ModelFixtures.md
mr-v: Thanks Martin for suggestions! 👍

mr-v:
> Can you also include in your proposal how we should deal with situations like these?
> 
> ```swift
> public enum MockPatientDetailsDTO {
>     public static func makeWithPartner() -> PatientDTO { ... }
>     public static func makeUpdated() -> PatientDTO { ... }
>     public static func makeMinor() -> PatientDTO { ... }
>     public static func makeEphemeralMinor() -> PatientDTO { ... }
> }
> ```
> 
> There are cases like in subscriptions where we need multiple mocks for monthly, yearly etc.

@gtsiap @diegopetrucci I think last part of [Proposed solution](https://github.com/Babylonpartners/ios-playbook/blob/e4c2538e400ad0610e8cca447d4d92ebda708151/Cookbook/Proposals/ModelFixtures.md#proposed-solution) touches on this. Goal is to compose existing helpers, so in many cases you won't need to do anything more that in example below. There are so many parameters, that I might have missed something, but I think it illustrates the general idea:

```swift
let minorPatient = PatientDTO.fixture(
    subscription: PatientSubscriptionDTO.fixture(patientIdentifier: PatientID.minor)
    isMinor: true
)

// if this is something that is needed many times in different test cases,
// then it could be extracted into a helper

let minorPatient = PatientDTO.fixtureMinor()

```

There's also this bit in section about [Object Mother](https://github.com/Babylonpartners/ios-playbook/blob/e4c2538e400ad0610e8cca447d4d92ebda708151/Cookbook/Proposals/ModelFixtures.md#use-object-mother-pattern-to-setup-test-data).

> If we encounter case where there are many instances of passing the same parameters to the fixture method, then we could consider adding methods like this on case by case basis, rather than making it a default.

If proposal is accepted, one of the actions will be to add documentation, so I'll make sure to include examples like this in there.

mr-v:
> I'd say that the important part of this standardisation will be to have all the models, at least which are available in the core framework, to start with, to have fixtures defined in the test utilities framework.

I think the steps would be:

1. Convert existing helpers.
2. Move to standardised location if they're not there already.
3. Add new helpers on an as-needed basis.

> The question stays for types from **feature frameworks** that should be also accessible in the tests of other targets (i.e. we can have appointment type fixture in the appointments framework, but how do we use it in the app target unit tests if we need it)

We could use `BabylonSnapshotTestsUtilities` for it, which is de facto the place for feature frameworks-related test utilities. We discussed renaming this project a few times, we could do it to lessen the confusion.

mr-v:
> IMO I would like to see this proposal elevated into use of Sourcery to codegen all fixtures.

@inamiy I'd say that introducing a codegen tool is a separate proposal. I think it could nicely compliment what I propose here, e.g. it would ease some maintenance burden in the long run.

However, this is not required to make this proposal, so I'd definitely look at it separately.
Another point is that we still need to figure out some points, e.g. what values to use for default arguments. I'm sure there will be other learnings as well, once we start using it. I think that only after this manual phase we will be properly equipped to think about using codegen for this.

Cookbook/Proposals/ModelFixtures.md
mr-v: Regarding the naming of method issue. Options I see:

- keep `fixture`,
- add `test` prefix, i.e. `testFixture`.

I don't see that anybody is strongly against keeping the original name. It should also be simple enough to rename it afterwards if we find it problematic after adoption.

> It's something that I used to see in other languages like Ruby and JavaScript so seems like a common term for this kind of stuff.

I used the name after Ilya's suggestion as it nicely expresses the purpose of helper. Note about "fixture" term from xUnit Patterns:

> In generic xUnit a test fixture is all the things we need to have in place in order to run a test and expect a particular outcome.

More details:
- [test fixture (in xUnit)](http://xunitpatterns.com/test%20fixture%20-%20xUnit.html),
- [test fixture (disambiguation)](http://xunitpatterns.com/test%20fixture%20-%20ambiguous.html).

---

> just btw. I used to call it static func fake(...) just because I used to use Faker gem in my ruby tests 😅

@TheAdamBorek Most of the time those are value types, so they're real values, only configured for tests. I think `fake` would give wrong impression.

> As much as I love the approach and proposal, am not a huge fan of a fancy name fixture. Could we make it something more obvious like Foo.testObject?

The `Object` could be misleading, because most of the times our models are value types.

> (Sorry for unsolicited review! Don't know what the system is around that) 😬

@chitrakotwani  All feedback is welcome! Thanks for reviewing. Ideally everyone should review. The more, the better! For example I might be too used to this approach, so it helps me with understanding other perspectives and see what needs documenting.

MichalTKwiecien (approved):

viorel15 (approved):

daniloaliberti (approved):
great one! ✨

mr-v:
# Actions

## Documentation

Add documentation to playbook:
- keep it more focused than proposal, more guideline-like,
- code examples with helpers composition, repeating setup,
- explain the fixture naming.

## Implementation

1. Split migration into following steps:
- MockPatientDetailsDTO.make (first, as it has the most occurences),
- update types matching `Mock\w+\DTO\.\w+` pattern,
- any other inconsistencies mentioned in proposal.

2. Adapt convention in Pull Requests for new types.

3. Ideally also update old helper once you used it (if it wasn't covered in the migration phase above).

Things to keep track of during migration:
- How much data we need to customise in tests.
- What kind of values make sense for default argument. For example I'm leaning towards using obviously irrelevant values, e.g. for `String` it could be `"Any"` or `"Irrelevant data"`. This would help with people not relying on hidden "realistic" data and for example force to provide something more realistic if it's needed in for example snapshot test.

## Future
- Generate the helpers with for example Sourcery.

# Other

- This proposal is only concerned with setting up data for tests. Verifying behaviour with mocks and stubs is a separate concern.

PR #109, opened 2019-04-25
Adds XcodeTips to provide steps for fixing common errors

jadams-babylon:


Cookbook/Technical-Documents/XcodeTips.md
```diff
@@ -0,0 +1,11 @@
+## Tips and tricks
+
+### Errors when pulling code/switching branches
+Various errors can occur when pulling code/switching branches. Typically all that is needed is running `pod install` and closing and reopening the project in XCode. The following directions are intended to be a more thorough reset to deal with bigger changes.
+
+- Clean Build: Product > Clean (shortcut: cmd + shift + k)
+- Close Xcode
+- Update bundle: run `bundle install`
+- Update pods: `pod repo update && pod deintegrate && pod install`
```
ilyapuchka: We have git hooks that are checking if pods need to be updated after switching to another branch or pulling/merging (in `/githooks`). We can as well add it to check for bundler to be in sync

```suggestion
- Update pods: `pod deintegrate && pod install --repo-update`
```

Cookbook/Technical-Documents/XcodeTips.md
```diff
@@ -0,0 +1,11 @@
+## Tips and tricks
+
+### Errors when pulling code/switching branches
+Various errors can occur when pulling code/switching branches. Typically all that is needed is running `pod install` and closing and reopening the project in XCode. The following directions are intended to be a more thorough reset to deal with bigger changes.
```
ilyapuchka:
```suggestion
Various errors can occur when pulling code/switching branches. Typically all that is needed is running `pod install` and closing and reopening the project in Xcode. The following directions are intended to be a more thorough reset to deal with bigger changes.
```

Cookbook/README.md
```diff
@@ -40,3 +40,4 @@ iOS Cookbook 👩‍🍳
 
 * [✏️] What is Gherkin (the When / Given / Then/ And formatting used by QA) and how can i determine it's correctness in QA PRs
 * [Fastlane Match](./Technical-Documents/FastlaneMatch.md)
+* [XCode Tips & Tricks](./Technical-Documents/XcodeTips.md)
```
ilyapuchka:
```suggestion
* [Xcode Tips & Tricks](./Technical-Documents/XcodeTips.md)
```

Cookbook/Technical-Documents/XcodeTips.md
```diff
@@ -0,0 +1,11 @@
...
+
+- Clean Build: Product > Clean (shortcut: cmd + shift + k)
+- Close Xcode
+- Update bundle: run `bundle install`
+- Update pods: `pod repo update && pod deintegrate && pod install`
+- Clean DerivedData: can be found at ~/Library/Developer/Xcode/DerivedData and/or babylon-ios/DerivedData
```
ilyapuchka: I doubt it will ever be in `babylon-ios/DerivedData` unless some one configured it for themselves. Anyway `Option-Cmd-Shift-K` should do the job (I always run it with `Cmd-Shift-K`)

ghost (approved):
looks good to me.

Cookbook/Technical-Documents/XcodeTips.md
inamiy: This is _very_ important ☝️

Cookbook/Technical-Documents/XcodeTips.md
inamiy: Don't we need to `bundle exec` ?

jadams-babylon: Good call. I think it depends on one's setup. I don't need it on my machine, but adding bundle exec should be the safer option and work regardless. I will go ahead and add it.

Cookbook/Technical-Documents/XcodeTips.md
jadams-babylon: I've been seeing it pop up in both locations for some reason. Still not sure why (don't think I did anything to configure as such, and it still populates normally in ~Library/Developer/Xcode). So for now, I'd like to keep it here just in case others run into the same thing.

ilyapuchka: my only guess is that it can be possible if one makes a build with fastlane, but otherwise it is not used when building with Xcode

jadams-babylon: That would make sense. I had been exploring those commands around that same time.

Cookbook/Technical-Documents/XcodeTips.md
```diff
@@ -0,0 +1,33 @@
...
+### Errors when pulling code/switching branches
+Various errors can occur when pulling code/switching branches. 
+Typically all that is needed is running `bundle exec pod install` and closing
+and reopening the project in Xcode. The following directions are
+intended to be a more thorough reset to deal with bigger changes.
+
+- Clean Build: Product > Clean (shortcut: cmd + shift + k)
+- Close Xcode
+- Update bundle: run `bundle install`
+- Update pods: `bundle exec pod deintegrate && bundle exec pod install --repo-update`
```
bryantgtx: From the comments in a recent Slack thread (https://babylonhealth.slack.com/archives/CFLN31FM5/p1556724542020400), deintegrate is not recommended, except as a last resort, as it can have some negative side-effects.

jadams-babylon: Yes. Trying to figure out what the preferred solution actually is before updating this PR.

jadams-babylon: Done

bryantgtx (approved):
Thanks - that's a lot easier to read, too!

PR #110, opened 2019-04-25
[Proposal] Integrate Danger

AliSoftware:
This proposal aims to integrate Danger to our iOS repository, in order to have it generate comments on our Pull Requests for common mistakes, nits, or to post informative messages about significant things to focus our attention on.

This initial proposal focuses on integrating Danger on our CI and implementing a first initial rule to make it useful. More proposals will come in the future to propose additional rules we might want to add to Danger, if this initial proposal gets accepted.

* [X] I will send a meeting invitation, using this [template](Template_Proposal_Meeting_Invitation.MD), scheduled for 2 weeks after this proposal is made, so an agreement can be reached.


Cookbook/Proposals/Danger-Integration.md
```diff
@@ -0,0 +1,152 @@
...
+
+This tool won't have any impact on the codebase at all.
+
+It will only have an impact on our review process: we will now be informed by Danger when a PR is bigger than normal, and will be able to decide to reject that PR for being too big (or accept that it's an acceptable exception if that's justified)
+
+## Alternatives considered
+
+### Integration in the CI
+
+We could potentially trigger the `danger` CLI earlier in the CI workflow, to be notified about anything we missed ASAP in the PR. But this brings the risk of not having access to the code (if we do that before the checkout step), limiting our possibilties at what rule we can write in the future (and to be honest I'm not even sure every feature of Danger would work if we haven't checked out the code yet)
```
ilyapuchka: I don't think there will be any issue with not having a source code when running danger, but then we will need to access the Dengerfile somehow, probably via GitHub API, which is doable, but may not worth it.

I'd suggest though to think about an option to run danger not in parallel to tests jobs, but right after it is installed (after `save_gems_cache` step in `checkout_code_steps`) so that we don't wait for things like cocoapods cache to be restored (~4-5 minutes) and persisting workspace (~2 minutes), so that the feedback will be provided after ~3 minutes (which is still long) instead of ~10

Cookbook/Proposals/Danger-Integration.md
```diff
@@ -0,0 +1,152 @@
...
+
+Those rules could also have been picked as initial candidate for the first rule we implement in Danger, but the rule about the PR size seems more straitforward to implement as a good way to test the initial usage of Danger on our process.
+
+All the rules mentionned in the list above will get their own separate proposal, so that we can discuss their benefits, the exact conditions to trigger each rule, and alternatives considered for each of them separately.
+
+_Note: I am hoping to write proposals for some of those additional rules soon, so that we could discuss them altogether in the upcoming proposal meeting alongside this very proposal for Danger, if time allows. I'll update the list above with links as I publish those other proposals_
+
+### Different conditions for PR size Rule
+
+The rule propose to make the condition based on the "overall lines of code added/removed in the diff" (that's the value returned by `git.lines_of_code` as documented in Danger docs). Which means it checks the value of `insertions` + `deletions`. That means that a PR which adds 20 lines and remove 30 lines will be seen as a size of 50.
```
ilyapuchka: This was something that was not very clear for me in our latest discussions and it might be that other team members thought about different way of applying this limit (so far exceeding the limit was raised on demand when it PR clearly too big), so definitely worth fixing it in a rule.

Another alternative could be to assert on additions and deletions separately, where each of them should not exceed the limit and so `+800/-800` is the maximum size of the PR. That was my understanding of the limit so far.

ilyapuchka (commented):
Great to see this happening

Cookbook/Proposals/Danger-Integration.md
```diff
@@ -0,0 +1,152 @@
...
+### Initial Rule to activate
+
+_As this doesn't make sense to integrate Danger in the repo without having any rule at all in the `Dangerfile` (as this would do nothing and thus would not solve any issue listed in the Motivation section), this proposal also suggest to implement one simple but useful rule during this initial Danger integration in our repo._
+
+The first rule that we propose to activate, and one which can be easy to write, is one to warn about a Pull Request which exceeds the maximum number of changes
+
+This simply consists of this code in the `Dangerfile`:
+
+```ruby
+warn("This PR is too big. Please split it in smaller PRs. If you can't split it, please justify why.") if git.lines_of_code > 800
```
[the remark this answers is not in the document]
AliSoftware: @ilyapuchka That's a great question.

The short answer is that Swift Danger is sure easier to use to write Dangerfiles, but way less advanced in terms of possibilities and plugin support, so I see a risk of limiting us when writing the rules. Orta even says on the "[Help me choose](https://danger.systems/js/js-vs-ruby.html)" page:

> You may notice that there was a Swift logo in the index page, Swift is the first language supported by Danger JS’s process environment. Once a new language has been built out, and is stable, then I’ll be building out some documentation in this site.

(For example, plugins for `swliftint` or similar are not available in Danger Swift yet afaik)

* [ ] I'll add that to "Alternatives Considered" though 👍 

As for the details of rule implementation, it's also mentioned already in the "Alternatives Considered" section.

* [ ] I'll add the idea about ignoring localisation or snapshots, great idea 👍

Cookbook/Proposals/Danger-Integration.md
```diff
@@ -0,0 +1,152 @@
...
+
+We could also decide to trigger `danger` later, for example _after_ the `test_babylon` job. This would give the advantage of allowing Danger to have access to test results for example. Which means that in the future we could use Danger to post a comment in the PR to tell which test failed or post the link to the HTML report maybe.  
+We think that those potential future rules (based on test results) are not our top priorities for now though and it's too early to decide to move Danger so late in the build pipeline, as it would mean only having feedback fron Danger _after_ the `test_pr` job has finished – which is usually one of the last to finish on PRs. It's totally imaginable to move the `danger` step after the test job in a later PR if it becomes a limitation for a rule we want to implement in the future.
```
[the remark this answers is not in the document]
AliSoftware: @ilyapuchka True. Note though that Danger don't create a new comment each time it's run, it only updates the existing one. So typically, after the author initially created the PR, the CI will kick in and Danger will post a comment, likely before any initial feedback from other team members. Then team members would comment on the PR later, and if there are new commits being pushed to the PR and thus CI being re-run, Danger will only update its comment, which doesn't generate a GitHub notification.

Cookbook/Proposals/Danger-Integration.md
AliSoftware: * [ ] Add alternative to count `insertions` and `deletions` separately

Cookbook/Proposals/Danger-Integration.md
AliSoftware: * [ ] Add alternative to run `danger` after `save_gems_cache`

@ilyapuchka Note that if we do that, that means that `danger` will be triggered by any job that uses the `checkout_code_steps`, not limiting it to `test_pr` and `test_release` then, right?

Cookbook/Proposals/Danger-Integration.md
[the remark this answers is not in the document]
ilyapuchka: This behaviour can be configured though IIRC and I can see how having comment updated can be confusing for author of PR. So I'd say if it is possible we should explore other options of feedback, i.e. update comment, but send the private Slack message to the author

Cookbook/Proposals/Danger-Integration.md
ilyapuchka: right, but nothing that we can't change in the config - the step can have a condition depending on the job name like we have it for steps like running `brew install`

Cookbook/Proposals/Danger-Integration.md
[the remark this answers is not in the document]
ilyapuchka: true, but then again I doubt there is anything tricky in ruby plugins that would block rewriting them in Swift, unless the Danger framework for Swift does not have a feature parity with Danger module for Ruby

Cookbook/Proposals/Danger-Integration.md
```diff
@@ -0,0 +1,194 @@
...
+
+At this point, it is worth reminding the reader that Danger only posts one comment (after the initial run of the CI) and then _updates_ said comment on subsequent runs; which means that even if you commented or watched the PR, you won't be spammed by GitHub notifications, as comment _updates_ don't trigger notifications on GitHub. You'd get a notification if you commented _before_ Danger had time to post its initial comment after the initial CI run though, which is worth thinking about when discussing how soon or late `danger` should be triggered too.
+
+### Picking a different first rule
+
+There are other ideas for initial rules to include in our `Dangerfile`, including but not limiting to:
+
+* Analyse the `pbxproj` files to find missing references and potential `(null)` values and warn about them
```
AliSoftware:
```suggestion
* Analyse the `pbxproj` files to find missing references and potential `(null)` values and warn about them (see #112)
```

Cookbook/Proposals/Danger-Integration.md
```diff
@@ -0,0 +1,194 @@
...
+* Analyse the `pbxproj` files to detect any files that was added to a target but is typically one to exclude from the final bundle – examples including `Info.plist` files but also `Snapshot` files
```
AliSoftware:
```suggestion
* Analyse the `pbxproj` files to detect any files that was added to a target but is typically one to exclude from the final bundle – examples including `Info.plist` files but also `Snapshot` files (see #112)
```

Cookbook/Proposals/Danger-Integration.md
```diff
@@ -0,0 +1,194 @@
...
+* Generate an informative comment if any of the build config changed in a project, so that reviewers would not miss those changes in a hard-to-read pbxproj diff
+* Generate an informative comment if the `Podfile.lock` changed, so that reviewers take extra care when pods are added, removed or updated on the project
+* Warn if the `CHANGELOG.md` hasn't been updated, and/or enforcing updating the SDK changelog if there was any change in `SDK/*`
+* Warn if the PR doesn't have the author assigned to it
+* Report SwiftLint warnings in the PR (though it would require to move the execution of Danger after the `test_pr` job, and besides, as of today those SwiftLint violations are reported as errors anyway)
+* ...
```
AliSoftware:
```suggestion
* Warn when the code introduce a new `// swiftlint:disable` usage in the codebase (especially outside of tests), to invite reviewers to ensure that this SwiftLint exception is justified
* ...
```

AliSoftware (commented):
Added a new proposal to list other rule ideas that we might want to discuss during this Proposal's meeting (if times allows)

Cookbook/Proposals/Danger-Integration.md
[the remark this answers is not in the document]
AliSoftware: It does not 100% sadly (well things might have been better since last time I looked, but some PR metadata are not retranscripted in the Swift DSL for example, requiring to parse the raw metadata ourselves and defeating the benefits of the DSL)

Cookbook/Proposals/Danger-Integration.md
```diff
@@ -0,0 +1,195 @@
...
+This proposal suggests that we adopt Danger to automatically check cases that we missed in the past during some PR reviews.
+
+This proposal also present an initial rule where Danger will be useful for us. Note that we will add more rules in the future via separate proposals or discussions (see [Alternatives Considered](#alternatives-considered)); this proposal focuses only on the first integration.
+
+
+## Motivation
+
+We have seen some common mistakes happen multiple times in our past PRs, like:
+
+ * `(null)` references in `pbxproj` files
```
daniloaliberti: have you seen those "Recovered References"? Could we also cover it?

AliSoftware: Good point. I'd need a fixture project containing some to spike if it's easy to detect and implement, but I think it should be.

I'm hoping it's not a special magic group shown by Xcode, but is considered as a normal group similar to any other in the `pbxproj` file format; if so, it would be easy enough to detect during the same loop as this grepping for null references

daniloaliberti (approved):
looks good! let's discuss 👍

Cookbook/Proposals/Danger-Integration.md
daniloaliberti: yes it is a normal group created by Xcode after a build. would make sense to be part of this loop 👍

AliSoftware:
```suggestion
 * `(null)` and `"Recovered References"` in `pbxproj` files
```

Cookbook/Proposals/Danger-Integration.md
```diff
@@ -0,0 +1,195 @@
...
+
+We could also decide to trigger `danger` later, for example _after_ the `test_babylon` job. This would give the advantage of allowing Danger to have access to test results for example. Which means that in the future we could use Danger to post a comment in the PR to tell which test failed or post the link to the HTML report maybe. But following that logic we should also trigger `danger` after `test_telus` and other jobs running tests, which will likely report some test failures multiple times (one per job if the failure is common to those targets). Which brings other questions and complexities into consideration (like does that mean one Danger comment on the PR... per test job, in each PR? Or a dedicated job after everything synchronizing all the previous test jobs but having to wait on them all before being able to report?) that might be too early to discuss for this initial proposal.
```
mr-v: > We could also decide to trigger `danger` later, for example _after_ the `test_babylon` job. This would give the advantage of allowing Danger to have access to test results for example. Which means that in the future we could use Danger to post a comment in the PR to tell which test failed or post the link to the HTML report maybe. 

I'm working on proposal for this. I think PR etiquette checks and build/test failure reporting could be done separately. This is related to issue that there's currently no way to wait for/hook into the end of the workflow (e.g. to post all those reports together) or to run extra job after failing one (e.g. fan-out to run test checks, fan-in to report any failures).

> Which brings other questions and complexities into consideration (...) that might be too early to discuss for this initial proposal.

There's a lot to discuss about on how to approach those reports, so we can take it to my proposal once I submit it.

Cookbook/Proposals/Danger-Integration.md
```diff
@@ -0,0 +1,195 @@
...
+
+At this point, it is worth reminding the reader that Danger only posts one comment (after the initial run of the CI) and then _updates_ said comment on subsequent runs; which means that even if you commented or watched the PR, you won't be spammed by GitHub notifications, as comment _updates_ don't trigger notifications on GitHub. You'd get a notification if you commented _before_ Danger had time to post its initial comment after the initial CI run though, which is worth thinking about when discussing how soon or late `danger` should be triggered too.
+
+### Picking a different first rule
+
+There are other ideas for initial rules to include in our `Dangerfile`, including but not limiting to:
```
mr-v: Rule about PR size seems about right to be the first one. 👍

Cookbook/Proposals/Danger-Integration.md
```diff
@@ -0,0 +1,195 @@
...
+
+* Use a ponderation of `A * insertions + B * deletions`, typically with `A > B > 0`, arguing that it's faster to review deleted code than added code. I'm not convinced that it's really the case – or worth going to such detail – but we might consider trying it in a future PR if we see that Danger warns too often about false positives
+
+* Ignore some files when computing the PR size, considering they don't play a significant role in estimating the PR _review time and complexity_. Here are some file types that we might consider ignoring when computing `insertions` and `deletions`: Snapshot files, Localization files, Assets (like images), ... (Note that this would complexify the implementation of the rule though, as we'd need to compute the number of additions and deletions ourselves by analyzing the diff file by file in the `Dangerfile` instead of direclty using the `insertions` and `deletions` metrics provided by the metadata)
```
mr-v: > Here are some file types that we might consider ignoring when computing `insertions` and `deletions`: **Snapshot files**, Localization files, Assets (like images), ... 

Definite yes for snapshots! I've seen PRs where most of the changes were related to those files, e.g.

```
# just snapshot changes
git log -1 --shortstat 03cf31d839466c0817a4d6640545aac5b85a4663 -- '*.png'
 102 files changed, 204 insertions(+), 204 deletions(-)

# all changes
git log -1 --shortstat 03cf31d839466c0817a4d6640545aac5b85a4663 -- '*.png'
 106 files changed, 326 insertions(+), 234 deletions(-)
```

It would even help with assessing real size/potential complexity of the PR if this calculated number was always posted, not just on warnings.

> (Note that this would complexify the implementation of the rule though, as we'd need to compute the number of additions and deletions ourselves by analyzing the diff file by file in the `Dangerfile` instead of direclty using the `insertions` and `deletions` metrics provided by the metadata)

Agree that we may want to introduce tweaks to the original rule later. btw I checked and it looks like there's already a nice method for this - [`info_for_file`](https://github.com/danger/danger/blob/master/lib/danger/danger_core/plugins/dangerfile_git_plugin.rb#L138-L148).

mr-v (approved):
Regarding using Swift version of Danger. FAQ mentions:

> [Circle CI doesn’t run my build consistently](https://danger.systems/swift/guides/faq.html#circle-ci-doesn--t-run-my-build-consistently)
>
> Yeah… We’re struggling with that one. It’s something we keep taking stabs at improving, so [keep an eye on the issues](https://github.com/danger/danger-js/search?q=circle&state=open&type=Issues&utf8=✓). Ideally this issue will get resolved and we’ll get it [fixed for free](https://discuss.circleci.com/t/pull-requests-not-triggering-build/1213).

---

I'm in favour of using **Ruby** version:

- more mature and battle tested,
- documentation,
- we already need to know some Ruby to maintain fastlane scripts.

For short rules it should be easy enough to pick up some Ruby scripting. We may want to add some guide on Ruby in the future (e.g. "iOS developer's guide to Ruby"), so everyone can be up to speed with at least basics of Ruby and its ecosystem.

TheAdamBorek (approved):

TheAdamBorek:
I agree with Witold's argumentation to use Ruby for danger file :)

inamiy (approved):
Count additions only will be great 👍

AliSoftware:
Actions agreed during proposal meeting:

- CI Integration
  - [ ] We'll integrate Danger as a step in the `checkout_code` job
  - [ ] We'll benchmark how much time it takes – using a big PR for the benchmarks as it's likely to take more time with bigger PRs
- [ ] PR Size rule
  - We'll integrate the PR size rule as a warning
  - The limit would only take `additions` into account, not deletions
  - The limit would not make any exception for Snapshots or Tests
  - The limit used by the rule should be `850` – to give some margin compared to our 800 soft limit recommended in our PR Etiquette
  - When the limit is reached and Danger warns about it, the warning message will give some stats to separate the counts of additions/deletions of code vs snapshots – or order for us to have some figures if we want to tweak the rule later.
 - [ ] I'll schedule a separate meeting for #112

PR #112, opened 2019-04-26
[Proposal] Danger rule for xcodeproj

AliSoftware:
This proposal is a follow-up of #110 and aims to propose an additional set of Danger Rules focused on making sure our `pbxproj` files are consistent, and that we don't risk pushing result of bad pbxproj merges, or dangerous changes in the project, on `develop`.

This proposal will be reviewed in the same meeting as #110 if time allows.

It will also serve during the discussions of #110 if there are strong opinions against the initial "PR size" rule proposed in #110, to propose to pick one of those rules as alternative initial rule to integrate Danger.

* [X] I will send a meeting invitation, using this [template](Template_Proposal_Meeting_Invitation.MD), scheduled for 2 weeks after this last meeting, so an agreement can be reached.

### DIFF
--- a/Cookbook/Proposals/Danger-Rule-Xcodeproj.md
+++ b/Cookbook/Proposals/Danger-Rule-Xcodeproj.md
@@ -15,19 +15,19 @@ It is dependent on [#110 - Integrate Danger](https://github.com/Babylonpartners/
 
 This means that some mistakes have ended up in `develop` in the past, like:
 
-* `(null)` references in `pbxproj` files
+* `(null)` references and `"Recovered References"` groups in `pbxproj` files
 * Files being added as part of a target membership while they shouldn't:
   * Snapshot files
   * xcconfig files
   * Info.plist files
 
-Some of those mistakes, especially `(null)` references, are generally the result of merge conflicts; so their appearances during a PR could be an indicator that something else could have gone wrong during the merge of the pbxproj file. It's generally a good incentive to double-check that there was no other side effect from the pbxproj merge.
+Some of those mistakes, especially `(null)` and "Recovered References" in `pbxproj`, are generally the result of merge conflicts; so their appearances during a PR could be an indicator that something else could have gone wrong during the merge of the pbxproj file. It's generally a good incentive to double-check that there was no other side effect from the pbxproj merge.
 
 Other mistakes like unexpected files in target membership could lead to files ending up in the final bundle uploaded to the AppStore, and in addition to being useless in the final `ipa`, bloat the app size for no reason.
 
 ## Proposed solution
 
-* We can use a simple "find" in every `*.pbxproj` for the string `/* (null) */`, and make Danger generate an inline comment on the found line(s) if any
+* We can use a simple "find" in every `*.pbxproj` for the strings `/* (null) */` and `name = "Recovered References";`, and make Danger generate an inline comment on the found line(s) if any
 
 This can be done pretty easily with Ruby, conceptually this will look like the logic below:
 
@@ -36,6 +36,7 @@ pbxprojs = git.modified_files.select { |f| f.end_with?('pbxproj') }
 pbxprojs.each do |filename|
   File.foreach(filename).with_index do |line, line_num|
     warn("null reference found in pbxproj", file: filename, line: line_num) if line.include?('/* (null) */') }
+    warn("recovered references found in pbxproj", file: filename, line: line_num) if line.include?('name = "Recovered References";') }
   end
 end
 ```
--- a/Cookbook/Proposals/Danger-Rule-Xcodeproj.md
+++ b/Cookbook/Proposals/Danger-Rule-Xcodeproj.md
@@ -7,7 +7,7 @@
 
 This proposal suggests to add some Danger rules checking for `xcodeproj` consistenty.
 
-It is dependent on [#110 - Integrate Danger](https://github.com/Babylonpartners/ios-playbook/pull/110) being accepted first.
+It is a follow-up on [#110 - Integrate Danger](https://github.com/Babylonpartners/ios-playbook/pull/110).
 
 ## Motivation
 
@@ -27,7 +27,17 @@ Other mistakes like unexpected files in target membership could lead to files en
 
 ## Proposed solution
 
-* We can use a simple "find" in every `*.pbxproj` for the strings `/* (null) */` and `name = "Recovered References";`, and make Danger generate an inline comment on the found line(s) if any
+We propose to use Danger rules to analyze the pbxproj and find all those issues during review:
+
+All those rules listed below, if implemented in the `Dangerfile`, will help us:
+
+* Keep the Project files sanitised and avoid commiting invalid pbxproj files, which are often the result of merge conflicts.
+* Be aware of potential higher risks of a merge gone wrong on a pbxproj file, to avoid commiting a corrupted pbxproj (e.g. one that could still be readable by `xcodebuild` but in which a reference to an image file got removed from the pbxproj due to a bad merge, making the app crash at runtime)
+* Avoid including large chunk of unrelevant resources (like test snapshots) in the final bundle
+
+### Null / Recovered References
+
+We can use a simple "find" in every `*.pbxproj` for the strings `/* (null) */` and `name = "Recovered References";`, and make Danger generate an inline comment on the found line(s) if any
 
 This can be done pretty easily with Ruby, conceptually this will look like the logic below:
 
@@ -41,7 +51,9 @@ pbxprojs.each do |filename|
 end
 ```
 
-* Using the `xcodeproj` gem (which is already part of our dependencies as it's used by `cocoapods`), we can also quickly analyze the structure of the project and detect resources that shouldn't have been added to build phases.
+### Unexpected Resources in Target
+
+Using the `xcodeproj` gem (which is already part of our dependencies as it's used by `cocoapods`), we can also quickly analyze the structure of the project and detect resources that shouldn't have been added to build phases.
 
 The draft ruby code below would for example loop on all `pbxproj` files modified by the PR, and for each of those Xcode projects, loop on all their targets to check if the "Resources Build Phase" of each target contain any of the files we want to forbid there:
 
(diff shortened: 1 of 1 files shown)


Cookbook/Proposals/Danger-Rule-Xcodeproj.md
```diff
@@ -0,0 +1,87 @@
...
+
+This means that some mistakes have ended up in `develop` in the past, like:
+
+* `(null)` references in `pbxproj` files
+* Files being added as part of a target membership while they shouldn't:
+  * Snapshot files
+  * xcconfig files
+  * Info.plist files
+
+Some of those mistakes, especially `(null)` references, are generally the result of build conflicts; so their apparition during a PR could be an indicator that something else could have gone wrong during the merge of the pbxproj file. It's generally a good incentive to double-check that there was no other side effect from the pbxproj merge.
```
diegopetrucci:
```suggestion
Some of those mistakes, especially `(null)` references, are generally the result of build conflicts; so their appearances during a PR could be an indicator that something else could have gone wrong during the merge of the pbxproj file. It's generally a good incentive to double-check that there was no other side effect from the pbxproj merge.
```

diegopetrucci (approved):

Cookbook/Proposals/Danger-Rule-Xcodeproj.md
AliSoftware: 👍 Actually, even:
```suggestion
Some of those mistakes, especially `(null)` references, are generally the result of merge conflicts; so their appearances during a PR could be an indicator that something else could have gone wrong during the merge of the pbxproj file. It's generally a good incentive to double-check that there was no other side effect from the pbxproj merge.
```

Cookbook/Proposals/Danger-Rule-Xcodeproj.md
```diff
@@ -0,0 +1,87 @@
...
+  * xcconfig files
+  * Info.plist files
+
+Some of those mistakes, especially `(null)` references, are generally the result of merge conflicts; so their appearances during a PR could be an indicator that something else could have gone wrong during the merge of the pbxproj file. It's generally a good incentive to double-check that there was no other side effect from the pbxproj merge.
+
+Other mistakes like unexpected files in target membership could lead to files ending up in the final bundle uploaded to the AppStore, and in addition to being useless in the final `ipa`, bloat the app size for no reason.
+
+## Proposed solution
+
+* We can use a simple "find" in every `*.pbxproj` for the string `/* (null) */`, and make Danger generate an inline comment on the found line(s) if any
```
ilyapuchka: Will Danger automatically provide timing values for all its tasks or its something that we would need to implement ourselves? I think it's something we can be interested in monitoring to avoid that those checks are taking a very long time considering the size of our project (actually same can be applied to other tools like swiftlint @diegopetrucci @TheAdamBorek )

ilyapuchka (approved):

Cookbook/Proposals/Danger-Rule-Xcodeproj.md
AliSoftware: Good question.
I don't think Danger provides timing info. What we could imagine doing is get the time at the beginning and end of Dangerfile and include a `message('Danger took 160ms to run its rules')` at the very end.

Not sure it's worth it though tbh, as like it's explained in the main #110 proposal, as opposed to SwiftLint — which needs to go thru every file in the project (changed or not) one by one and parse their content and even analyze the SIL representation for some rules — Danger only do some tests on the PR metadata… or grep on a very small subset of files (like one or two pbxproj which have changed in the PR), so it's way less time-consuming compared to e.g. SwiftLint.

Cookbook/Proposals/Danger-Rule-Xcodeproj.md
```diff
@@ -0,0 +1,87 @@
...
+* We can use a similar logic to warn about `Info.plist` files added to targets' resource phases by mistake.
+* We can also use a similar rule to prevent `__Snapshots__` directories be added in resources... and even to be added as reference to any `xcodeproj` at all.
+
+
+
+All those rules listed above, if implemented in the `Dangerfile`, will help us:
+
+* Keep the Project files sanitised and avoid commiting invalid pbxproj files, which are often the result of merge conflicts.
+* Be aware of potential higher risks of a merge gone wrong on a pbxproj file, to avoid commiting a corrupted pbxproj (e.g. one that could still be readable by `xcodebuild` but in which a reference to an image file got removed from the pbxproj due to a bad merge, making the app crash at runtime)
+* Avoid including large chunk of unrelevant resources (like test snapshots) in the final bundle
```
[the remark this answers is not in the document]
AliSoftware: - Testing that a file named `*Tests.swift` is included in the test target: should not be that complicated to write, I've written rules similar to that logic in the past and the xcodeproj gem is quite efficient for this kind of things.

- Checking it's also in the test action of the schemes: I have never tested writing such a rule for Danger; but from what I remember from when I contributed to the XCScheme parser years ago, we should have everything we need in the xcodeproj gem to write such a test as well, so I think it's totally doable.

---

What might be more "difficult" is for us to write down and define what the rules for such a check should be: on what criterion do we decide if a test file is in the *right* test target? what if a test file is in a test target but not the right one, or in one test action of one scheme but not the right one, what check could we imagine to detect that? Or do we consider that testing that the test file is "at least in one test target no matter which one" is a good enough test?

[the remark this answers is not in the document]
AliSoftware: Gave it a quick try and came up with this logic: https://gist.github.com/AliSoftware/447c5ed5e04d2f25f14d1922757048b7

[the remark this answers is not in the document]
ilyapuchka: I wouldn't complicate it. We just need to make sure that there are no tests which are not linked to any test target.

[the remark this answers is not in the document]
ilyapuchka: but that's totally not a priority, so please don't invest more effort into it!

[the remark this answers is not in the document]
AliSoftware: Yeah, I agree, I wasn't planning on adding that to the proposal yet anyway (don't worry I just toyed with it for like 5mn while waiting for a compilation to end in the background :P )

sergdort (approved):

daniloaliberti (approved):
Bringing [this](https://github.com/Babylonpartners/ios-playbook/pull/110#discussion_r281533976) from the Proposal #110 it would be nice to detect also those "Recovered References" folder.

👍

Cookbook/Proposals/Danger-Rule-Xcodeproj.md
```diff
@@ -0,0 +1,87 @@
...
+
+It is dependent on [#110 - Integrate Danger](https://github.com/Babylonpartners/ios-playbook/pull/110) being accepted first.
+
+## Motivation
+
+`pbxproj` project format is hard to parse, which means it's easy to miss some mistakes that could happen when an Xcode project has changed during a PR.
+
+This means that some mistakes have ended up in `develop` in the past, like:
+
+* `(null)` references in `pbxproj` files
```
AliSoftware:
```suggestion
* `(null)` references and `"Recovered References"` groups in `pbxproj` files
```

Cookbook/Proposals/Danger-Rule-Xcodeproj.md
```diff
@@ -0,0 +1,87 @@
...
+* Files being added as part of a target membership while they shouldn't:
+  * Snapshot files
+  * xcconfig files
+  * Info.plist files
+
+Some of those mistakes, especially `(null)` references, are generally the result of merge conflicts; so their appearances during a PR could be an indicator that something else could have gone wrong during the merge of the pbxproj file. It's generally a good incentive to double-check that there was no other side effect from the pbxproj merge.
```
AliSoftware:
```suggestion
Some of those mistakes, especially `(null)` and "Recovered References" in `pbxproj`, are generally the result of merge conflicts; so their appearances during a PR could be an indicator that something else could have gone wrong during the merge of the pbxproj file. It's generally a good incentive to double-check that there was no other side effect from the pbxproj merge.
```

Cookbook/Proposals/Danger-Rule-Xcodeproj.md
AliSoftware:
```suggestion
* We can use a simple "find" in every `*.pbxproj` for the strings `/* (null) */` and `name = "Recovered References";`, and make Danger generate an inline comment on the found line(s) if any
```

Cookbook/Proposals/Danger-Rule-Xcodeproj.md
```diff
@@ -0,0 +1,87 @@
...
+
+* We can use a simple "find" in every `*.pbxproj` for the string `/* (null) */`, and make Danger generate an inline comment on the found line(s) if any
+
+This can be done pretty easily with Ruby, conceptually this will look like the logic below:
+
+```ruby
+pbxprojs = git.modified_files.select { |f| f.end_with?('pbxproj') }
+pbxprojs.each do |filename|
+  File.foreach(filename).with_index do |line, line_num|
+    warn("null reference found in pbxproj", file: filename, line: line_num) if line.include?('/* (null) */') }
```
AliSoftware:
```suggestion
    warn("null reference found in pbxproj", file: filename, line: line_num) if line.include?('/* (null) */') }
    warn("recovered references found in pbxproj", file: filename, line: line_num) if line.include?('name = "Recovered References";') }
```

RuiAAPeres:

dmcrodrigues (approved):

dmcrodrigues:
Great stuff @AliSoftware ❤️

AliSoftware:
This PR has been merged too soon, before we had a chance to have the meeting.
It has thus been reverted then re-opened as #152 and #152 will be merged only when (and if) the meeting participants agree to the proposal.

For those who have already read and approved this PR, please re-approve in #152 (@ilyapuchka @diegopetrucci @dmcrodrigues @sergdort @daniloaliberti 😉)

PR #118, opened 2019-05-02
Style guide: add note about extension access qualifiers because of new SwiftLint rule.

diegopetrucci:
As promised, given [the new SwiftLint rule](https://github.com/Babylonpartners/babylon-ios/pull/7399).

### DIFF
--- a/Cookbook/Style-guide/README.md
+++ b/Cookbook/Style-guide/README.md
@@ -762,7 +762,7 @@ Only explicitly use `open`, `public`, and `internal` when you require a full acc
 
 Use access control as the leading property specifier. The only things that should come before access control are the `static` specifier or attributes such as `@IBAction`, `@IBOutlet` and `@discardableResult`.
 
-Use `private`, `fileprivate`, `internal`, and `public` visibility qualifiers on each function implemented in an extension, rather than have a single qualifier for the entire extension.
+Use `private`, `fileprivate`, `internal`, and `public` access-level modifiers on each function implemented in an extension, rather than having a single access-level modifier for the entire extension.
 
 **Preferred**:
 ```swift
--- a/Cookbook/Style-guide/README.md
+++ b/Cookbook/Style-guide/README.md
@@ -762,8 +762,6 @@ Only explicitly use `open`, `public`, and `internal` when you require a full acc
 
 Use access control as the leading property specifier. The only things that should come before access control are the `static` specifier or attributes such as `@IBAction`, `@IBOutlet` and `@discardableResult`.
 
-Use `private`, `fileprivate`, `internal`, and `public` access-level modifiers on each function implemented in an extension, rather than having a single access-level modifier for the entire extension.
-
 **Preferred**:
 ```swift
 private let message = "Great Scott!"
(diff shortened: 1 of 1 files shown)


Cookbook/Style-guide/README.md
```diff
@@ -764,6 +762,8 @@ Only explicitly use `open`, `public`, and `internal` when you require a full acc
 
 Use access control as the leading property specifier. The only things that should come before access control are the `static` specifier or attributes such as `@IBAction`, `@IBOutlet` and `@discardableResult`.
 
+Use `private`, `fileprivate`, `internal`, and `public` visibility qualifiers on each function implemented in an extension, rather than have a single qualifier for the entire extension.
+
```
diegopetrucci: It now applies to _all_ extensions, not just protocol's.

mluisbrown (approved):

Cookbook/Style-guide/README.md
AliSoftware: @diegopetrucci The "Preferred / Not Preferred" examples that are currently after this paragraph are related to the access control order rule, so the point before the one you moved there.

I'd rather keep the existing "Preferred / Not Preferred" example about "access control as a leading property" near the related paragraph, and add a dedicated "Preferred / Not Preferred" snippet for the rule you moved here. Maybe moving that rule before the "access control as leading property" instead of after would also make more sense.

Cookbook/Style-guide/README.md
```diff
@@ -764,6 +762,8 @@ Only explicitly use `open`, `public`, and `internal` when you require a full acc
 
 Use access control as the leading property specifier. The only things that should come before access control are the `static` specifier or attributes such as `@IBAction`, `@IBOutlet` and `@discardableResult`.
 
+Use `private`, `fileprivate`, `internal`, and `public` visibility qualifiers on each function implemented in an extension, rather than have a single qualifier for the entire extension.
```
AliSoftware: Maybe we should be consistent in our vocabulary, and either use "access control" or "access control qualifiers" or "visibility qualifiers", but not a mix of both between rules (e.g. this one and the previous one). Not sure what's best

RuiAAPeres:

Cookbook/Style-guide/README.md
inamiy:
```suggestion
Use `private`, `fileprivate`, `internal`, and `public` access-level modifiers on each function implemented in an extension, rather than having a single access-level modifier for the entire extension.
```

Swift doc uses the phrase "access-level modifier", so we can follow it:
https://docs.swift.org/swift-book/LanguageGuide/AccessControl.html

Cookbook/Style-guide/README.md
Ben-Henshall: updated

Ben-Henshall (approved):

Cookbook/Style-guide/README.md
```diff
@@ -782,6 +780,22 @@ class TimeMachine {
 }
 ```
 
+Use `private`, `fileprivate`, `internal`, and `public` access-level modifiers on each function implemented in an extension, rather than having a single access-level modifier for the entire extension.
```
AliSoftware: I'm nitpicking here, but the paragraph says "on each function" and the exemple shows a `var` not a `func` 😉and better be clear anyway 

```suggestion
Use `private`, `fileprivate`, `internal`, and `public` access-level modifiers on each property, type or function declared in an `extension`, rather than having a single access-level modifier for the entire `extension`.
```

AliSoftware (approved):

PR #132, opened 2019-05-15
Extend our approach to comments

TheAdamBorek:
I'd like to suggest extending our point on comments in the StyleGuide. Sometimes in PRs I had a feeling we have a different approach to comments. Currently, the point on comments in the StyleGuide is not clear enough. 

Related PR: https://github.com/Babylonpartners/babylon-ios/pull/7606

### DIFF
--- a/Cookbook/Proposals/Comments.md
+++ b/Cookbook/Proposals/Comments.md
@@ -221,4 +221,3 @@ It doesn't have any impact on existing codebase as this is about extending a rul
 ## Alternatives considered
 1. Leave the StyleGuide as it is, aka. reject this proposal.
 2. Modify the list of good comments. Maybe we don't agree on everything in this proposal but we like part of it.
-3. We can also extend this proposal to define our approach to `MARK`. Currentely, `MARK` is not covered in the Style Guide neither in this proposal.
--- a/Cookbook/Proposals/Comments.md
+++ b/Cookbook/Proposals/Comments.md
@@ -5,7 +5,7 @@
 
 ## Introduction
 
-I'd like to extend what our style guide says on comments. Right now it says very little and I think that can be improved extended. Currently it says:
+I'd like to extend what our style guide says on comments. Right now it says very little and I think that can be improved extended. Currently, it says:
 
 
 > Comments
@@ -19,11 +19,11 @@ I'd like to extend what our style guide says on comments. Right now it says very
 
 ## Motivation
 
-I think we sometimes overuse comments. The reason for this our StyleGuide doesn't say much on comments and this leads us to having different opinions whether a comment is worth to be added or not in our Pull Requests.
+I think we sometimes overuse comments. The reason for this our StyleGuide doesn't say much on comments and this leads us to have different opinions whether a comment is worth to be added or not in our Pull Requests.
 
-I think comments are rather bad than good because of 2 following reasons: 
+I think comments are rather bad than good because of the 2 following reasons: 
 
-1. Comments eventually starts to lie.
+1. Comments eventually start to lie.
 2. Usage of comments gives a false-positive feeling to the author that the code is readable.
 
 Look at this comment:
@@ -40,7 +40,7 @@ public enum ConsultationMethod: String {
     case faceToFace = "face to face"
 }
 ```
-At the beginning the `ConsultationMethod` had indeed 2 cases. This may be trivial here without big consequences however that not always a case and a lie may cost us many of hours.
+At the beginning, the `ConsultationMethod` had indeed 2 cases. This may be trivial here without big consequences however that not always a case and a lie may cost us many of hours.
 
 "Cross-files" comments are even more dangerous to become a lie at some point. Consider the following comment:
 ```
@@ -52,7 +52,7 @@ public func makeBookingConfirmation(
     dismiss: @escaping () -> Void
 ) -> UIViewController 
 ```
-Such a comment is a leak of information from another file into Builder. Builder doesn't know how BusinessController works. A change in a BusinessController can change the behaviour of given  screen, while there won't be a need to even enter to the builder to find out there is a comment. At the end, comment stays there, unchanged.
+Such a comment is a leak of information from another file into Builder. Builder doesn't know how BusinessController works. A change in a BusinessController can change the behavior of given screen, while there won't be a need to even enter the builder to find out there is a comment. In the end, comment stays there, unchanged.
 
 In our Pull Requests I've seen sometimes unnecessary comments
 ```swift
@@ -93,13 +93,13 @@ extension AppointmentDTO {
 ## Proposed solution
 
 ### Our approach 
-The most important thing is to improve definition of our approach to comments as currently it isn't said clearly. Moreover, it would be nice to have a common way how we comment our code when we are forced to add them. 
+The most important thing is to improve the definition of our approach to comments as currently it isn't said clearly. Moreover, it would be nice to have a common way how we comment our code when we are forced to add them. 
 
 In my opinion, our approach should insist on writing always a **self-documented code**. It should be our goal. However, writing self-documented code **doesn't** mean we shouldn't use comments. It means, that commenting our code is a **necessary evil**:
 
 > The proper use of comments is to compensate for our failure to express ourself in code. Note that I used the word failure. I meant it. Comments are always failures. We must have them because we cannot always figure out how to express ourselves without them, but their use is not a cause for celebration.
-	// Robert C. Martin - "Clean Code"
-	
+    // Robert C. Martin - "Clean Code"
+    
 I think that adding the above section to our style guide would make it clear what's our approach to comments.
 
 ### What is a good comment?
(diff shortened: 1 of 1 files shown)


Cookbook/Proposals/Comments.md
```diff
@@ -0,0 +1,223 @@
...
+> Avoid the use of C-style comments (/* ... */). Prefer the use of double- or triple-slash.
+
+
+## Motivation
+
+I think we sometimes overuse comments. The reason for this our StyleGuide doesn't say much on comments and this leads us to have different opinions whether a comment is worth to be added or not in our Pull Requests.
+
+I think comments are rather bad than good because of the 2 following reasons: 
+
+1. Comments eventually start to lie.
```
NSMyself:
```suggestion
1. Comments eventually start to lie since the code frequently gets updated but comments remain unchanged
```

Cookbook/Proposals/Comments.md
```diff
@@ -0,0 +1,224 @@
...
+- **TODO comments**
+    This already has a good definition of our approach in the StyleGuide and I think we are fine with that approach.
+- **Doc headers for public API**
+    Good documentation helps consumers of the SDK to understand what the code does. However, I don't know if it's worth to add a comment to every public function/field, even if a comment is just a repetition. The following comment doesn't add any value (unless it's needed by a doc generation tool if we use any):
+    ```swift
+     /// Supported biometric types
+     public var supportedBiometry: BiometryType
+    ```
+- **Explanation of Intent/Clarification**
+    It's a kind of comment when a developer tries to explain what he's trying to achieve when a code is not verbose enough. Example: Math code, sometimes parsing.
```
AliSoftware: I feel like those kind of comments are the main ones we overuse, explaining the code while trying to make the code self-explanatory should be the better approach. So not sure I'd list those in the "Good comments" category in that regard...

```suggestion
    It's a kind of comment when a developer tries to explain what they're trying to achieve when a code is not verbose enough. Example: Math code, sometimes parsing.
```

Cookbook/Proposals/Comments.md
```diff
@@ -0,0 +1,223 @@
...
+```swift
+public enum ConsultationMethod: String {
+    case phone
+    case video
+    case faceToFace = "face to face"
+}
+```
+At the beginning, the `ConsultationMethod` had indeed 2 cases. This may be trivial here without big consequences however that not always a case and a lie may cost us many of hours.
+
+"Cross-files" comments are even more dangerous to become a lie at some point. Consider the following comment:
```
NSMyself:
```suggestion
"Cross-files" comments are even more likely of becoming lies at any given point. Consider the following comment:
```

Cookbook/Proposals/Comments.md
```diff
@@ -0,0 +1,224 @@
...
+    If we had a test which takes an hour to run, it would be nice to have a comment warning about it.
+- **Amplification**
+    Sometimes some piece of code seems to be less important than it really is. In such cases, it's worth to mark it.
+- **Reference to external sources**
+    For example, ISO 639 for country codes.
+
+The above list is what defines a good comment. **However**, we still have to remember that **every** comment **is our failure** to name our code better. If someone gives you a good hint how to remove a comment without sacrificing readability **we should go for it**.
+
+### How can we replace a comment?
+In case of a **what** comments usually it is very easy to replace a comment. Usually, it's all about taking what comment says and create a variable or function with such a name.
```
AliSoftware:
```suggestion
In case of a **what** comment it is usually very easy to replace it by taking what the comment says and create a variable or function with such a name.
```

Cookbook/Proposals/Comments.md
```diff
@@ -0,0 +1,223 @@
...
+        /// paid by user
+        case paid
+        /// doctor finished adding notes on the portal
+        case completed
+        /// patient did not answer the doctor's call
+        case noShow = "no_show"
+    }
+}
+```
+Can be simplified without any lose in understanding what the code represents to:
```
NSMyself:
```suggestion
Can be simplified without any loss in understanding what the code represents to:
```

Cookbook/Proposals/Comments.md
```diff
@@ -0,0 +1,223 @@
...
+
+```swift
+extension AppointmentDTO {
+    public enum State: String, Codable {
+        /// waiting for payment
```
NSMyself: Shouldn't this comment be removed as well?

Cookbook/Proposals/Comments.md
```diff
@@ -0,0 +1,223 @@
...
+        case doctorFinishedAddingNotesOnThePortal = "completed"
+        case patientDidntAnswerTheCall = "no_show"
+    }
+}
+```
+
+## Proposed solution
+
+### Our approach 
+The most important thing is to improve the definition of our approach to comments as currently it isn't said clearly. Moreover, it would be nice to have a common way how we comment our code when we are forced to add them.
```
NSMyself:
```suggestion
The most important thing is to improve the definition of our approach to comments as currently it isn't clearly defined. Moreover, it would be nice to have a common way how we comment our code when we are forced to add them. 
```

Cookbook/Proposals/Comments.md
```diff
@@ -0,0 +1,224 @@
...
+    For example, ISO 639 for country codes.
+
+The above list is what defines a good comment. **However**, we still have to remember that **every** comment **is our failure** to name our code better. If someone gives you a good hint how to remove a comment without sacrificing readability **we should go for it**.
+
+### How can we replace a comment?
+In case of a **what** comments usually it is very easy to replace a comment. Usually, it's all about taking what comment says and create a variable or function with such a name.
+
+
+### How do we comment?
+Even with the proposed approach, we will be forced to comment our code at some point. I would like to propose a way of doing it. The biggest difficulty for me when I read a comment (especially a long one, explaining "why") is to understand to what scope the comment refers to. I think that we should always try to export commented code into a separate method/class and put the comment as a header to this method.
```
AliSoftware: Even for comments applying to only one line, and especially todo-comments and reference-to-links comments?

For comments regarding a workaround for example I agree. for a comment referencing ISO codes or a jira ticket, not so sure.

Cookbook/Proposals/Comments.md
```diff
@@ -0,0 +1,223 @@
...
+    }
+}
+```
+
+## Proposed solution
+
+### Our approach 
+The most important thing is to improve the definition of our approach to comments as currently it isn't said clearly. Moreover, it would be nice to have a common way how we comment our code when we are forced to add them. 
+
+In my opinion, our approach should insist on writing always a **self-documented code**. It should be our goal. However, writing self-documented code **doesn't** mean we shouldn't use comments. It means, that commenting our code is a **necessary evil**:
```
NSMyself:
```suggestion
In my opinion, our approach should insist on writing always a **self-documented code**. It should be our goal. However, writing self-documented code **doesn't** mean we shouldn't use comments. Commenting our code is, at times, a **necessary evil**:
```

Cookbook/Proposals/Comments.md
```diff
@@ -0,0 +1,223 @@
...
+
+> The proper use of comments is to compensate for our failure to express ourself in code. Note that I used the word failure. I meant it. Comments are always failures. We must have them because we cannot always figure out how to express ourselves without them, but their use is not a cause for celebration.
```
NSMyself: Nice 👍

Cookbook/Proposals/Comments.md
```diff
@@ -0,0 +1,224 @@
...
+
+### How do we comment?
+Even with the proposed approach, we will be forced to comment our code at some point. I would like to propose a way of doing it. The biggest difficulty for me when I read a comment (especially a long one, explaining "why") is to understand to what scope the comment refers to. I think that we should always try to export commented code into a separate method/class and put the comment as a header to this method.
+
+The above solution has 2 benefits:
+
+1. It's visible to what part of a code comment refers.
+2. It improves readability for a developer who doesn't need to know why a particular piece of code is written. Sometimes, during debugging for example, what you need to know is **what** happens, not **why**.
+
+To illustrate point `2.`. If there is an UIKit bug, with dismissing a modal screen which is ugly but works just fine, all we care about is the fact that the code has been dismissed. During a debuting we now we can go just further without a need to understand how `dismissing` is handled.
```
AliSoftware: 👍 

```suggestion
To illustrate point `2.`. If there is an UIKit bug, with dismissing a modal screen which is ugly but works just fine, all we care about is the fact that the code has been dismissed. During a debugging we now we can go just further without a need to understand how `dismissing` is handled.
```

Cookbook/Proposals/Comments.md
```diff
@@ -0,0 +1,223 @@
...
+I think that adding the above section to our style guide would make it clear what's our approach to comments.
+
+### What is a good comment?
+
+To make our StyleGuide more verbose we can also add a list of what kind of comments we think are "good".
+
+**Good comments**:
+
+- **TODO comments**
+    This already has a good definition of our approach in the StyleGuide and I think we are fine with that approach.
```
NSMyself:
```suggestion
    This already has a good definition of our approach in the StyleGuide and I think we are fine with that approach. Don't forget to add JIRA's ticket number, if applicable.
```

Cookbook/Proposals/Comments.md
```diff
@@ -0,0 +1,223 @@
...
+- **Doc headers for public API**
+    Good documentation helps consumers of the SDK to understand what the code does. However, I don't know if it's worth to add a comment to every public function/field, even if a comment is just a repetition. The following comment doesn't add any value (unless it's needed by a doc generation tool if we use any):
```
NSMyself:
```suggestion
    Good documentation helps consumers of the SDK understand what the code does. However, I don't know if it's worth to add a comment to every public function/field, even if a comment is just a repetition. The following comment doesn't add any value (unless it's needed by a doc generation tool if we use any):
```

Cookbook/Proposals/Comments.md
```diff
@@ -0,0 +1,224 @@
...
+
+
+## Impact on existing codebase
+It doesn't have any impact on existing codebase as this is about extending a rule in our StyleGuide. What will change is during code reviews we may have more suggestions on how to replace a comment with a code.
+
+
+## Alternatives considered
+1. Leave the StyleGuide as it is, aka. reject this proposal.
+2. Modify the list of good comments. Maybe we don't agree on everything in this proposal but we like part of it.
+3. We can also extend this proposal to define our approach to `MARK`. Currentely, `MARK` is not covered in the Style Guide neither in this proposal.
```
AliSoftware: I think it's worth having a discussion about inclusion of `// MARK`, but outside of this proposal, since this proposal is focused on explanatory comments, while considering using `// MARK` or not is more of a code-organisation matter and belongs in a separate discussion.

Cookbook/Proposals/Comments.md
```diff
@@ -0,0 +1,223 @@
...
+### How can we replace a comment?
+In case of a **what** comments usually it is very easy to replace a comment. Usually, it's all about taking what comment says and create a variable or function with such a name.
+
+
+### How do we comment?
+Even with the proposed approach, we will be forced to comment our code at some point. I would like to propose a way of doing it. The biggest difficulty for me when I read a comment (especially a long one, explaining "why") is to understand to what scope the comment refers to. I think that we should always try to export commented code into a separate method/class and put the comment as a header to this method.
+
+The above solution has 2 benefits:
+
+1. It's visible to what part of a code comment refers.
```
NSMyself:
```suggestion
1. Clearly defines what part of a code comment refers to.
```

Cookbook/Proposals/Comments.md
```diff
@@ -0,0 +1,223 @@
...
+2. It improves readability for a developer who doesn't need to know why a particular piece of code is written. Sometimes, during debugging for example, what you need to know is **what** happens, not **why**.
+
+To illustrate point `2.`. If there is an UIKit bug, with dismissing a modal screen which is ugly but works just fine, all we care about is the fact that the code has been dismissed. During a debuting we now we can go just further without a need to understand how `dismissing` is handled.
```
NSMyself: Didn't understand the `debuting` bit; can you please clarify?

AliSoftware: @NSMyself it's likely a typo: https://github.com/Babylonpartners/ios-playbook/pull/132#discussion_r284149699

NSMyself: That would make sense indeed, thanks Olivier

Cookbook/Proposals/Comments.md
TheAdamBorek: I thought I have :D

TheAdamBorek:
```suggestion
```

Cookbook/Proposals/Comments.md
TheAdamBorek: @AliSoftware I agree! However, I think such a comment is better than not adding it and not providing good naming. The comment is fine however,  we still can disagree with putting it there. That's why a few lines below I've written :

> The above list is what defines a good comment. **However**, we still have to remember that **every** comment **is our failure** to name our code better. If someone gives you a good hint how to remove a comment without sacrificing readability **we should go for it**.

Cookbook/Proposals/Comments.md
TheAdamBorek: > For comments regarding a workaround for example I agree.
👍 

>  for a comment referencing ISO codes or a jira ticket, not so sure.
👍 

I agree I wasn't thinking about iSO codes etc. I'll mention that somewhere

Cookbook/Proposals/Comments.md
TheAdamBorek: yes, typo ;)

Cookbook/Proposals/Comments.md
```diff
@@ -0,0 +1,222 @@
...
+    }
+}
+```
+
+## Proposed solution
+
+### Our approach 
+The most important thing is to improve the definition of our approach to comments as currently it isn't clearly defined. Moreover, it would be nice to have a common way of commenting our code when we are forced to do so. 
+
+In my opinion, our approach should insist on writing always a **self-documented code**. It should be our goal. However, writing self-documented code **doesn't** mean we shouldn't use comments. Commenting our code is, at times, a **necessary evil**:
```
TheAdamBorek:
```suggestion
In my opinion, our approach should insist on writing always a **self-documented code**. It should be our goal. However, writing self-documented code **doesn't** mean we shouldn't use comments. Commenting our code is a **necessary evil**:
```

Cookbook/Proposals/Comments.md
TheAdamBorek: I've already removed this line as I agree.
And I was originally wrong. The StyleGuide already defines our approach to `// MARK` ;)

Cookbook/Proposals/Comments.md
```diff
@@ -0,0 +1,222 @@
...
+}
+```
+Can be simplified without any loss in understanding what the code represents to:
+
+```swift
+extension AppointmentDTO {
+    public enum State: String, Codable {
+        case waitingForPayment = "pending"
+        /// timeout == 10 min
+        case paymentTimedOut = "timed_out"
```
RuiAAPeres: Actually I find the original comment better than this open, since it provides context as to the condition that leads to the timeout (not just the time).

TheAdamBorek: Maybe it's a good exception for the comment. It's hard to say for me now as I didn't know what's important when I took this snippet on the same level the author knows while writing the code. 

However if that's a valid information from app perspective we can change the case naming to `case userFailedToPayInRequiredTime // timeOut == 10 min` to express that.

inamiy:
I added my own opinion on doc-comments in https://github.com/Babylonpartners/babylon-ios/pull/7606#issuecomment-494341340 .
I will check this PR later :)

mluisbrown:
IMO the current Style Guide is sufficient, and says *exactly* how comments should be used (in my view): Comment the *WHY* not the *HOW*.

This slide from [this WWDC '19 presentation](https://developer.apple.com/videos/play/wwdc2019/239/) sums up the value of comments:
image

I don't agree with the current trend to view comments as a "necessary evil".

Cookbook/Proposals/Comments.md
mluisbrown: I disagree with this paragraph 100% 👎  Comments are *not* a failure. Comments are *not* a necessary evil. Good comments are a *critical* component of writing good code. That said, not all code needs to be commented, and the existing style guide explains when they are needed: explain the WHY, not the HOW.

mluisbrown (changes requested):
I reject this proposal, as IMO the current style guide on comments is sufficient and correct.

TheAdamBorek:
The problem which I see with the current Styleguide is I don't know if I may request changes when I don't like how comments were used. **Am I the only one who has doubts here?** 

Take this PR as an example https://github.com/Babylonpartners/babylon-ios/pull/7606 

I want to avoid a situation when we think that adding a comment is fine and we don't even try to write more readable code by extracting a variable or method. That's what I mean by treating comments as  "necessary evil". Keep in mind, I don't want to stop using comments. I'm saying we should be more careful when we use them.

inamiy (changes requested):
I wish we have TLDR section for this proposal, like Apple's slide https://github.com/Babylonpartners/ios-playbook/pull/132#issuecomment-501210891 which is very simple.

One thing I really want to stress out here is **the risk of renaming to cause the breaking changes** in our understanding. 

If we have 30 people working on the same class and suddenly its name changed with tons of diffs, that will be very annoying.
Even worse, if we didn't add a perfect name for it, we will have another renaming war coming soon, so it will be an endless war.
Comments, on the other hand, ease such troubles by having less association with the syntax.

So, I'm personally not in favor of renaming things all the time.

Cookbook/Proposals/Comments.md
TheAdamBorek: @mluisbrown do you agree with a statement that code is better than a comment and if you can express yourself with code clearly enough then comment is not needed?

mluisbrown: @TheAdamBorek yes, of course, if you can make your code clear enough then a comment is not needed. It should be rare that we need comments to explain *WHAT* the code does, but IMO we don't have enough comments to explain *WHY* the code is doing what it does.

However I think the current comments section of the style guide already says basically this ☝️ so I don't see the need for any expansion on that.

mluisbrown:
> May I not give an approve to a PR when I'm not comfortable how comments are used?

Absolutely. You can always request changes if you think the comments are unnecessary or if the code could be improved to remove the need for them 👍

TheAdamBorek:
Closing this proposal as we have agreed that a style guide already said what it needs to say on comments.

PR #145, opened 2019-05-31
Add TabBarAndNavigationFramework.md (WIP)

inamiy:
TBD

### DIFF
--- a/Cookbook/Proposals/TabBarAndNavigationFramework.md
+++ b/Cookbook/Proposals/TabBarAndNavigationFramework.md
@@ -1,0 +1,65 @@
+# Refactoring TabBar and Navigation Framework
+
+* Author(s): Yasuhiro Inami
+* Review Manager: Rui Peres
+
+## Introduction
+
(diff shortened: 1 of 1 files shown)


zzcgumn:
> in contrast to Cocoa's Responder Chain system.

My experience from working with one single, centralised router is that this entity can become very complicated. I came across this approach while working with a cross platform framework called MvvmCross. 

Therefore I think we should try to architecture the routing service with a domain model in mind and think more like the responder chain and less like notification the centre.

inamiy:
@zzcgumn 
Router can be split into multiple nested structure, but the event needs to be handled and routed from top to bottom, just like Universal Link's URL path structure. 
So, these nested structures will be hidden from outside, and for any presenting screens, the core Router is only visible as an entrypoint facade.

RuiAAPeres:
Closing this for now. Please re-open when you can continue. ❤️

inamiy:
This proposal has been superceded by https://github.com/Babylonpartners/ios-playbook/pull/232 Root & Tab Framework.

PR #152, opened 2019-06-06
[Proposal] Danger rule for xcodeproj

AliSoftware:
This is the revival of #112 which was accidentally prematurely merged before the proposal meeting.

---

This proposal is a follow-up of #110 and aims to propose an additional set of Danger Rules focused on making sure our `pbxproj` files are consistent, and that we don't risk pushing result of bad pbxproj merges, or dangerous changes in the project, on `develop`.

* [X] I will send a meeting invitation, using this [template](Template_Proposal_Meeting_Invitation.MD), scheduled for 2 weeks after this last meeting, so an agreement can be reached.

### DIFF
--- a/Cookbook/Proposals/Danger-Rule-Xcodeproj.md
+++ b/Cookbook/Proposals/Danger-Rule-Xcodeproj.md
@@ -15,19 +15,19 @@ It is dependent on [#110 - Integrate Danger](https://github.com/Babylonpartners/
 
 This means that some mistakes have ended up in `develop` in the past, like:
 
-* `(null)` references in `pbxproj` files
+* `(null)` references and `"Recovered References"` groups in `pbxproj` files
 * Files being added as part of a target membership while they shouldn't:
   * Snapshot files
   * xcconfig files
   * Info.plist files
 
-Some of those mistakes, especially `(null)` references, are generally the result of merge conflicts; so their appearances during a PR could be an indicator that something else could have gone wrong during the merge of the pbxproj file. It's generally a good incentive to double-check that there was no other side effect from the pbxproj merge.
+Some of those mistakes, especially `(null)` and "Recovered References" in `pbxproj`, are generally the result of merge conflicts; so their appearances during a PR could be an indicator that something else could have gone wrong during the merge of the pbxproj file. It's generally a good incentive to double-check that there was no other side effect from the pbxproj merge.
 
 Other mistakes like unexpected files in target membership could lead to files ending up in the final bundle uploaded to the AppStore, and in addition to being useless in the final `ipa`, bloat the app size for no reason.
 
 ## Proposed solution
 
-* We can use a simple "find" in every `*.pbxproj` for the string `/* (null) */`, and make Danger generate an inline comment on the found line(s) if any
+* We can use a simple "find" in every `*.pbxproj` for the strings `/* (null) */` and `name = "Recovered References";`, and make Danger generate an inline comment on the found line(s) if any
 
 This can be done pretty easily with Ruby, conceptually this will look like the logic below:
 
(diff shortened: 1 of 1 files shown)


dmcrodrigues (approved):

daniloaliberti (approved):

diegopetrucci (approved):

jtomanik (approved):

Cookbook/Proposals/Danger-Rule-Xcodeproj.md
```diff
@@ -0,0 +1,132 @@
...
+      fail("#{path} should be removed from #{project_file}:#{target.name}" if path.end_with? '.xcconfig'
+    end
+  end
+end
+```
+
+We can use this logic to warn about those type of files/folders added to targets' resource phases by mistake:
+ - `.xcconfig` files
+ - `Info.plist` 
+ - `__Snapshots__` directories
```
AliSoftware: + README.md

Cookbook/Proposals/Danger-Rule-Xcodeproj.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Detect tests not added to scheme
+
+We could once again use the `xcodeproj` gem to detect if any new tests file is properly added to a test target, and that the corresponding test is enabled in the test action in the scheme, to ensure that no tests are skipped.
```
AliSoftware: Team seem to agree we don't need this rule and it might trigger some false positives if we did (since things are not uniform rn) so don't implement this one.

Cookbook/Proposals/Danger-Rule-Xcodeproj.md
```diff
@@ -0,0 +1,132 @@
...
+`<PBXSourcesBuildPhase UUID=`58E92B48208799C90037BD6A`>` attempted to initialize an object with an unknown UUID. `5FFC934F2296E0980088BFAA` for attribute: `files`. This can be the result of a merge and  the unknown UUID is being discarded.
+```
+
+Some of those mistakes, especially `(null)`, "Recovered References" and invalid UUIDs in `pbxproj`, are generally the result of merge conflicts; so their appearances during a PR could be an indicator that something else could have gone wrong during the merge of the pbxproj file. It's generally a good incentive to double-check that there was no other side effect from the pbxproj merge.
+
+Other mistakes like unexpected files in target membership could lead to files ending up in the final bundle uploaded to the AppStore, and in addition to being useless in the final `ipa`, bloat the app size for no reason.
+
+## Proposed solution
+
+We propose to use Danger rules to analyze the pbxproj and find all those issues during review:
```
AliSoftware: Make sure that the Danger comment contains suggestions about how to solve the issues when they happen.

AliSoftware (commented):
Summary of the decision at end of meeting:

 - Overall agreement about necessity of rules
 - When Danger warns about those, we need to make sure the Danger comment gives good advice on how to fix them
   * For example, if you have `(null)` in your pbxproj, explain the steps to fix them – which often ends up being removing the offending lines... but not always, so you should be careful and inspect what was actually removed by the bad xcodeproj merge and if it was intentional)
 - The rule about tests in right folders and targets was rejected
   * because we don't have consistency in the way we organize our tests in folders and folder names yet (we have suggestions for new tests but changes in legacy tests don't respect the new rules, etc) so that would make the rule trigger false positives
   * and besides it's not really an issue people have encountered or made their life harder lately so it's not worth it.

Cookbook/Proposals/Danger-Rule-Xcodeproj.md
AliSoftware:
```suggestion
 - `__Snapshots__` directories
 - `.md` files (especially we have one `README.md`, maybe more in the future)
```

PR #153, opened 2019-06-07
[Proposal] Additional rules for Danger and SwiftLint

NSMyself:

### DIFF
--- a/Cookbook/Proposals/AdditionalRulesForDanger.md
+++ b/Cookbook/Proposals/AdditionalRulesForDanger.md
@@ -89,7 +89,7 @@ In this particular case I believe we could use both approaches actually.
 
 ### Detect Current usage outside Builders et al
 
-`Current` is a very powerful concept whose usage is only allowed in a very restrict number of places. Check [Current Guide](./Technical-Documents/Current-guide.md). As such, we should enable automatic monitoring in order to prevent it from being abused throughout our codebase. Nevertheless, Current is still being used in a lot of places (598 results across 144 files) and we should go through every case individually before commiting to a decision:
+`Current` is a very powerful concept whose usage is only allowed in a very restrict number of places. More detailed information about Current can be found [here in our Current Guide](../Technical-Documents/Current-guide.md). As such, we should enable automatic monitoring in order to prevent it from being abused throughout our codebase. Nevertheless, Current is still being used in a lot of places (598 results across 144 files) and we should go through every case individually before commiting to a decision:
 
 `Current` should only be allowed in:
 - `AppDelegate.swift`
--- a/Cookbook/Proposals/AdditionalRulesForDanger.md
+++ b/Cookbook/Proposals/AdditionalRulesForDanger.md
@@ -205,4 +205,4 @@ Furthermore, we would be increasing Danger's processing time - although for the 
 
 * Don't implement these rules at all
 
-Regardless of what we chose, the should still strive for reducing our tremendous amount of warnings because if we had a lower amount then we could leverage SwiftLint even further by adding more rules that would detecting new issues. While this would, in turn, raise the number of warnings yet again we would then be able to detect new problematic cases in real-time.
+Regardless of what we chose, we should still strive to reduce the tremendous amount of warnings our codebase currently has.
--- a/Cookbook/Proposals/AdditionalRulesForDanger.md
+++ b/Cookbook/Proposals/AdditionalRulesForDanger.md
@@ -9,14 +9,15 @@ The goal of this proposal is to introduce several new rules for Danger and/or Sw
 
 ## Motivation
 
-We have a large and particularly opinionated codebase (one could say "avant-gard" even) with significant deprecated code to address.
-New engineers face a somewhat steep learning curve before mastering our practices and even experienced engineers make mistakes, so these rules would surely benefit everyone. These rules are already known to us - we apply them on our PR reviews on a daily basis, after all - but, nevertheless, why rely on memory alone when we can automate them?
+We have a large and particularly opinionated codebase (one could say "avant-gard" even) with a significant amount of deprecated code to address. New engineers face a somewhat steep learning curve before mastering our practices and the fact that we have at times two or three ways of addressing a particular problem (mostly due to legacy code) only brings more confusion to the mix.
+
+SwiftLint and Danger are two invaluable tools that we can continue to leverage in order improve issue detection before these reach the `develop` branch. The rules I am about to introduce are already known to us - we apply them on our PR reviews on a daily basis, after all - but, nevertheless, why rely on memory alone when we can automate them?
 
 ## Proposed solution
 
 We will now go through the list of suggested rules and consider:
-    a) if it should be applied to our codebase at all
-    b) which tool should be we use to minimize further occurrences of the problem at hand
+    a) if they should be applied to our codebase (at all)
+    b) which tool (therefore, which approach) should we use to minimize further occurrences of the problem at hand
 
 ### Custom ViewModels in the test suite
 
@@ -26,13 +27,10 @@ This rule would be rather simple to enforce via Danger, using the following rege
 To do so, we should whitelist every file that matched `Tests.*`.
 
 Nevertheless, as of June 7th, 2019, we have 254 occurences of this rule in 251 files. 
-As such, I believe this rule should only be enforced via Danger because using SwiftLint to monitor this would:
-a) trigger 254 new warnings
-b) mean that we'd have to refactor all of them to get rid of the already tremendous amount of warnings we currently have (~500); this would require a rather significant effort for very little benefit.
 
-Enforcing the rule with Danger wouldn't cause those issues but we would only be able to detect new occurences of this anti-pattern since Danger will only inspect changes made in pull requests.
+As such, I believe this rule should only be enforced via Danger because using SwiftLint to monitor this would trigger 254 new warnings  (obviously). Since we already have a tremendous amount of warnings (~500), we'd have to refactor all of these new warnigs just to get back to our starting point. This would require a rather significant effort for very little benefit.
 
-Therefore, I would suggesting using Danger to enforce it.
+On the other hand, enforcing the rule with Danger wouldn't cause those issues but we would only be able to detect new occurences of this anti-pattern since Danger will only inspect changes made in pull requests.
 
 ### Monitoring ViewController lifecycle directly instead of relying on `ScreenLifecycleEvent`
 
--- a/Cookbook/Proposals/AdditionalRulesForDanger.md
+++ b/Cookbook/Proposals/AdditionalRulesForDanger.md
@@ -89,7 +89,7 @@ In this particular case I believe we could use both approaches actually.
 
 ### Detect Current usage outside Builders et al
 
-`Current` is a very powerful concept. Since its usage is only allowed in a very restrict number of places we should enable automatic monitoring in order to prevent it from being abused throughout our codebase.
+`Current` is a very powerful concept whose usage is only allowed in a very restrict number of places. As such, we should enable automatic monitoring in order to prevent it from being abused throughout our codebase. Nevertheless, Current is still being used in a lot of places (598 results across 144 files) and we should go through every case individually before commiting to a decision:
 
 `Current` should only be allowed in:
 - `AppDelegate.swift`
@@ -98,12 +98,12 @@ In this particular case I believe we could use both approaches actually.
 - Builders (using the `*Builder(\+.*).swift` regex)
 - `*AppConfiguration.swift`
 - AppDependencies
-- Tests, [since Current is using a mock version of itself in `SnapshotTestCase.swift`'s `setUp` method](https://github.com/Babylonpartners/babylon-ios/pull/7806/files#diff-74895f4da31ec40d83d342bf612540b4R17)
 
 In order to detect this, SwiftLint would likely be our best bet since it can easily detect these issues in real-time and warn us.
-Here is a possible implementation, as suggested by Olivier Halligon:
+Here is a possible implementation:
 
-```custom_rules:
+```
+custom_rules:
   world_current:
     excluded: # Files in which it's OK to use Current
      - ".*Builder(\+.*)?\\.swift"
@@ -114,36 +114,61 @@ Here is a possible implementation, as suggested by Olivier Halligon:
     name: "Current:World usage"
     regex: "\wCurrent\."
     message: "You should only use Current in Builders"
-    severity: warning```
+    severity: warning
+```
 
-Nevertheless, Current is still being used in a lot of places (598 results across 144 files) and we should go through every case individually before commiting to a decision:
+However, we have numerous cases that need to be analyzed one by one:
 
-Very-probably not OK in:
  - ReceiptRenderer (locale, tz)
+
  - *ViewModel
    - CountryCodePickerViewModel (locale)
    - MetricDetailsViewModel (now)
    - AvatarViewModel (default params for init: calendar, now)
    - AddAddressViewModel ( |> analytics.track)
+ 
  - AdditionalInfoForm
  - VisualDependencies.swift (restrictedAgeDatePickerStyle)
  
   - *FlowController (abTestingService)
    - PrescriptionFlowController
    - HomeFlowController
- - *ViewController: viewWillAppear: Current.analyticsService.track
+
+ - *ViewController:
    - MapViewController
    - IntroViewController
- - FocusedChatViewController (locale)
+   - FocusedChatViewController (locale)
  
  - Debug*
     - DebugHomeRenderer (renderVisualLanguageSwitch, check current VL)
     - DebugSignUpAccountGenerator
+
  - Other
-   - MockReceiptDTO: date = ...(locale, tz)
-   - BabylonBoxViewController (init(..., appearance = Current....))
+   - MockReceiptDTO (locale, tz)
+   - BabylonBoxViewController (appearance)
    - OpeningHoursFormatter
    - LocalNotificationService: extension DebugAction, check abTestingService
+
+
+### Mutating Current in `setUp()` but not restoring it in `tearDown()`
+
+Every object (susceptible to global mutation) that is mutated in `setUp()` must be restored in `tearDown()`.
+
+Mutations in `Current` in particular are the prime suspects of a number of flaky tests that have been recently affecting our test suite. [While some mitigations have been implemented to minimize this problem](https://github.com/Babylonpartners/babylon-ios/pull/7806/files#diff-74895f4da31ec40d83d342bf612540b4R17), there are still plenty of cases through which Current could be modified and our team would be none the wiser.
+
+In order to detect this problem we have to use Danger since SwiftLint only accepts regular expressions.
+We would have to whitelist every test file (`.*Tests.*`) and use Danger to manually parse the test line-by-line until the `setUp()` function was found. 
+
+If found, we need to make a counter that increases by one each time a curly bracket and does the reverse when a closing curly bracket appears. Once the counter reaches zero again, then we've successfully delimited the contents of the `setUp()` function and we'd use the following regexes to detect changes in Current: 
+a) `Current\..*\ = `.
+b) `Current.changeTheme\(.*`
+c) `Current.set\(.*`
+
+The algorithm is then be executed once more but for the `tearDown()` method instead.
+If the number of changes inside each function matches, then no issue would be found. If not, then Current appeared to have been mutated but not restored; we'd issue a warning for the lines in which the algorithm detected a change and alert the developers in GitHub using Danger.
+
+Nevertheless, this rule is rather naïve because we have no actual way of detecting if Current was reset or not; that would require compiling and running the actual Swift code. Therefore, this approach is not a silver bullet or anything remotely similar but it would still be useful for one use case in particular: should the developer forget to restore Current, this rule will alert us of this situation. Furthermore, it can easily be side-stepped if the mutations and resets are done in auxiliar methods instead.
+
 
 ### Translation errors
 
@@ -153,11 +178,10 @@ As such, I'd like to suggest another rule for Danger that would:
     1. detect new static String variables in every Localizable.swift file.
     2. if we detected a new static variable we would then scan the localization.string files and look for that key; if we didn't find them that would mean that there was probably a typo and the actual translation would never  appear on-screen (only the key itself).
 
-Having said that, our handling of L10n could definitely be improved 
-
 Furthermore, the reverse situation would also be problematic (albeit at a different scale): a key in the `.strings` file without an equivalent constant in its Swift counterpart would likely mean that the developer had added unused strings to the project. This would also be detectable using Danger.
 
 We would need to whitelist `Localizable.*\.swift` as well as the actual `Localizable.string` files in order to implement both rules.
+
 
 ### Feedbacks must have the `when` prefix
 
--- a/Cookbook/Proposals/AdditionalRulesForDanger.md
+++ b/Cookbook/Proposals/AdditionalRulesForDanger.md
@@ -94,7 +94,7 @@ Every object (susceptible to global mutation) that is mutated in `setUp()` must 
 
 Mutations in `Current` in particular are the prime suspects of a number of flaky tests that have been recently affecting our test suite. [While some mitigations have been implemented to minimize this problem](https://github.com/Babylonpartners/babylon-ios/pull/7806/files#diff-74895f4da31ec40d83d342bf612540b4R17), these are restricted to snapshot tests for the time being. Ideally, this solution should be extended to Unit and UI Test cases but that lies outside the scope of this proposal.
 
-Until that is implemented, we could alternatively monitor mutations in Current by counting the number of times Current is changed in the `setup()` and `tearDown()` methods. We would have to whitelist every test file (`.*Tests.*`) and use Danger to manually parse the test line-by-line.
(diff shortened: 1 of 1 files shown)


AliSoftware (commented):
Please give more details about the rules and the specs about them.
I think for most of the rules many will agree they could be useful so there won't be much debate probably, but there's a big need for discussion about the details specs.

For example if you pitch a rule about detecting that Current should only be used in builders, if you dig a little deeper (which is the point of writing a proposal, for the author to think about all the implications and the solution in more details) you realize that it's a little too light to describe the exact rule to implement and that we need to specify it more in depth. Eg we should make sure it's not triggered in AppDelegate as Current is initialized here, probably AppConfiguration too… should we whitelist or blacklist…

Having a quick draft, even if not perfect, of the danger or SwiftLint rule and how it will be shaped helps a lot forcing you to think more in depth about all those questions that are the real things we'll want to discuss when reviewing the proposal as a team.

Cookbook/Proposals/AdditionalRulesForDanger.md
```diff
@@ -0,0 +1,131 @@
...
+
+We'd have to restrict this rule to Builders and FlowControllers only, therefore we'd have to whitelist filenames that matched `.*Builder.swift` and `.*FlowController.swift`.
+
+This rule could be enforced with either tool but I'd recommend using SwiftLint for this since due to the low number of warnings it would trigger in the current state of our database. Also, I believe we should try to use Danger for particularly important warnings in order not to spam the pull request with comments.
+
+
+### Deprecated calls of DesignLibrary components (aka pre-NVL components)
+
+Given that NVL is scheduled for release within two weeks, using NVL components is pivotal to Babylon's overall look and feel.
+With the exception of the empty space component (even though it has its own NVL equivalent), every other pre-NVL component should trigger a warning.
```
MichalTKwiecien: I think we were using this because there was no NVL equivalent - if we have it now let's just switch to it and do not make any exceptions.

MichalTKwiecien: Should only be checking `Renderer`s for those or ignoring `BabylonUI` to no trigger warnings for proper usage.

```suggestion
With the exception of the empty space component (even though it has its own NVL equivalent), every other pre-NVL component used inside Renderer should trigger a warning.
```

NSMyself: Well said, thanks 👍

MichalTKwiecien (approved):

Cookbook/Proposals/AdditionalRulesForDanger.md
```diff
@@ -0,0 +1,133 @@
...
+            break
+        }
+    }
+```
+
+As of June 7th, 2019, we still have 16 results in 13 different files.
+
+In order to detect this, we could simply monitor occurences of the following regex:
+
+`\.signal\(for: \#selector\(UIViewController\.`
```
dmcrodrigues: This only addresses part the problem, we have a reasonable number of custom view controllers doing a similar logic too.

Cookbook/Proposals/AdditionalRulesForDanger.md
```diff
@@ -0,0 +1,133 @@
...
+
+If found, we need to make a counter that increases by one each time a curly bracket and does the reverse when a closing curly bracket appears. Once the counter reaches zero again, then we've successfully delimited the contents of the `setUp()` function and we'd use the following regex to detect changes in Current: `Current\..*\ = `.
+
+The algorithm is then be executed once more but for the `tearDown()` method instead.
+If the number of changes inside each function matches, then no issue would be found. If not, then Current appeared to have been mutated but not restored; we'd issue a warning for the lines in which the algorithm detected a change and alert the developers in GitHub.
+
+Nevertheless, this rule is rather naïve because we have no actual way of detecting if Current was reset or not; that would require compiling and runniing the actual Swift code. Therefore, this approach is not a silver bullet or anything remotely similar but it would still be useful for one use case in particular: should the developer forget to restore Current, this rule will alert us of this situation.
```
dmcrodrigues:
```suggestion
Nevertheless, this rule is rather naïve because we have no actual way of detecting if Current was reset or not; that would require compiling and running the actual Swift code. Therefore, this approach is not a silver bullet or anything remotely similar but it would still be useful for one use case in particular: should the developer forget to restore Current, this rule will alert us of this situation.
```

Cookbook/Proposals/AdditionalRulesForDanger.md
```diff
@@ -0,0 +1,133 @@
...
+Translations are particularly hard to catch during PR code reviews, especially if a lot of new phrases are added at once.
+While, in theory, we have a fastlane set up to deal with these translations there is always the risk that developers mistype the localization key when setting up a new static var.
+
+As such, I'd like to suggest another rule for Danger that would:
+    1. detect new static String variables in every Localizable.swift file.
+    2. if we detected a new static variable we would then scan the localization string file (`Babylon/Supporting Files/Localization/en.lproj/Localizable.strings`) and look for that key
+
+We would need to whitelist `Localizable.*\.swift` in order to make this work.
+
+### Feedbacks must have the `when` prefix
```
dmcrodrigues: I'm not sure about this one, we have Feedbacks which start their effects when the system is initialised and not depending on a certain state so I think it would be misleading enforcing this prefix to them.

dmcrodrigues (approved):
Looks good, I think it would be beneficial to have this set of rules to kind of enforce but also promote our own conventions.

Cookbook/Proposals/AdditionalRulesForDanger.md
```diff
@@ -0,0 +1,133 @@
...
+With the exception of the empty space component (even though it has its own NVL equivalent), every other pre-NVL component used inside Renderer should trigger a warning.
+
+The following regex currently yields 1333 results in 216 files: `Component\.(?!EmptySpace)\b`.
+The sheer number of warnings immediately excludes SwiftLint; Danger would be our only option.
+
+Nevertheless, this would require that everyone relied exclusively on NVL components and I am not sure this is the case right now.
+
+
+### Mutating Current in `setUp()` but not restoring it in `tearDown()`
+
```
chitrakotwani: I was under the impression that `Current` should not be done at all even if it is restored in `tearDown()` 

While restoring it in `tearDown()` mitigates the problem, it does not eliminate it.

The sequence of events can be : 
`Test1.setUp(){Set UK locale on Current}`
`Test2.setUp(){Set USA locale on Current}` 
`Test1.test_xyz  {Expect time to be rendered according to UK locale} `  ❌ 

I think this rule can be changed to not allow mutating Current at all.

Cookbook/Proposals/AdditionalRulesForDanger.md
```diff
@@ -0,0 +1,133 @@
...
+### Deprecated calls of DesignLibrary components (aka pre-NVL components)
+
+Given that NVL is scheduled for release within two weeks, using NVL components is pivotal to Babylon's overall look and feel.
+With the exception of the empty space component (even though it has its own NVL equivalent), every other pre-NVL component used inside Renderer should trigger a warning.
+
+The following regex currently yields 1333 results in 216 files: `Component\.(?!EmptySpace)\b`.
+The sheer number of warnings immediately excludes SwiftLint; Danger would be our only option.
+
+Nevertheless, this would require that everyone relied exclusively on NVL components and I am not sure this is the case right now.
+
```
chitrakotwani: Enforcing `Danger` to apply this rule will be too late. Significant effort and time will have been used till then to create a screen with deprecated components. (With my limited understanding of what NVL encompasses)

Cookbook/Proposals/AdditionalRulesForDanger.md
```diff
@@ -0,0 +1,133 @@
...
+## Proposed solution
+
+We will now go through the list of suggested rules and consider:
+    a) if they should be applied to our codebase (at all)
+    b) which tool (therefore, which approach) should we use to minimize further occurrences of the problem at hand
+
+### Custom ViewModels in the test suite
+
+Custom ViewModels should not be defined as part of our tests; we should rely on `StubViewModel`s (from `BabylonSnapshotTestUtilities`) instead (although, in theory, setting up custom ViewModels could be useful in particular circumstances).
```
chitrakotwani: I don't have much context here. Why and what are these? In which circumstances can they be useful?

chitrakotwani (commented):
Much needed undertaking! 👏

Cookbook/Proposals/AdditionalRulesForDanger.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Mutating Current in `setUp()` but not restoring it in `tearDown()`
+
+The rule is simple: every object (susceptible to global mutation) that is mutated in `setUp()` must be restored in `tearDown()`.
+Mutations in `Current` in particular are the prime suspects of a number of flaky tests that have been recently affecting our test suite.
+
+In order to detect this problem we have to use Danger since SwiftLint only accepts regular expressions.
+We would have to whitelist every test file (`.*Tests.*`) and use Danger to manually parse the test line-by-line until the `setUp()` function was found. 
+
+If found, we need to make a counter that increases by one each time a curly bracket and does the reverse when a closing curly bracket appears. Once the counter reaches zero again, then we've successfully delimited the contents of the `setUp()` function and we'd use the following regex to detect changes in Current: `Current\..*\ = `.
```
AliSoftware: We need to also detect `Current.changeTheme` or other `Current.setXXX`, not just `Current = `

Cookbook/Proposals/AdditionalRulesForDanger.md
```diff
@@ -0,0 +1,133 @@
...
+Nevertheless, this rule is rather naïve because we have no actual way of detecting if Current was reset or not; that would require compiling and running the actual Swift code. Therefore, this approach is not a silver bullet or anything remotely similar but it would still be useful for one use case in particular: should the developer forget to restore Current, this rule will alert us of this situation.
+
+### Translation errors
+
+Translations are particularly hard to catch during PR code reviews, especially if a lot of new phrases are added at once.
+While, in theory, we have a fastlane set up to deal with these translations there is always the risk that developers mistype the localization key when setting up a new static var.
+
+As such, I'd like to suggest another rule for Danger that would:
+    1. detect new static String variables in every Localizable.swift file.
+    2. if we detected a new static variable we would then scan the localization string file (`Babylon/Supporting Files/Localization/en.lproj/Localizable.strings`) and look for that key
```
AliSoftware: 1. I think we need to do that in both directions. A key in `Localization*.swift` but not present in `Localizable.strings` is the biggest problem, but a key in the `.strings` file without a constant in the `.swift` file also means that we pulled a strings from Lokalise that we in fact don't use in this PR and is probably part of a different, unrelated feature, so we should warn about this too
2. What about other `.strings` files, especially `TargetSpecificLocalizable.strings`?

Also, I think we need to address the problem about handling the L10n across our modules and in tests. e.g. if that's what we've decided today (I'm not even sure what the actual concensus is today tbh), ensure that any reference to a `Localizable.strings` in a test target of a project file is just a reference to the `Localizable.strings` file of Babylon, so containing the same keys since it's the same file, and is not a separate `.strings` file where keys have been copy/pasted instead.

I think all that might even require a TechnicalDocument on its own jsut to describe how to do L10n on Babylon (using the dedicated lane to pull translations, clean up translations unrelated to the PR, in which files to add the translations, in which file to add the constants, depending on the feature they're related to (`LocalizationUI.swift` vs `LocalizationOctopus.swift`, etc), how to reference localized strings from tests by switching the bundle...

And once all those rules are clearly written black on white, the Danger rule will become much clearer too. e.g. I can already see a new rule we should have, similar to the one about `Current`, to ensure the bundle is reset in `tearDown` if it was changed in `setUp`...

Cookbook/Proposals/AdditionalRulesForDanger.md
```diff
@@ -0,0 +1,133 @@
...
+
+While our codebase itself would not suffer any impact at all, our code review process would now have additional comments that would warn us of a particular situation. Should a PR contain numerous anti-patterns that break our rules the bot could potentially spam the PR with these warnings, depending on the way Danger's rules are implemented.
+
+Furthermore, we would be increasing Danger's processing time - although for the time being this is nearly negligible (roughly 4 seconds) - so this is a minor drawback that we could easily live with, in my honest opinion.
+
+## Alternatives considered
+
+* Implement a subset of those rules (and determine the most adequate tool to detect infractions for each rule).
+
+* Don't implement these rules at all
```
AliSoftware: * Fix some of the existing warnings (maybe some of them are in great number but relatively simple to fix with search/replace for example?) first, so that we could then re-consider make new warnings like SwiftLint's be more visible and thus more useful.

* Use SwiftLint for some of the rules mentioned above, even if we already have violations for them in our current codebase, but keep them as warnings. Would increase the number of warnings again, but would also allow us to see warnings in any new code we write immediately, instead of only at PR time. Would also allow us to see warnings on existing code when we work on an existing file and hopefully take that occasion to refactor and fix the warning while working on our feature (e.g. if I work on an existing, legacy screen, if I only get the warning via Danger I might not be aware that it's legacy and I should take the occasion to refactor it a bit while I'm at it... until I've done my ticket and open the PR)

* Use SwiftLint and make the violations be error, but use `// swiftlint:disable` on existing code already violating the rules, so that only the new code gets the warnings. (I'm not fan of this solution, since it would risk to legitimize the use of `swiftlint:disable` and end up defeating the purpose of SwiftLint in the long run, but had to mention it for the sake of argument)

Cookbook/Proposals/AdditionalRulesForDanger.md
ilyapuchka: @dmcrodrigues then it can be called `whenInitialized` or `whenStarted` or something like that? I'm +1 for enforcing common naming patterns in general. Similarly it would be nice to enforce common pattern in Actions/Events, i.e. that they start or should not start with `did` and they don't match state names. But it might be trickier as we are dealing with natural language processing here and a simple regex can be too limiting.

ilyapuchka (commented):
I'm generally +1 for the direction of ensuring that our codebase is uniform, some implementation details need to be discussed but otherwise good ideas 👍

Cookbook/Proposals/AdditionalRulesForDanger.md
```diff
@@ -0,0 +1,133 @@
...
+Nevertheless, as of June 7th, 2019, we have 254 occurences of this rule in 251 files. 
+
+As such, I believe this rule should only be enforced via Danger because using SwiftLint to monitor this would trigger 254 new warnings  (obviously). Since we already have a tremendous amount of warnings (~500), we'd have to refactor all of these new warnigs just to get back to our starting point. This would require a rather significant effort for very little benefit.
+
+On the other hand, enforcing the rule with Danger wouldn't cause those issues but we would only be able to detect new occurences of this anti-pattern since Danger will only inspect changes made in pull requests.
+
+### Monitoring ViewController lifecycle directly instead of relying on `ScreenLifecycleEvent`
+
+With the introduction of `ScreenLifecycleEvent`, developers should no longer manually monitor the viewController's lifecycle events.
+Instead, this kind of monitoring can now easily be done via the `send(_: ScreenLifecycleEvent)` method.
```
ilyapuchka: tbh I don't know what it is 🤷‍♂

Cookbook/Proposals/AdditionalRulesForDanger.md
AliSoftware: I'm all for using a code generation tool for that.

But we'll need a lot of cleanup of our existing codebase first I think, since as of now, the name of the constants we manually create and the corresponding Localization.strings keys they point to don't match names. So when we'll replace that with codegen, the name of our constants will completely change to match the keys used in the `.strings` and we'll have a lot of updating to do on all the code using them. Which means it's not a trivial task given our legacy and number of existing keys (we might consider a progressive transition though)

Since idk when we'll get there (transition to codegen), that might be an intermediate rule worth having in the meantime

ilyapuchka: we'll have to do that anyway. if we need an intermediate solution that will help with it so that we can do that gradually rather than all in one go I think it's good. But again even if we start to code generate these strings today I don't think anything enforces us to use generated code all in, we can keep existing structures and just mark it as deprecated. Lots of options here.

Cookbook/Proposals/AdditionalRulesForDanger.md
NSMyself: Good point. I will adjust the proposal, this seems like a good trade-off.

Cookbook/Proposals/AdditionalRulesForDanger.md
NSMyself: Fair points, will revisit this section.

ilyapuchka:
Can I suggest to add the rule that would validate the PR metadata itself, its title at least 🙏 Currently we have a bot that will fail if there is no properly formatted ticket number in the title, but it does not explain the failure. We can extend this further to validating that if there are no test changes in the PR then there should be a note in the PR description about that

Cookbook/Proposals/AdditionalRulesForDanger.md
ilyapuchka: We used to define stubs for view models in the snapshot tests as they are required to create the screens. Now we have a generic `StubViewModel` for that and we don't need custom implementations, but they are still being used somewhere. I'd prefer this to be solved with deprecation annotation though. We can for example write a variant of `verifyScreenForAllSizesAndVisualLanguages` that would only accept `StubViewModel` for `viewModel` parameter and deprecate the variant that is generic over type of ViewModel. @NSMyself  what do you think?

NSMyself: @chitrakotwani apologies for not replying, I couldn't do it at the time and completely forgot about it. 

Basically, by using custom ViewModels instead of stubbing the ones we already have technically we aren't testing the actual ViewModels, but the ones we've just created for the purpose of this test. Depending on the actual implementation, that may not be the exact same thing.

Regarding the circumstances in which this approach could be useful, well:
a) we may need to test something that because of it's nature and/or complexity is not easily achievable and, as such, using a custom ViewModel could theoretically help (although that fact could indicate that the VM needs to be refactored, but that's another story).

b) a concrete example: when testing a Renderer and we need our VM to hold some value that we cannot define via the initializers and/or easily get the state machine to the point where we need it to be.

NSMyself: @ilyapuchka i think that's a great idea. I need to revisit this proposal and update it because in some cases marking the method via a deprecation annotation is indeed the right approach imho. Will do so in the following days.

Cookbook/Proposals/AdditionalRulesForDanger.md
NSMyself: Actually, @daniloaliberti pointed out that this issue is [already covered](https://github.com/Babylonpartners/babylon-ios/pull/7806/files#diff-74895f4da31ec40d83d342bf612540b4R17) 👍

I will remove this section entirely

NSMyself: > Can I suggest to add the rule that would validate the PR metadata itself, its title at least 🙏 Currently we have a bot that will fail if there is no properly formatted ticket number in the title, but it does not explain the failure. We can extend this further to validating that if there are no test changes in the PR then there should be a note in the PR description about that

I think that's a good idea @ilyapuchka but we should perhaps make another proposal because this one already has a lot of stuff for just a 15m meeting

NSMyself:
I've just up updated the proposal after considering feedback from @AliSoftware, @MichalTKwiecien, @dmcrodrigues, @chitrakotwani and @ilyapuchka.

Additionally, I also added another section after chatting with @daniloaliberti and @AliSoftware regarding monitoring the usage of Current outside builders. That section was made by the three of us, which was why I added them as authors as well.

Cookbook/Proposals/AdditionalRulesForDanger.md
```diff
@@ -0,0 +1,208 @@
...
+
+The sheer number of warnings immediately excludes SwiftLint, which leaves us with:
+a) marking these methods as deprecated
+b) using Danger to raise a comment whenever a PR is raised
+
+In this particular case I believe we could use both approaches actually.
+
+### Detect Current usage outside Builders et al
+
+`Current` is a very powerful concept whose usage is only allowed in a very restrict number of places. As such, we should enable automatic monitoring in order to prevent it from being abused throughout our codebase. Nevertheless, Current is still being used in a lot of places (598 results across 144 files) and we should go through every case individually before commiting to a decision:
```
daniloaliberti:
```suggestion
`Current` is a very powerful concept whose usage is only allowed in a very restrict number of places. 
Check [Current Guide](./Technical-Documents/Current-guide.md) . As such, we should enable automatic monitoring in order to prevent it from being abused throughout our codebase. Nevertheless, Current is still being used in a lot of places (598 results across 144 files) and we should go through every case individually before commiting to a decision:
```

Cookbook/Proposals/AdditionalRulesForDanger.md
```diff
@@ -0,0 +1,208 @@
...
+   - BabylonBoxViewController (appearance)
+   - OpeningHoursFormatter
+   - LocalNotificationService: extension DebugAction, check abTestingService
+
+
+### Mutating Current in `setUp()` but not restoring it in `tearDown()`
+
+Every object (susceptible to global mutation) that is mutated in `setUp()` must be restored in `tearDown()`.
+
+Mutations in `Current` in particular are the prime suspects of a number of flaky tests that have been recently affecting our test suite. [While some mitigations have been implemented to minimize this problem](https://github.com/Babylonpartners/babylon-ios/pull/7806/files#diff-74895f4da31ec40d83d342bf612540b4R17), there are still plenty of cases through which Current could be modified and our team would be none the wiser.
```
daniloaliberti:
```suggestion
Mutations in `Current` in particular are the prime suspects of a number of flaky tests that have been recently affecting our test suite. [While some mitigations have been implemented to minimize this problem](https://github.com/Babylonpartners/babylon-ios/pull/7806/files#diff-74895f4da31ec40d83d342bf612540b4R17), there are still plenty of cases through which Current could be modified and our team would be none the wiser, like extending the linked solution to base Unit and UI Test cases.
```

NSMyself:
After considering the fact that our proposal meeting only lasts 15m I decided to move the new section (Current outside builders) to a new proposal

Cookbook/Proposals/AdditionalRulesForDanger.md
```diff
@@ -0,0 +1,151 @@
...
+
+Translations are particularly hard to catch during PR code reviews, especially if a lot of new phrases are added at once. While, in theory, we have a fastlane set up to deal with these translations there is always the risk that developers mistype the localization key when setting up a new static var.
+
+As such, I'd like to suggest another rule for Danger that would:
+    1. detect new static String variables in every Localizable.swift file.
+    2. if we detected a new static variable we would then scan the localization.string files and look for that key; if we didn't find them that would mean that there was probably a typo and the actual translation would never  appear on-screen (only the key itself).
+
+Furthermore, the reverse situation would also be problematic (albeit at a different scale): a key in the `.strings` file without an equivalent constant in its Swift counterpart would likely mean that the developer had added unused strings to the project. This would also be detectable using Danger.
+
+We would need to whitelist `Localizable.*\.swift` as well as the actual `Localizable.string` files in order to implement both rules.
```
daniloaliberti: let's not forget the `TargetSpecificLocalizable.strings`

daniloaliberti: also let's not forget what @dmcrodrigues mentioned: SwiftGen to do not rely on developers

karabatov (approved):

NSMyself:
Outcome of this meeting:

`Detecting custom ViewModels in snapshot tests`: ✅
`ScreenLifecycleEvent instead of observing VC lifecycle directly`: ✅
`Deprecated calls of DesignLibrary components`: ✅
`Mutating Current in `setUp()` but not restoring it in `tearDown()`: ❌
`Feedbacks must have the `when` prefix`: ✅

Cookbook/Proposals/AdditionalRulesForDanger.md
NSMyself: @daniloaliberti @dmcrodrigues I added some tweaks regarding the localization rule, please have a look when you've time

Cookbook/Proposals/AdditionalRulesForDanger.md
```diff
@@ -0,0 +1,155 @@
...
+    1. detect new static String variables in every Localizable.swift file.
+    2. if we detected a new static variable we would then scan the localization.string files and look for that key; if we didn't find them that would mean that there was probably a typo and the actual translation would never  appear on-screen (only the key itself).
+
+Furthermore, the reverse situation would also be problematic (albeit at a different scale): a key in the `.strings` file without an equivalent constant in its Swift counterpart would likely mean that the developer had added unused strings to the project. This would also be detectable using Danger. 
+
+On the other hand:
+- we have no reliable way of mapping the association between each target and their respective strings
+- this process should be automated by using SwiftGen instead
+
+We would need to whitelist `Localizable.*\.swift` as well as the `Localizable.string` and its corresponding target-specific files in order to implement both rules.
```
daniloaliberti:
```suggestion
We would need to whitelist `Localizable.*\.swift` as well as the `Localizable.string` and its corresponding target-specific files (TargetSpecificLocalizable.strings) in order to implement both rules.
```

daniloaliberti (approved):
great one man!

PR #162, opened 2019-06-21
Remove UIFlow from the list of OSS projects

karabatov:
The `UIFlow` project by @diegopetrucci still lives in its private repo and thus is unreachable, and this has been flagged twice already.

This change is to remove it until it is either scrapped or open sourced.


mr-v (approved):

diegopetrucci (approved):
Should we remove the style guide, or link to the specific page here in the playbook? It was added here a long time ago, before the playbook was so big, and now I’m not sure it deserves such prominence.

Cc @RuiAAPeres

karabatov:
@diegopetrucci I can remove it for now as well :) As we basically have it in the Cookbook now.

PR #166, opened 2019-06-27
First pass on creating a feature that works with the backend

karabatov:
Such a document would have been extremely useful to me when I had just started, which is why I decided to write it.

I don't think this part of the app will change significantly, so it should be reasonably future-proof.

This is only a first pass, I'm intending to add sections on testing in the future and some strategies when working with more complicated data like e.g. using `decodeArraySafely`, subkeys without a response, form requests etc. 

This first section should be useful for newcomers, and in updates I want to collect some common practices which other people may find useful without the need to look for a place in the source where something is implemented in a certain way.

My intention with this version is that it is merged and is available while I'll be updating it with further PRs in the future.

### DIFF
--- a/Cookbook/Technical-Documents/BackendFeatureHowTo.md
+++ b/Cookbook/Technical-Documents/BackendFeatureHowTo.md
@@ -182,6 +182,30 @@ private static func whenLoading(
 
 Note that `citiesBusinessController` is provided as a parameter to the feedback function and doesn't require `self`.
 
+There is one last piece of the puzzle missing: how does the view model get the concrete class of the business controller? This involves something called a *builder*, which is responsible for constructing objects and providing the dependencies they need.
+
+### Builder
+A builder, usually defined as a `struct`, is used to construct what essentially are screens (i.e. view controllers) at runtime, which is why it's using concrete classes. During testing, view models, business controllers and renderers are tested separately, so there is no need for a builder to bring it all together.
+
+The relevant part of a builder, where we provide our new business controller to the view model, may be implemented like this:
+``` swift
+protocol CitiesBuildingMaterials {
+    let session: AuthenticatedAccessible
+}
+
+struct CitiesBuilder {
+    func make(with buildingMaterials: CitiesBuildingMaterials) -> UIViewController {
+        let session = buildingMaterials.session
+
+        let viewModel = CitiesViewModel(
+            citiesBusinessController: CitiesBusinessController(accessible: session)
+        )
+
+        …
+    }
+}
+```
+
 ### Feature structure
 The business controller, service and related data types should all be grouped in an `API` directory if they are all for the same single feature:
 ```
--- a/Cookbook/Technical-Documents/BackendFeatureHowTo.md
+++ b/Cookbook/Technical-Documents/BackendFeatureHowTo.md
@@ -16,7 +16,7 @@ Imagine some endpoint on the backend returns this piece of JSON:
 }
 ```
 
-First, we want to define some data types, to which we will map this JSON. This is known as a /response/.
+First, we want to define some data types, to which we will map this JSON. This is known as a *response*.
 
 ### Response
 A `Response` is a data type which wraps the whole response returned from a network call. In many cases it is unnecessary, just like in this example, where there is only one key. For the sake of this exercise, let's stick with it though. A later section will describe a way to extract the contents directly without creating a separate data type.
@@ -34,7 +34,7 @@ struct City: Equatable, Decodable {
 
 Note that we put `Decodable` directly on the structs in this simple case, but in general you'll want to put it in an extension because some customisation is needed when decoding the objects. This will be shown in a later section.
 
-When the response data type is ready, the next thing you'll want to create is a /service/. But before that, a few words about a /backend resource/.
+When the response data type is ready, the next thing you'll want to create is a *service*. But before that, a few words about a *backend resource*.
 
 ### Backend resource
 `BackendResource` is an abstraction over a network request, allowing to provide all the necessary context to actually perform it.
@@ -57,7 +57,7 @@ As you can see, it contains everything that the network layer may need to make a
 * request method
 * any additional headers we want to add
 
-A /service/ that we are about to define is responsible for generating a `BackendResource` for each request that we want to make.
+A *service* that we are about to define is responsible for generating a `BackendResource` for each request that we want to make.
 
 ### Service
 The purpose of a `Service` is to provide a `BackendResource` to execute network requests. As you may have guessed, this is also a simple struct, used as a namespace.
@@ -83,7 +83,7 @@ The URL is defined as a path, which is then constructed into a full URL with a h
 
 `ConcreteAuth` is the type of auth this endpoint expects as there are several, for different backends.
 
-Now that we have our `BackendResource`, we can define a /business controller/ which will tie everything together and make the network call.
+Now that we have our `BackendResource`, we can define a *business controller* which will tie everything together and make the network call.
 
 However, first we need to learn about accessing backend resources.
 
@@ -147,7 +147,7 @@ struct State {
 The entities we're loading are usually supplied through an `Event`:
 ```
 enum Event {
-    case didLoad([City])
+    case didLoad(CitiesResponse)
 }
 ```
 
@@ -155,8 +155,8 @@ When the view model gets an event, `reduce` is called and we can set the new val
 ```
 static func reduce(state: State, event: Event) -> State {
     switch event {
-    case let .didLoad(cities):
-        return state.with(set(\.cities, cities))
+    case let .didLoad(citiesResponse):
+        return state.with(set(\.cities, citiesResponse.cities))
     }
 }
 ```
--- a/Cookbook/Technical-Documents/BackendFeatureHowTo.md
+++ b/Cookbook/Technical-Documents/BackendFeatureHowTo.md
@@ -67,7 +67,7 @@ The purpose of a `Service` is to provide a `BackendResource` to execute network 
 As a rule, `GET` requests are named starting with `fetch`:
 ``` swift
 struct CitiesService {
-    static func fetchCitiesResource() -> BackendResource<Void, CitiesResponse, ConcreteAuth> {
+    static func fetchCitiesResource() -> BackendResource<Void, CitiesResponse, KongAuth> {
         let citiesPath = "/v1/cities"
         return BackendResource(
             path: .api(.core, citiesPath),
@@ -83,7 +83,7 @@ The `Service` usually provides `static` functions with parameters needed to crea
 
 The URL is defined as a path, which is then constructed into a full URL with a helper enum. This accommodates the use of different environments while not requiring any changes from the clients.
 
-`ConcreteAuth` is the type of auth this endpoint expects as there are several, for different backends.
+`KongAuth` is the concrete type of auth this endpoint expects as there is also `ClinicalAuth`, used for different backend endpoints.
 
 Now that we have our `BackendResource`, we can define a *business controller* which will tie everything together and make the network call.
 
@@ -100,7 +100,7 @@ func access<Request, Response>(_ resource: BackendResource<Request, Response, No
 You may have noticed `NoAuth` in the signature – this is only good for unauthenticated requests, like fetching data before the user logs in. After that, you'll likely want to use `AuthenticatedAccessible` which provides a slightly different method:
 
 ``` swift
-func access<Request, Response>(_ resource: BackendResource<Request, Response, ConcreteAuth>, with request: Request) -> SignalProducer<Response, CoreError>
+func access<Request, Response>(_ resource: BackendResource<Request, Response, KongAuth>, with request: Request) -> SignalProducer<Response, CoreError>
 ```
 
 Using either `Accessible` or `AuthenticatedAccessible`, you can perform the network request defined by a backend resource, and this is exactly what we do in the business controller.
--- a/Cookbook/Technical-Documents/BackendFeatureHowTo.md
+++ b/Cookbook/Technical-Documents/BackendFeatureHowTo.md
@@ -206,6 +206,8 @@ struct CitiesBuilder {
 }
 ```
 
+Note that the approach to builders will be changed with the “[Builders to functions](../Proposals/BuildersToFunctions.md)” proposal, and they will become free functions.
+
 ### Feature structure
 The business controller, service and related data types should all be grouped in an `API` directory if they are all for the same single feature:
 ```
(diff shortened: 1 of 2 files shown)


Cookbook/Technical-Documents/BackendFeatureHowTo.md
```diff
@@ -0,0 +1,200 @@
+# How to implement a feature that works with the backend
+Let's look at a minimal and simple example of working with the backend to get some items into our view model, before moving on to more difficult topics.
+
+## A simple example
+Imagine some endpoint on the backend returns this piece of JSON:
+```
```
AliSoftware: nit: you should use `json` after the 3 backticks to have JSON syntax highlighting of the snippet

Cookbook/Technical-Documents/BackendFeatureHowTo.md
```diff
@@ -0,0 +1,200 @@
...
+}
+```
+
+First, we want to define some data types, to which we will map this JSON. This is known as a *response*.
+
+### Response
+A `Response` is a data type which wraps the whole response returned from a network call. In many cases it is unnecessary, just like in this example, where there is only one key. For the sake of this exercise, let's stick with it though. A later section will describe a way to extract the contents directly without creating a separate data type.
+
+A response data type is a plain struct conforming to `Equatable`  (this is usually necessary for tests) and `Decodable`. We also define an additional data type for the city itself. You will want to define all the new data types necessary to describe a response:
+```
```
AliSoftware: nit: you should use `swift` after the 3 backticks to get Swift syntax highlighting of the snippet (and same with all snippets below)

Cookbook/Technical-Documents/BackendFeatureHowTo.md
```diff
@@ -0,0 +1,200 @@
...
+struct CitiesResponse: Equatable, Decodable {
+    let cities: [City]
+}
+
+struct City: Equatable, Decodable {
+    let name: String
+}
```
AliSoftware: I'm wondering, do we take the habit of defining these structs all at top level, or do we prefer declaring them as nested types?

```swift
struct CitiesResponse: Equatable, Decodable {
  struct City: Equatable, Decodable {
    let name: String
  }
  let cities: [City]
}
```

Cookbook/Technical-Documents/BackendFeatureHowTo.md
```diff
@@ -0,0 +1,200 @@
...
+struct CitiesBusinessController: CitiesBusinessControllerProtocol {
+    private let accessible: AuthenticatedAccessible
+
+    init(accessible: AuthenticatedAccessible) {
+        self.accessible = accessible
+    }
+
+    func fetch() -> SignalProducer<CitiesResponse, CoreError> {
+        let resource = CitiesService.fetchCitiesResource()
+        return accessible.access(resource)
```
AliSoftware: I'm confused, isn't `AuthenticatedAccessible`'s `access` method expecting 2 parameters? I was not sure what the `with request: Request` was for when I saw your description of the protocol in the previous section, and since now in this example it's not even used, I'm not even sure if it's a typo or something I missed?

Cookbook/Technical-Documents/BackendFeatureHowTo.md
```diff
@@ -0,0 +1,200 @@
...
+In order to trigger loading, we define a `Feedback` and transform a network response in order to conform to the signature of the effect:
+```
+{ _ -> SignalProducer<Event, NoError> in
+    citiesBusinessController.fetch()   // SignalProducer<CitiesResponse, CoreError>
+        .map(Event.didLoad)            // SignalProducer<Event, CoreError>
+        .replaceError(Event.didFail)   // SignalProducer<Event, NoError>
+}
+```
+
+Note that `citiesBusinessController` is provided as a parameter to the feedback function and doesn't requite `self`.
```
AliSoftware:
```suggestion
Note that `citiesBusinessController` is provided as a parameter to the feedback function and doesn't require `self`.
```

Cookbook/Technical-Documents/BackendFeatureHowTo.md
```diff
@@ -0,0 +1,200 @@
...
+      "name": "Bath"
+    }
+  ]
+}
+```
+
+First, we want to define some data types, to which we will map this JSON. This is known as a *response*.
+
+### Response
+A `Response` is a data type which wraps the whole response returned from a network call. In many cases it is unnecessary, just like in this example, where there is only one key. For the sake of this exercise, let's stick with it though. A later section will describe a way to extract the contents directly without creating a separate data type.
```
AliSoftware: I don't see that "later section" in the document about extracting without having to define a dedicated `struct CitiesResponse` data type though?

AliSoftware (commented):
Very interesting document! I'll probably never have to implement that myself but it is clear and was interesting to learn how all of this is architectured 👍

horothesun (approved):
Great introduction! It'll save a ton of time during onboarding 👍

Cookbook/Technical-Documents/BackendFeatureHowTo.md
karabatov: Ah right, I keep forgetting about it :)

Cookbook/Technical-Documents/BackendFeatureHowTo.md
karabatov: AFAIK generally it makes more sense to define them at the top level, as you keep passing around the deepest-nested types around (for rendering etc.), if they were defined as subtypes you'd have to have typealiases everywhere or it would be much more typing.

karabatov: I'll make a note about that.

Cookbook/Technical-Documents/BackendFeatureHowTo.md
karabatov: This is a convenience for when there is no `Request` (which is *quite* often). I'll make a note about it.

Cookbook/Technical-Documents/BackendFeatureHowTo.md
karabatov: Yeah, about that 😅 
I think I'll allude to more being written a little more subtly (just “this will be added later”) to get a more coherent document.

karabatov:
@AliSoftware I made some changes based on your comments :)

- [x] Snippet types specified
- [x] Note about defining structs on the top level
- [x] Note about `accessible.access` second parameter
- [x] Typo fixed
- [x] Removed references to “later sections”, I'll put them back when I actually write them

Cookbook/Technical-Documents/BackendFeatureHowTo.md
```diff
@@ -0,0 +1,204 @@
...
+}
+
+struct City: Equatable, Decodable {
+    let name: String
+}
+```
+
+Note that we put `Decodable` directly on the structs in this simple case, but in general you'll want to put it in an extension because some customisation is needed when decoding the objects.
+
+As a rule, all the structs are defined at the top level rather than as subtypes (i.e. `CitiesResponse.City`). Most of the structs are used throughout the app, all the way up to the renderers, and it would be inconvenient to always use a fully-qualified name or set up type aliases.
```
AliSoftware: 👍

Cookbook/Technical-Documents/BackendFeatureHowTo.md
```diff
@@ -0,0 +1,204 @@
...
+    func fetch() -> SignalProducer<CitiesResponse, CoreError> {
+        let resource = CitiesService.fetchCitiesResource()
+        return accessible.access(resource)
+    }
+}
+```
+
+As simple as that: we get a backend resource from `CitiesService` , access it with the provided `AuthenticatedAccessible` and return the result.
+
+Note that we didn't specify the second parameter (request) for `accessible.access(…)`. Earlier we defined the request type as `Void`, which allows us to use a single-parameter convenience method on `AuthenticatedAccessible`, it simply calls `access(resource, with: ())`.
```
AliSoftware: Makes sense now 👍

AliSoftware (approved):
Looking great now 👍 

Maybe now that you removed the in-document paragraph about "later sections" you could now add a section at the very end of the document like this?

```md
## Future documentation

This is is only an initial pass on the topic. Feel free to help improve that documentation. Ideas of topics to address include:

 - Provide an example of a resource access with a non-`Void` `Request`
 - Explain how we can avoid creating a `Decodable` dedicated type to decode the request for simpler responses
 - ...
```

karabatov:
@AliSoftware I have a local todo on what I want to add to it, I'll be doing that soonish, so I don't feel there's a need to put future plans on the doc itself.

Cookbook/Technical-Documents/BackendFeatureHowTo.md
```diff
@@ -0,0 +1,204 @@
...
+```
+
+First, we want to define some data types, to which we will map this JSON. This is known as a *response*.
+
+### Response
+A `Response` is a data type which wraps the whole response returned from a network call. In many cases it is unnecessary and you can extract the values from a single key directly, just like in this example, where there is only one key. For the sake of this exercise, let's stick with it though.
+
+A response data type is a plain struct conforming to `Equatable`  (this is usually necessary for tests) and `Decodable`. We also define an additional data type for the city itself. You will want to define all the new data types necessary to describe a response:
+``` swift
+struct CitiesResponse: Equatable, Decodable {
```
TheAdamBorek: You don't need that wrapper. In the BackendResource you can put the response handler with `Parser.decode(Parser.parse(data, key: "cities")` inside.

I know you have written that for the sake of this exercise you choose to stick with it, however it's something I would mark on PR to have less types if possible.

karabatov: I say that in the previous paragraph :) I will make an example of how to invoke `Parser` with a subkey later.

Cookbook/Technical-Documents/BackendFeatureHowTo.md
```diff
@@ -0,0 +1,204 @@
...
+        return state.with(set(\.cities, citiesResponse.cities))
+    }
+}
+```
+
+Now `state.cities` contains the values we fetched and parsed from the network!
+
+In order to trigger loading, we define a `Feedback` and transform a network response in order to conform to the signature of the effect:
+``` swift
+{ _ -> SignalProducer<Event, NoError> in
```
TheAdamBorek: can you also put the function signature/definition? 
it just feels odd to not have it.

karabatov: Yeah, this was extracted from a convoluted example, I think I can define a super simple feedback here and have its full signature ✨

dmcrodrigues: Agree, we should have a function like `whenLoading(cities:)` which aligns with our conventions.

dmcrodrigues (approved):
This looks good @karabatov, one detail or two but great initiative!

Cookbook/Technical-Documents/BackendFeatureHowTo.md
```diff
@@ -0,0 +1,204 @@
...
+    │   ├── CitiesRendererSnapshotTests.swift
+    ├── API
+    │   ├── CitiesBusinessController.swift
+    │   ├── CitiesService.swift
+    │   ├── CitiesResponse.swift
+    │   └── City.swift
+    ├── CitiesViewModel.swift
+    ├── CitiesRenderer.swift
+    ├── CitiesFlowController.swift
+    └── CitiesBuilder.swift
```
RuiAAPeres: You mention the `Builder`, but no description regarding it. Could you put a small reference to it and mention the proposal `Builders as functions` #62.

Cookbook/Technical-Documents/BackendFeatureHowTo.md
```diff
@@ -0,0 +1,204 @@
...
+``` swift
+func access<Request, Response>(_ resource: BackendResource<Request, Response, NoAuth>, with request: Request) -> SignalProducer<Response, CoreError>
+```
+
+`CoreError` is preferred as it allows both network and parsing errors and is used throughout the network layer.
+
+You may have noticed `NoAuth` in the signature – this is only good for unauthenticated requests, like fetching data before the user logs in. After that, you'll likely want to use `AuthenticatedAccessible` which provides a slightly different method:
+
+``` swift
+func access<Request, Response>(_ resource: BackendResource<Request, Response, ConcreteAuth>, with request: Request) -> SignalProducer<Response, CoreError>
```
viorel15: Maybe it's worth introducing the two available concrete auth  types: KongAuh and ClincialAuth.
Also an important thing is how you construct your resource using the KongAuth.patientId.

karabatov: I thought this was a little out of scope (also exposing unnecessary internal details), and anyway auth methods are probably going to change going forward with the work that @jtomanik is doing.

viorel15 (commented):
I think it's worth adding some paragraphs for testing each component. 
For example testing the business controller should contain testing encoding/decoding of requests/responses  models, any other request attributes and specific errors decoding if any.
Also for view models we have specific strategy for testing them

Cookbook/Technical-Documents/BackendFeatureHowTo.md
viorel15: At minimum I suggest in removing the ConcreteAuth since there is no such type. 
Otherwise this will confuse even more and the reader would need at least another document to build a feature that works with backend.

karabatov:
@viorel15 I'm going to introduce the testing section in a different PR after this is merged, as it's not that small. I think it's easier to process and review in chunks, that's why I only included a simple example here.

Cookbook/Technical-Documents/BackendFeatureHowTo.md
karabatov: Ah, confusing the reader, now that makes sense. I'll see what I can do about it. I just didn't want to  complicate the doc too much.

viorel15:
> @viorel15 I'm going to introduce the testing section in a different PR after this is merged, as it's not that small. I think it's easier to process and review in chunks, that's why I only included a simple example here.

Ok. it makes sense since this is not trivial.

jtomanik (approved):

karabatov:
@viorel15 ✅ used `KongAuth` and mentioned `ClinicalAuth`.
@dmcrodrigues @TheAdamBorek ✅ added full signature for `Feedback`.
@RuiAAPeres ✅ briefly described a builder and linked to the proposal.

karabatov:
Looks like most everyone is fine with it, so I'll merge tomorrow conditional on no additional feedback, and will start working on making a second pass:
- Testing
- Various JSON parsing approaches
- Resource access with requests
- ✨

PR #156, opened 2019-06-10
Add DebugBuilder.md

inamiy:
This PR is a documentation for Debug Builder https://github.com/Babylonpartners/babylon-ios/pull/7645 .
(Confirmed that this doesn't need a proposal)

The main purpose of this feature is to quickly open the deeply-nested screen for faster UI debugging iteration.

#### How to make a debug-builder

1. Add a new case in `enum DebugBuilderPattern`
2. Create a (mock) builder in `DebugAuthenticatedRootBuilder`
3. Update `Settings.bundle/Root.plist` by adding a new case key
4. Set `Settings.bundle`'s userDefaults flag via `Settings.app`
5. Launch the app

PR #172, opened 2019-07-03
[Proposal] - Define Mocks and stubs next to the implementation

sergdort:
Not sure if it was proposed by someone before. I feel like @mr-v was talking about something similar.

### DIFF
--- a/Cookbook/Proposals/MocksAndStubsNextToImplementation.md
+++ b/Cookbook/Proposals/MocksAndStubsNextToImplementation.md
@@ -1,0 +1,53 @@
+# Put Mocks, Stubs Next to Implementation
+
+* Author(s): Sergey Shulga
+* Review Manager: TBD
+
+## Introduction
+
+Put implementationf of mocks next to the implementation of the protocol
+
+## Motivation
+
+
+- Improves Desccoverability of mocks
+- Prevents mocks duplication (Because we usually define mocks in the `FrameworkTests` target which means they can only be used there)
+- Easy to change mock implementation once public API changes
+
+
+## Proposed solution
+
+
+```swift
+public protocol BiometricAuthControllerProtocol {
+    var isRegistered: Bool { get }
+    var supportedBiometry: BiometryType { get }
+    var biometricDataHasChanged: Bool { get }
+    
+    func register() -> SignalProducer<Never, BiometricAuthError>
+    func unregister() -> SignalProducer<Never, BiometricAuthError>
+    func signedNonce() -> SignalProducer<Data, BiometricAuthError>
+}
+
+public final class BiometricAuthController: BiometricAuthControllerProtocol {
+// Implementation
+}
+
+#if DEBUG
+
+public final class MockBiometricAuthController: BiometricAuthControllerProtocol {
+// Implementation
+}
+
+#endif
+
+```
+
+## Impact on existing codebase
+
+This proposal does not break anything, allows gradual migration.
+
+## Alternatives considered
+
+Leave as it is
+
--- a/Cookbook/Proposals/MocksAndStubsNextToImplementation.md
+++ b/Cookbook/Proposals/MocksAndStubsNextToImplementation.md
@@ -5,7 +5,7 @@
 
 ## Introduction
 
-Put implementationf of mocks next to the implementation of the protocol
+Put implementation of mocks next to the implementation of the protocol
 
 ## Motivation
 
--- a/Cookbook/Proposals/MocksAndStubsNextToImplementation.md
+++ b/Cookbook/Proposals/MocksAndStubsNextToImplementation.md
@@ -10,7 +10,7 @@ Put implementationf of mocks next to the implementation of the protocol
 ## Motivation
 
 
-- Improves Desccoverability of mocks
+- Improves Discoverability of mocks
 - Prevents mocks duplication (Because we usually define mocks in the `FrameworkTests` target which means they can only be used there)
 - Easy to change mock implementation once public API changes
 


Cookbook/Proposals/MocksAndStubsNextToImplementation.md
```diff
@@ -0,0 +1,53 @@
...
+## Introduction
+
+Put implementationf of mocks next to the implementation of the protocol
+
+## Motivation
+
+
+- Improves Desccoverability of mocks
+- Prevents mocks duplication (Because we usually define mocks in the `FrameworkTests` target which means they can only be used there)
+- Easy to change mock implementation once public API changes
```
sergdort: By this I mean that we would not need to wait for Tests to fail to compile when we change interface of the protocol

Cookbook/Proposals/MocksAndStubsNextToImplementation.md
```diff
@@ -0,0 +1,53 @@
+# Put Mocks, Stubs Next to Implementation
+
+* Author(s): Sergey Shulga
+* Review Manager: TBD
+
+## Introduction
+
+Put implementationf of mocks next to the implementation of the protocol
```
NSMyself:
```suggestion
Put implementation of mocks next to the implementation of the protocol
```

Cookbook/Proposals/MocksAndStubsNextToImplementation.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Motivation
+
+
+- Improves Desccoverability of mocks
```
NSMyself:
```suggestion
- Improves Discoverability of mocks
```

NSMyself (changes requested):

Cookbook/Proposals/MocksAndStubsNextToImplementation.md
viorel15: Usually mocks and stubs are shared between multiple tests. Discoverability will be very low in that case

NSMyself (commented):
On the other hand, this would "pollute" the implementation if you weren't working on tests (more visual data to process). I'm on the fence on this one

Cookbook/Proposals/MocksAndStubsNextToImplementation.md
sergdort: > Usually mocks and stubs are shared between multiple tests

Not sure what do you mean here?

My point here is that in order to fine a mock we would just need to go to the protocol definition

sergdort:
@NSMyself 

> On the other hand, this would "pollute" the implementation if you weren't working on tests

For example if we are not adding any tests in a PR that introduces some business controllers we could just put `fatalErrors()` in the methods

ilyapuchka:
I think we already have an issue with our current approach where we don't put tests in the Tests subfolder (my preferred way) next to implementation but just in the same folder instead. That really pollutes the project structure IMO. So such details should be well defined in the proposal.

sergdort:
> I think we already have an issue with our current approach where we don't put tests in the Tests subfolder (my preferred way) next to implementation but just in the same folder instead. That really pollutes the project structure IMO.

Yeah we can agree for example that we alway put mocks in the end of the file ¯\_(ツ)_/¯

diegopetrucci (changes requested):
Very similar to this https://github.com/Babylonpartners/ios-playbook/pull/106/files?short_path=cbf3c01#diff-cbf3c0179a8a83d1db71d563da122d3d

Maybe we need an update from @mr-v?

ilyapuchka:
@diegopetrucci the proposal from @mr-v didn't touch on where the files should be. do you suggest we amend it rather than use separate proposal for that?

Cookbook/Proposals/MocksAndStubsNextToImplementation.md
```diff
@@ -0,0 +1,53 @@
...
+    func register() -> SignalProducer<Never, BiometricAuthError>
+    func unregister() -> SignalProducer<Never, BiometricAuthError>
+    func signedNonce() -> SignalProducer<Data, BiometricAuthError>
+}
+
+public final class BiometricAuthController: BiometricAuthControllerProtocol {
+// Implementation
+}
+
+#if DEBUG
```
zzcgumn: This should be a separate pre-compiler directive if we accept the proposal. There exist scenarios when we want to run unit tests with optimisations enable. My suggestion would be `#if DEBUG` -> `#if UNITTEST`.

zzcgumn:
I am with @ilyapuchka that we should be try to make an effort to move unit test code to a `Test` folder next to the feature before considering this proposal. If the issue with breaking mock too easily persists we can reconsider this proposal.

Cookbook/Proposals/MocksAndStubsNextToImplementation.md
ilyapuchka: I'm not sold on the idea to have different compilation condition. It will force us to create a new build configuration for each and every target unless there is a better way.
Regarding having mock types in the same file I think it will be better to keep them separate from production code and close to test cases, so in the `Tests` subfolder

ilyapuchka:
After reading the proposal more carefully I think I tend to agree with its motivation but not the proposed solution.

> Improves Discoverability of mocks

I'm not sure how this will solve the issue. IMO the problem lays partially in the fact that we don't have consistent naming for test doubles, so when one tries to find it as, let's say, `StubAppointmentBusinessController` they don't find it, though there is `MockAppointmentBusinessController` already.

> Prevents mocks duplication
Having mocks in the tested target, i.e. in `AppointmentsUI` framework, will allow to reuse them in different test targets just by linking to `AppointmentsUI`, but it seems to me that a cleaner solution would be for each feature framework to provide a test utilities as a framework that other tests can use. The disadvantage of this approach is that we will have more inter dependencies to manage. We can maybe explore if we can put all the mocks in the shared test utilities framework that we already have, but this will make us link all the frameworks to it which that might be not optimal as well.

> Easy to change mock implementation once public API changes

This can be achieved not only by including mocks in the targets, but by enabling building tests together with the target. For example if we enabled building Babylon-UITests with Babylon target and introduce an error there then it will be reported when building Babylon (not that the column is called Run, but in fact the tests are built also when building without running the app)

Screenshot 2019-07-08 at 11 04 18
Screenshot 2019-07-08 at 11 05 07

Cookbook/Proposals/MocksAndStubsNextToImplementation.md
sergdort: > Tests subfolder
That defeats the purpose of discoverability. You should be able to go the the Interface and straight a way see the implementation and mocks

ilyapuchka: I agree with that, but still not sure about having mocks code right there is. How do we ensure that we don't forget to wrap this code with `#if ... #end`? Or we don't care a lot if we accidentally ship it?

sergdort:
> I'm not sure how this will solve the issue.

It's solve the issue because in oder to find a mock you just go to the protocol definition.

> but it seems to me that a cleaner solution would be for each feature framework to provide a test utilities as a framework

This increase build time as well, as test utilities would need to embed all other frameworks. So if I want to run tests only for a framework I would need to build all.

> This can be achieved not only by including mocks in the targets, but by enabling building tests together with the target.

I do not think it's a good idea as it will increase the build time.

The main point of this proposal is that it has almost 0 effort, we just need to move mocks next to the original implementation and we would not need to change our infrastructure.

Cookbook/Proposals/MocksAndStubsNextToImplementation.md
anilputtabuddhi: If keeping mocks close to the source files is required, they could also be in a separate files but right next to the source file with a ```XXX+Mock.swift```. I suggest suffixes so that alphabetically they are as close to the source file as possible. But the mock file itself IMHO should only be in the test target else we could end up in a lot of trouble by accidentally shipping mock files in production code.

daniloaliberti: I like `Mock`s the `Tests` subfolder and it is an improvement considering it is in the same target. 
In the same file is very risky. Another file prefixed with `Mock` is more explicit.

emeseuk (commented):
I haven't worked much with the mocks and stubs yet, but this would make it easy to find them. I think adding them at the end of the files is a nice idea. Up until we don't have a consistent naming wouldn't this solve the issue? You just check the implementation and figure out the name.

Cookbook/Proposals/MocksAndStubsNextToImplementation.md
daniloaliberti: @sergdort just imagine `MockPatientResponse`

daniloaliberti (commented):
I agree with adding them in the `Test` subfolder - but also depending how cross-target the entity is. 
In the same file is to trust too much in the `#if` ... and I can see thoughts about adding another Danger rule to cover this, which I don't like either.

ilyapuchka:
@sergdort including mocks in the source code will increase the build time as well, and the size of debug binary. How will this solution ensure that all the mocks are wrapped within `#if DEBUG` and are not shipped with release build? or that's not the issue?

inamiy (approved):
I'm 👍 on using `#if TEST` macro, as similar approach has been already done in Rust
e.g. `#[cfg(test)]` for simple unit-test https://doc.rust-lang.org/book/ch11-03-test-organization.html


This can be said for organizing fixtures (cf. #106 ) as well, and with that, integrating #156 Debug Builder can be achieved more easily.
(I was first thinking about gathering fixture data in some new framework, but now I find `#if TEST` is easier)

And I think having `Tests` subfolder is not mandatory.

mr-v (commented):
At first I was sceptical about the proposal, but the more I think about it the more it makes sense. Right now we're missing a clearly defined place for Stubs and Mocks and it leads to all sorts of problems. This simple idea has great benefits, e.g. once you need a protocol stub from some framework in tests, then all you need to do is add `@testable import SomeFramework` and it's ready.  No messing around with test helper frameworks.

I'm not sure about using `#DEBUG` flag though. It might make compile times even longer when you don't need the mocks (and potentially fixtures). On the other hand it might prevent us from creating dependency of test utilities on feature frameworks, which is much worse for compile times.

I'm not sure if adding a new compile flag would be worth the trouble, unless we somehow automate project changes.

> The disadvantage of this approach is that we will have more inter dependencies to manage. We can maybe explore if we can put all the mocks in the shared test utilities framework that we already have, but this will make us link all the frameworks to it which that might be not optimal as well.

> (I was first thinking about gathering fixture data in some new framework, but now I find #if TEST is easier)

I'm in the middle of untangling dependency on feature frameworks in snapshot test utilities framework. It's really easy to accidentally introduce change that will slow down compile times due to some transient dependencies, e.g. UI components tests use snapshot test utilities, which in turn import feature UI framework, causing the wait for UI frameworks compilation before UI component tests could be run.

> This can be said for organizing fixtures (cf. #106 ) as well, and with that, integrating #156 Debug Builder can be achieved more easily.

That would be pretty neat. I had some `FakeAPI` implementations (partially) for this purpose and had to recreate placeholder models code there as well.

@inamiy btw thank you for adding this Debug Builder, it's brilliant and saved me a lot of time when I was working on some interaction details. 👍

TheAdamBorek:
At first it felt weird however as Witold said: it solves few problems we have like not being able to import a `import ATest` into `BTest`.

I would stay with `#if DEBUG` flag to not have to maintain another build configurations.

To make it safer we may think about Danger rule to check:
`if type.contains("Stub | Mock")  then file.contains("#if DEBUG)` or as @ilyapuchka suggested on a meeting: use the exclude option during build.


It's ✅ from me for the idea. We need to decide on details implementations.

TheAdamBorek (approved):

ilyapuchka:
Unfortunately `if type.contains("Stub | Mock") then file.contains("#if DEBUG)` won't be enough and the check should be much more complex @TheAdamBorek

ilyapuchka:
I'm +1 for including test mocks and fixtures in the framework targets if this improves reusability of these mocks, but I'd like this to be amended with well defined naming conventions for mocks and fixture files and types, agree that we put them in a separate folder in the `Tests` folder next to the feature (or another place). To avoid including by accident this code in production binary I'd suggest to use EXCLUDE_FILES build configuration. This will ensure that we don't ship it to production (if files are named correctly) and if at some point we reach the point when mocks can be generated then it's easier to do that with separate files.

horothesun (approved):
I agree with moving the mock definition into a separate file 👍

karabatov (approved):
Approving, and comments raised at the proposal meeting have already been expressed by others.

RuiAAPeres:
Closing due to inactivity. Please re-open when you have time to work on it. ❤️

sergdort:
Wait it was approved!

Ben-Henshall:
@sergdort / @RuiAAPeres wasn't this approved? Can we bring this back if so?

sergdort:
The agreement was to put it in the folder with the tests and exclude from linking in release mode the same we do for debug actions, right @ilyapuchka ?

Ben-Henshall:
@ilyapuchka https://github.com/Babylonpartners/ios-playbook/pull/172#issuecomment-527851737

ilyapuchka:
yes 👍

Ben-Henshall:
@sergdort https://github.com/Babylonpartners/ios-playbook/pull/172#issuecomment-530311494
we have a consensus so I guess we can put the outcome in the proposal and merge it 👍

inamiy:
https://github.com/babylonhealth/ios-playbook/pull/172#pullrequestreview-260620080
I think I misinterpretted this proposal a bit, and I'm negative to using `#if DEBUG` rather than `#if TEST` which @zzcgumn already pointed out the issue of release-mode-testing in https://github.com/babylonhealth/ios-playbook/pull/172#discussion_r300998537 .

sergdort:
@inamiy @zzcgumn At the moment we do not have any tests that run in `RELEASE` mode, what would be the value of having them? 🤔

inamiy:
@sergdort 
cf. https://babylonhealth.slack.com/archives/G22EWKXQT/p1572952030452100?thread_ts=1572948356.438000&cid=G22EWKXQT

> BTW reusing `#if DEBUG` for alternative of `#if TEST` will soon become hard to manage macros as we are probably aware in ObjC age working in horrible project (if you’ve ever done).
>
> All I can just say is Test ≠ DEBUG

As mentioned there, my comment above is just a removal of my previous approval, so it won’t affect our consensus.
So, you don't need to reopen this PR that already has consensus.

sdaniel55:
@sergdort can we merge this if it was approved? / @diegopetrucci @NSMyself

NSMyself (approved):

NSMyself:
Thanks for the heads-up @sdaniel55, totally forgot about this

Cookbook/Proposals/MocksAndStubsNextToImplementation.md
mluisbrown: So @ilyapuchka your main concern is Mocks accidentally getting shipped? And by having them in a separate file, they can be added to a separate target? The likelihood of forgetting to add `#if DEBUG` is pretty small IMO. And you could just as easily add the mock file to the main target instead of the test target by mistake and have the same problem.

ilyapuchka: 1. The problem is not just that but also the fact that we can break release build having debug build perfectly fine, this happened not so long time ago and it's hard to spot until we build for release which we don't do constantly.
With suggested approach of using excluded files build setting to filter out fixture files (which IIRC we agreed on on the proposal meeting) we only need to maintain file name convention (and we can use regexp I think) to ensure that fixtures are being excluded. I saw we already follow this convention of putting fixtures in `+fixture.swift` files, but probably not everywhere.

2. Having tests related files and actual production code all mixed together in single folder and in single group in Xcode also is not great IMO as it's harder to navigate project structure this way. And we agreed to put all tests into Tests subfolder, so fixtures can go there as well. One can argue that fixtures in this Tests folder will belong to framework target and tests will actually belong to tests target, which may be confusing and people can consider that an error. Then we can have a different sub folder just for Fixtures. 

3. It also makes it potentially easier to analyse what changed in the automation if we keep source code separate from tests. And if ever we will decide to move to SPM to manage internal dependencies it will be easier to migrate (but it will invalidate the whole approach of having Tests sub folder in the sources folder).

diegopetrucci (approved):
What happened to this?

Cookbook/Proposals/MocksAndStubsNextToImplementation.md
chitrakotwani: How does it become easier to change the mock implementation?

Cookbook/Proposals/MocksAndStubsNextToImplementation.md
```diff
@@ -0,0 +1,53 @@
...
+* Review Manager: TBD
+
+## Introduction
+
+Put implementation of mocks next to the implementation of the protocol
+
+## Motivation
+
+
+- Improves Discoverability of mocks
```
chitrakotwani: Can we not follow the pattern "{ModelName}+Fixture.swift" and put it in Test utilities.

chitrakotwani: This way we can easily find it by filtering out this name in project navigator

Cookbook/Proposals/MocksAndStubsNextToImplementation.md
sergdort: I mean that when we build main app the mocks won't compile as well, so compiler would force us to modify mocks as well, instead of doing it later when working on tests

chitrakotwani: I understood that according to proposal and comments on it, these files will be excluded when building the app. So, I am not sure if this will make any difference.

Cookbook/Proposals/MocksAndStubsNextToImplementation.md
chitrakotwani: I read some comments on conversation tab, and I see why we don't want to put all of them in test utilities framework

PR #177, opened 2019-07-09
[Proposal] Detect usage of Current outside builders

NSMyself:


Cookbook/Proposals/CurrentOutsideBuilders.md
```diff
@@ -0,0 +1,70 @@
...
+Here is a possible implementation:
+
+```
+custom_rules:
+  world_current:
+    excluded: # Files in which it's OK to use Current
+     - ".*Builder(\+.*)?\\.swift"
+     - "World\\.swift"
+     - "AppDelegate\.swift"
+     - ".*AppConfiguration\.swift"
```
AliSoftware: Those last two probably also needs the double backslash escape too (escaping it for YAML so that once SwiftLint have read the config there's one backslash left for the regex engine to interpret as escaping the dot)

Cookbook/Proposals/CurrentOutsideBuilders.md
```diff
@@ -0,0 +1,70 @@
...
+     - "AppConfiguration.swift" # For depreciation warnings
```
AliSoftware: Forgot a double-backslash here

Cookbook/Proposals/CurrentOutsideBuilders.md
```diff
@@ -0,0 +1,70 @@
...
+    name: "Current:World usage"
+    regex: "\wCurrent\."
```
AliSoftware: Needs double-escapes there too

Cookbook/Proposals/CurrentOutsideBuilders.md
```diff
@@ -0,0 +1,70 @@
...
+
+`Current` is a very powerful concept whose usage is only allowed in a very restrict number of places. More detailed information about Current can be found [here in our Current Guide](../Technical-Documents/Current-guide.md). As such, we should enable automatic monitoring in order to prevent it from being abused throughout our codebase. Nevertheless, Current is still being used in a lot of places (598 results across 144 files) and we should go through every case individually before commiting to a decision:
+
+`Current` should only be allowed in:
+- `AppDelegate.swift`
+- `DesignLibrary.swift`
+- `World.swift`
+- Builders (using the `*Builder(\+.*).swift` regex)
+- `*AppConfiguration.swift`
+- AppDependencies
```
AliSoftware: And in tests?

diegopetrucci:
Does this need a proposal? I don’t think it does. This stuff has already being agreed with and this would just be implementing it. And, worst case scenario, if it creates issues we just disable the rule.

NSMyself:
Thanks for the review @diegopetrucci but this wasn't ready just yet 😅 In any case, we think we needed a proposal mostly because of all those edge cases of files that are using Current but shouldn't, imho

RuiAAPeres:
Closing due to inactivity. Please re-open when you have time to work on it. ❤️

NSMyself:
Totally forgot about this. Thanks @RuiAAPeres, I'll reopen at a later stage

dmcrodrigues:
Do we want to re-open this one @NSMyself @AliSoftware?

PR #180, opened 2019-07-12
Describe using `AccessibilityContent` for elements and screens

karabatov:
As we can now use `AccessibilityContent` in automation, we should try and do that more, and rely less on copy, same for locating if a screen is displayed.

Also fixes some typos and header levels.

---

I'm interested in feedback re: `ScreenNaming`, as it's now used in home screen and notifications.

### DIFF
--- a/Cookbook/Technical-Documents/AutomationIdentifiers.md
+++ b/Cookbook/Technical-Documents/AutomationIdentifiers.md
@@ -85,12 +85,14 @@ After this identifier has been set on a control, we can `import BabylonDependenc
 
 ``` swift
 let promofield = app.tables
-    .cells[AccessibilityContent.PromoCode.codeCell]
+    .matching(.cell, identifier: AccessibilityContent.PromoCode.codeCell)
     .textFields
     .firstMatch
 ```
 
-Note that the identifier is usually set on the cell itself rather than its internal components, as automation has problems finding elements otherwise.
+Note that you may need to set the identifier on the cell itself rather than its internal components, as automation may have problems finding elements otherwise.
+
+It is also preferable to always use `.matching(.cell, identifier: …).firstMatch` rathen than a subscript (`[…]`) as the subscript variant only expects a single match. If, for some reason, there are several, it will throw an error.
 
 ### Taking advantage of ScreenNaming for screen identifiers
 
--- a/Cookbook/Technical-Documents/AutomationIdentifiers.md
+++ b/Cookbook/Technical-Documents/AutomationIdentifiers.md
@@ -94,17 +94,7 @@ Note that you may need to set the identifier on the cell itself rather than its 
 
 It is also preferable to always use `.matching(.cell, identifier: …).firstMatch` rathen than a subscript (`[…]`) as the subscript variant only expects a single match. If, for some reason, there are several, it will throw an error.
 
-### Taking advantage of ScreenNaming for screen identifiers
-
-The fact that view models conform to `ScreenNaming` for analytics tracking can be used in automation, because e.g. `BabylonBoxViewController` automatically sets the screen's accessibility identifier to the view model's `screenName`:
-
-``` swift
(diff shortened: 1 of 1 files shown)


Cookbook/Technical-Documents/AutomationIdentifiers.md
```diff
@@ -67,23 +67,90 @@ fileprivate enum NavigationBar {
...
+After this identifier has been set on a control, we can `import BabylonDependencies` in a Screen implementation in automation, and update the query to use this identifier:
+
+``` swift
+let promofield = app.tables
+    .cells[AccessibilityContent.PromoCode.codeCell]
+    .textFields
+    .firstMatch
+```
+
+Note that the identifier is usually set on the cell itself rather than its internal components, as automation has problems finding elements otherwise.
```
ilyapuchka: I don't think that happens always. It also probably makes sense to mention how the identifier then should be set on the cell in code.

```suggestion
Note that you may need to set the identifier on the cell itself rather than its internal components, as automation may have problems finding elements otherwise.
```

Cookbook/Technical-Documents/AutomationIdentifiers.md
```diff
@@ -67,23 +67,90 @@ fileprivate enum NavigationBar {
...
+
+### Taking advantage of ScreenNaming for screen identifiers
```
ilyapuchka: I find this section very confusing and I'd prefer if it would only mention that Bento is doing it this way and how then it should be reflected in the automation. And a note that ScreenNaming should not be used for setting accessibility identifiers manually, instead it should use AccessibilityContent.

P.S. IMO Bento should not assign accessibility identifiers automatically especially using something from completely different domain (analytics) and instead offer a way to set accessibility identifier for the screen manually. But that's not a discussion about Bento.

Cookbook/Technical-Documents/AutomationIdentifiers.md
```diff
@@ -67,23 +67,90 @@ fileprivate enum NavigationBar {
...
+        case codeCell
+    }
+}
+```
+
+After this identifier has been set on a control, we can `import BabylonDependencies` in a Screen implementation in automation, and update the query to use this identifier:
+
+``` swift
+let promofield = app.tables
+    .cells[AccessibilityContent.PromoCode.codeCell]
```
JasonDobo:
```suggestion
    .matching(.cell, identifier: AccessibilityContent.PromoCode.codeCell)
```

JasonDobo: I generally don't like using ["string"] in a query as it will throw an acceptation if multiple. matches are found. The code I suggested will work regards of a single or multiple matches for a Cell with the identifier

karabatov:
I'm going to address the comments soon 🙏

Cookbook/Technical-Documents/AutomationIdentifiers.md
karabatov: Yes, probably makes sense to downplay it a little bit.

karabatov:
Following discussion on Slack, add a note about preferring `matching(identifier: …).firstMatch` to using subscripts for identifiers.

Cookbook/Technical-Documents/AutomationIdentifiers.md
karabatov: Following discussion on Slack, mention avoiding using `ScreenNaming` raw identifiers without setting them manually as accessibility identifiers, as automatic setting of them will likely be removed.

karabatov:
@ilyapuchka @JasonDobo @inamiy I have made the requested changes:
- [x] Changed `[…]` to `.matching(…)` for elements (cells).
- [x] Made a note about using `matching(…).firstMatch` rather than `[…]`.
- [x] Replaced the section on `ScreenNaming` with using `Screen` accessibility identifiers in automation instead (after a Slack debate).

ilyapuchka (approved):

JasonDobo (approved):

PR #182, opened 2019-07-12
Update CODE_REVIEW.md

RuiAAPeres:
Added change based on the following:

> Given recent events, I think it would be good for us to agree on the following when it comes to PR etiquette. Because we are a remote team, with different background and cultures, sometimes our intentions and words are not clear to each other and, as such, we remain in this grey area where we don’t know what the other person means. I want to suggest the following:
>
> -> Any comment left in a PR is “just” a suggestion, except if the Author of said comment blocks the PR.
>
> The goal here is to differentiate between “This is wrong because `X` and I would advise doing `Y`” *versus* “Maybe you should do `Y` instead”. The former brings objective feedback, that can be actionable by the PR author, while the latter is an opinion. Said opinion, can be considered by the Author, but nothing is forcing him/her to follow it.
```

### DIFF
--- a/Etiquette/CODE_REVIEW.md
+++ b/Etiquette/CODE_REVIEW.md
@@ -59,7 +59,7 @@ Reviewing a Pull Request
 * Do not try to make everything perfect (it never will be and otherwise we'll never deliver anything), but do not compromise on quality
 * Do not comment if you don't have any suggestions to address your comment
 * Use GitHub suggestions as much as possible
-* If you feel strong against the change - please use `Request changes` and explain why you are doing that. **Any comment left in the PR is taken as a suggestion, and it's up to the author's discretion to accept them or not, except if you use `Request changes`.**
+* If you feel strongly against the change - please use `Request changes` and explain why you are doing that. **Any comment left in the PR is taken as a suggestion, and it's up to the author's discretion to accept them or not, except if you use `Request changes`.**
 * Approve changes if you are confident about them. 
 * If you are the first reviewer to approve the changes add the label "Needs one reviewer" so that others can see which PRs already have a feedback and which don't have any yet.
 * Comment on the changes if something is not clear to you and needs further clarification or actions
--- a/Etiquette/CODE_REVIEW.md
+++ b/Etiquette/CODE_REVIEW.md
@@ -59,7 +59,7 @@ Reviewing a Pull Request
 * Do not try to make everything perfect (it never will be and otherwise we'll never deliver anything), but do not compromise on quality
 * Do not comment if you don't have any suggestions to address your comment
 * Use GitHub suggestions as much as possible
-* If you feel strong against the change - request changes and explain why you are doing that. Do not request changes for typos, missing documentation comments or code style violations, trust your colleagues to address them.
+* If you feel strong against the change - please use `Request changes` and explain why you are doing that. **Any comment left in the PR is taken as a suggestion, and it's up to the author's discretion to accept them or not, except if you use `Request changes`.**
 * Approve changes if you are confident about them. 
 * If you are the first reviewer to approve the changes add the label "Needs one reviewer" so that others can see which PRs already have a feedback and which don't have any yet.
 * Comment on the changes if something is not clear to you and needs further clarification or actions
--- a/Etiquette/CODE_REVIEW.md
+++ b/Etiquette/CODE_REVIEW.md
@@ -59,7 +59,7 @@ Reviewing a Pull Request
 * Do not try to make everything perfect (it never will be and otherwise we'll never deliver anything), but do not compromise on quality
 * Do not comment if you don't have any suggestions to address your comment
 * Use GitHub suggestions as much as possible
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by RuiAAPeres** [comment]
@ilyapuchka can you approve this. 😉

**Review by MichalTKwiecien — APPROVED** [review]

**Review by zopa-simon-cass — APPROVED** [review]

File: Etiquette/CODE_REVIEW.md [hunk 1]
```diff
@@ -59,7 +59,7 @@ Reviewing a Pull Request
 * Do not try to make everything perfect (it never will be and otherwise we'll never deliver anything), but do not compromise on quality
 * Do not comment if you don't have any suggestions to address your comment
 * Use GitHub suggestions as much as possible
-* If you feel strong against the change - request changes and explain why you are doing that. Do not request changes for typos, missing documentation comments or code style violations, trust your colleagues to address them.
+* If you feel strong against the change - please use `Request changes` and explain why you are doing that. **Any comment left in the PR is taken as a suggestion, and it's up to the author's discretion to accept them or not, except if you use `Request changes`.**
```
mluisbrown:
```suggestion
* If you feel strongly against the change - please use `Request changes` and explain why you are doing that. **Any comment left in the PR is taken as a suggestion, and it's up to the author's discretion to accept them or not, except if you use `Request changes`.**
```

mluisbrown (approved):

inamiy (approved):

Etiquette/CODE_REVIEW.md
ilyapuchka: If this implies that such comments can be simply ignored without any response on them then I don't think that this is a good etiquette.

If the comment is something reviewer does not have a strong opinion on and does not care about what author of the change thinks about such suggestion ("take it or leave it") then they should indeed leave it up to the author of the change to accept it, but then they should just approve the change (assuming there is nothing else). 

If reviewer does not have a strong opinion but at the same time wants to discuss the matter before deciding to approve or request changes than just leaving a comment is reasonable imo. It starts conversation. It's fine if reviewer is not ready to approve or request changes and needs clarification or some discussion before that.

Maybe what we are missing is this conclusion part. So such discussion should maybe result in the comment being resolved (with "resolve" button) and approval if there is nothing left or with request changes if the result of conversation is that changes should be made. (though I thought that's clear for everyone)

We also have a point about such discussions where it seems to hard to reach conclusion quickly to better happen not on GitHub but in person or in a call.

ilyapuchka (changes requested):

Etiquette/CODE_REVIEW.md
ilyapuchka: When reviewer just goes and requests changes for something that author of the change considers as a matter of taste this will as well cause a conflict. That's why I think people in this case tend to leave a comment and neither approve, nor request changes

diegopetrucci: I agree with what @ilyapuchka is suggesting. Not every comment has to be addressed, but ideally every comment has to be (at least!) acknowledged.

RuiAAPeres: This doesn't imply what you mentioned @ilya. Please @diego add an addendum to this with your final sentence, since it's a clean and to the point rule.

RuiAAPeres: Added your suggestion Diego. Thank you.

Etiquette/CODE_REVIEW.md
```diff
@@ -59,7 +59,7 @@ Reviewing a Pull Request
 * Do not try to make everything perfect (it never will be and otherwise we'll never deliver anything), but do not compromise on quality
 * Do not comment if you don't have any suggestions to address your comment
 * Use GitHub suggestions as much as possible
-* If you feel strong against the change - request changes and explain why you are doing that. Do not request changes for typos, missing documentation comments or code style violations, trust your colleagues to address them.
+* If you feel strongly against the change - please use `Request changes` and explain why you are doing that. **Any comment left in the PR is taken as a suggestion, and it's up to the author's discretion to accept them or not, except if you use `Request changes`.** Not every comment has to be addressed, but every comment has to be acknowledged.
```
ilyapuchka: I'm not going to approve this in this formulation as it means that any comment that is not followed by request for changes can be ignored (equals acknowledged). It also must not be in bold as it is definitely not taking precedence of anything else in the doc. And the rest of the doc pretty clearly touches on how comments should be done and how they should be addressed. This change on the other hand makes etiquette worse as it justifies ignoring any comment author disagrees with and will only lead to more requests for changes and more frustration in the team instead of normal and healthy discussions (like one that was used to illustrate the issue this suggestion tries to address in Slack).

ilyapuchka (changes requested):

Etiquette/CODE_REVIEW.md
RuiAAPeres: I am closing this PR. You clearly have strong opinions about what you want with this piece. 👍 You can take the initiative an open a new PR with your ideas.

PR #187, opened 2019-07-16
Adopt SwiftGen for localizable strings

dmcrodrigues:


Cookbook/Proposals/SwiftGen.md
```diff
@@ -0,0 +1,83 @@
...
+
+As an example, `chatbot_send` should become `chatbot.send` to ensure we get a namespace for chatbot. Ideally we would like to keep the same namespaces to avoid a huge refactor in the codebase.
+
+```swift
+enum L10n {
+  enum Chatbot {
+    static let send = L10n.tr("chatbot.send")
+  }
+```
+
```
daniloaliberti:
```suggestion
To illustrate the flat structure, the mentioned key would be:
`let chatbotSend`
And we want to maintain the same structure aiming the less impact possible.
```

andersio:
A possible alternative is using dynamic member lookup, which doesn’t provide auto completion but also does not dump a huge pile of symbols into our binaries.

Should we have an estimation/spike on the impact of these auto generated properties in terms of binary size?

dmcrodrigues:
> A possible alternative is using dynamic member lookup, which doesn’t provide auto completion & guarantee of existence at compile time (to some extent)

I can include this as an alternative but it doesn't improve any of the problems identified.

> but also does not dump a huge pile of symbols into our binaries.

Even if we consider that 20% of the keys are outdated or no longer used we already have this huge pile of symbols in our binaries so the increment shouldn't be that much with the main benefit of having compile time checking and avoid any significant maintenance.

Cookbook/Proposals/SwiftGen.md
dmcrodrigues: I'll include a mention, thanks @daniloaliberti!

adiki:
Will it be included in some target like `BabylonLocalization`? How much time does it need to generate code for our strings base? Will it be smart enough to based on e.g. checksum doesn't run it every time?

daniloaliberti:
@adiki It's in the `BabylonDependencies`.

dmcrodrigues:
> How much time does it need to generate code for our strings base? Will it be smart enough to based on e.g. checksum doesn't run it every time?

At the moment we only pull strings manually so I think we should keep the same procedure unless we feel strong about getting it running automatically.

daniloaliberti:
@dmcrodrigues It's still not clear here what's the trigger of it. 
`Run script` in `Build Phases`? (that's my understanding when I used `R.swift` a long time ago). 
If so, I see the problem with our current approach that is "commit just your changes". 

for example: 
If I build and it includes all the changes (mine and all others) in the swift namespace, the work of removing others strings and swift is doubled. 

My suggestion: triggered manually **after removing others strings changes from git**. 
This way it creates the swift part of just my changes.

Or we need to rethink the strategy.

Cookbook/Proposals/SwiftGen.md
```diff
@@ -0,0 +1,83 @@
...
+### Implementation
+
+Adopting SwiftGen won't be a trivial task and would require a certain level of investment from us.
+
+##### Namespacing
+
+Based on how we have our keys defined we won't be able to mangle them into namespaces directly, we will have a huge flat list with all the strings. This can be changed but requires a major refactor of the keys to define the respective vertical/domain/module.
+
+As an example, `chatbot_send` should become `chatbot.send` to ensure we get a namespace for chatbot. Ideally we would like to keep the same namespaces to avoid a huge refactor in the codebase.
```
ilyapuchka: I was thinking about the same issue, but can't we instead adjusts the templates to treat `_`, let's say just first one to avoid "over namespacing" as `.`?

Cookbook/Proposals/SwiftGen.md
```diff
@@ -0,0 +1,83 @@
...
+
+```swift
+enum L10n {
+  enum Chatbot {
+    static let send = L10n.tr("chatbot.send")
+  }
+```
+
+Because this involves a major rename of keys this needs to be properly analyzed and coordinated to avoid any side-effects.
```
ilyapuchka: Rather than applying it to all the keys can we instead do that only for the keys that are following new pattern (if we can't use existing `_` as per my previous comment)? It will require adjusting template but it seems to be a simple filter.

Cookbook/Proposals/SwiftGen.md
```diff
@@ -0,0 +1,83 @@
...
+##### 2. Safety
+
+While this types are useful for development by enabling autocomplete they don't provide any safety since we can add a new property for a new key without necessarily adding the new string.
+
+### Proposed Solution
+
+SwiftGen is a code generator which is able to generate all this types and properties automatically by running a script which would fix the two main problems that we have with the current approach.
+
+- Maintenance will be zero because everything will be mapped based on the localisable strings;
+- Compile-time type safety will be guaranteed because everything will be generated from the localizable strings making it impossible to use something without being in the strings file or keep using something that was deleted.
```
ilyapuchka: This comes with a tradeoff that if we need specific string with a different key for one particular target we will need to put it into common strings file. I guess using SwiftGen will affect how we approach this and target specific strings as well, because i.e. code generated for target specific strings may collide with common generated code.

ilyapuchka (approved):
I'm definitely +100 for this, but I have some ideas about how to make transition smoother. I think we can achieve that and migrate to SwiftGent strings gradually. We will have +1 localisation access point for a while but in the end we will be able to get rid of all others.

adiki (approved):

Cookbook/Proposals/SwiftGen.md
AliSoftware: The split against `.` is done inside the SwiftGen engine (and then the templates can use it to loop over each groups to generate the code).

Maybe we can make the template do the split only for first level, but the split won't be easy to code via a template in a recursive way. There's some pending work to allow customization of the engine itself (similar to `params:` in the config file, but for arguments passed to the parser engine, not to the template) in order, typically for Strings, to provide a different spilt character (`_` instead of `.`) that would solve that. That require that I find some time to triage some pending PRs and make the overdue release and implement this feature first though (I really need to find some time to work on SwiftGen again, to review pending PRs and releases...)

Cookbook/Proposals/SwiftGen.md
```diff
@@ -0,0 +1,83 @@
...
+
+We current have two sets of strings, the main localization and one target specific to allow further customisation if/when needed. This requires custom handling where we query the target specific table first and only then we fallback to the main table. SwiftGen has support to search in other strings files but considering our specific flow we may need to write our own custom template to ensure we keep the same flow.
+
+##### Tests
+
+We have a couple of tests using localisable strings from the tests bundle and they will be impacted with this change. The alternative is to use the main bundle for them which can cause more failures with snapshot testing depending how often the strings get updated after being introduced. Alternatively, we can write our own custom template to keep the existent flow.
+
+### Alternatives
+
+There are other code generators that we can use as an alternative but SwiftGen seems the most popular in the community and we have the creator in our team 😃, if we have any issue we can have a great level of support.
```
AliSoftware: ❤️

karabatov:
Are SwiftGen-generated `.swift` files supposed to be committed to the repo or not? For example, when you work with R.swift the resulting `R.swift` file is _not_ committed, so each time you build the project your generate a new one if it's missing or out of date. This allows only committing string changes and ensures generation is working correctly because everyone has their own copy.

viorel15 (approved):

zzcgumn (approved):
I do agree that our current strategy for handling strings does not scale with the number of flavours and languages that we expect to have to support. As for which code generation tool to use I have no real opinion as I have never used any of them.

ilyapuchka:
@karabatov I think it's not about SwiftGen vs R.swift but about the principle of committing generated files or not. I used R.swift a lot and always committed it IIRC.

Cookbook/Proposals/SwiftGen.md
ilyapuchka: |If the only option then is to update the keys then we can do that with an API call https://lokalise.co/api2docs/curl/#transition-update-a-key-put

daniloaliberti: 1) list all keys
https://lokalise.co/api2docs/curl/#transition-list-all-keys-get
2) bulk update
https://lokalise.co/api2docs/curl/#transition-bulk-update-put
3) run to the hills 🏃🤣

karabatov:
@ilyapuchka I used it a lot too (for everything that it supports) and we didn't commit it :) At the same time, we did commit `Pods` there.

Cookbook/Proposals/SwiftGen.md
dmcrodrigues: Definitely 👍 I think it will be hard to make the transition in a single step so that would help, I'll update the proposal with the latest feedback.

Cookbook/Proposals/SwiftGen.md
dmcrodrigues: I think this won't be a problem @ilyapuchka because we will need our own function to obtain the strings in order to keep the existent flow.

1. Check target specific table
2. If available return
3. If not check the main table and return

Cookbook/Proposals/SwiftGen.md
AliSoftware: Updating the keys will probably mean breaking Android too. And in Android, dots in keys are problematic (since `R.java` doesn't expect dots in key names)

Cookbook/Proposals/SwiftGen.md
AliSoftware: That's in fact something that I've done in past projects, with a simple change via a custom template, so it's definitively doable.

AliSoftware:
> I used R.swift a lot and always committed it IIRC.
> I used it a lot too

@ilyapuchka @karabatov traitors! 😄 

More seriously, yeah it's not about SwiftGen vs R.swift but about "do we commit generated files" in general, and whatever codegen tool you use the question still stands. It's similar to the question "should we commit `Pods/` or not" indeed.

Personally I prefer to commit them, so that the change appear in PRs too, but YMMV and that's a matter of taste mainly

Cookbook/Proposals/SwiftGen.md
dmcrodrigues: I don't think we are using the same keys between Android and iOS, at least they were always defined without any synchronisation AFAIK, @daniloaliberti can you confirm?

MichalTKwiecien (approved):

brocoo (approved):

Cookbook/Proposals/SwiftGen.md
```diff
@@ -0,0 +1,83 @@
...
+}
+```
+
+### Problems
+
+The current approach works, as we would expect, but there's a couple of caveats in it.
+
+##### 1. Maintenance
+
+This types at the moment are created and managed by us manually which means any new addition/edition/removal needs to be done by us. This may not look very significant but considering the number of strings, more than 3k, it is a lot to be maintained/extended.
```
mluisbrown:
```suggestion
These types at the moment are created and managed by us manually which means any new addition/change/removal needs to be done by us. This may not look very significant but considering the number of strings, more than 3k, it is a lot to be maintained/extended.
```

Cookbook/Proposals/SwiftGen.md
```diff
@@ -0,0 +1,83 @@
...
+
+##### 2. Safety
+
+While this types are useful for development by enabling autocomplete they don't provide any safety since we can add a new property for a new key without necessarily adding the new string.
```
mluisbrown:
```suggestion
While these types are useful for development by enabling autocomplete they don't provide any safety since we can add a new property for a new key without necessarily adding the new string.
```

Cookbook/Proposals/SwiftGen.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Proposed Solution
+
+SwiftGen is a code generator which is able to generate all this types and properties automatically by running a script which would fix the two main problems that we have with the current approach.
```
mluisbrown:
```suggestion
SwiftGen is a code generator which is able to generate all these types and properties automatically by running a script which would fix the two main problems that we have with the current approach.
```

Cookbook/Proposals/SwiftGen.md
```diff
@@ -0,0 +1,83 @@
...
+
+- Maintenance will be zero because everything will be mapped based on the localisable strings;
+- Compile-time type safety will be guaranteed because everything will be generated from the localizable strings making it impossible to use something without being in the strings file or keep using something that was deleted.
+
+SwiftGen also supports strings with dynamic input generating a function to be called with all the parameters which also ensures safety as we clearly know that a certain string requires some dynamic values.
+
+We still need to ensure this script is correctly executed every time that we add/edit/remove a string but that's a easier task to manage over the current approach.
```
mluisbrown:
```suggestion
We still need to ensure this script is correctly executed every time that we add/edit/remove a string but that's an easier task to manage compared to the current approach.
```

Cookbook/Proposals/SwiftGen.md
mluisbrown: Android and iOS have different projects in Lokalise, so I don't see that there would be a problem.

mluisbrown (approved):
Just left a few small grammar / typo comments.

Cookbook/Proposals/SwiftGen.md
AliSoftware: Oh. you mean each text in the app appears twice (and thus is potentially translated twice) in Lokialize, once for the Android screen and once for the iOS screen, even though they are likely to be the same translations?

If so, that's great for us (because it would ease the transition to SwiftGen), but not sure how translators are coping with that double-work and handling match translations in common iOS/Android screens?

AliSoftware: BTW I remember now that there are some limitations with Lokalize and how they treat placeholders as opposed to literal percent characters. Especially, iirc, in Lokalize to specify a placeholder, you'd wrap it inside square brackets to tell it's a placeholder, while literal percents would be `[%]` – see https://docs.lokalise.co/en/articles/1400511-lokalise-placeholders

So you could have:

String in Lokalize | String exported in `Localizable.strings` | Intended call site | What SwiftGen sees
---|---|---|---
`"Hello [%@]!"` | `"Hello %@"` | Using `String(format:)` | contains a placeholder, will use `String(format:)`
`"Progress: [%d][%]"` | `"Progress: %d%%"` | Using `String(format:)` | contains placeholders, will use `String(format:)`
`"Yay, 10% discount!"` | `"Yay, 10% discount!"` | Using directly, no format | Contains a `% d`, which is a valid placeholder, so generates a function taking an `Int` and considers it needs to call `String(format:)` with it

So the problem is that when SwiftGen sees any translation containing a valid `%...` token as per the printf-spec (e.g. `%2d`, `%@`, `%3 d`...) it considers the localisation will need to take parameters and would need to go thru `String(format:)`, so it will generate a `func` taking those parameters. And if you want a literal `%` in there, you're supposed to then use `%%` to "escape" the literal percent character so that it produces a literal `%` after `String(format:)`.

But the problem is that it's not how Lokalise sees things, as **Lokalize expect for you to know at call site when you're supposed to call `String(format:)` for some keys and not use it for others** (see column 2), **based on human knowledge only**. So this decision of calling `String(format:)` or not on each key can't really be automated based on that, there's no real way for SwiftGen to know, when it sees `"Yay, 10% discount!"`, that it's supposed to be a literal and not to be used with `String(format:)`...

What we usually recommend is that every time you have a literal `%` in one of your translation, you should always double it in your `Localizable.strings` as `%%`, so that it's properly considered as a literal by SwiftGen and does not risk to accidentally be considered as a placeholder depending on the character that follow. But that's not how Lokalise see and export things :/

---

So if anyone have any idea how to solve that – either what algorithm we could use to determine **with certainty** if a localisation requires `String(format:)` or not based on what Lokalise exports, or a way to provide this info as some kind of metadata, or a way to hijack / post-process `lokalise pull` to replace every string that's considered to not require `String(format:)` by Lokalise but still contains `%`, to transform them to use `%%` instead... maybe with the Lokalise API to determine for each key if it exposes that it contains `[%...]` or not before export into `.strings` files...? – in any way I'm interested in any idea about this.

For more info on this limitation, see https://github.com/SwiftGen/SwiftGen/issues/577#issuecomment-455270805

AliSoftware: Mmmh after digging Lokalise might have fixed that since last time I looked at the issue (see my last comment on the SwiftGen issue), so maybe it's a false alarm after all 😅 but we definitively need to keep that case in mind and test it.

Cookbook/Proposals/SwiftGen.md
```diff
@@ -0,0 +1,83 @@
...
+
+##### 2. Safety
+
+While these types are useful for development by enabling autocomplete they don't provide any safety since we can add a new property for a new key without necessarily adding the new string.
+
+### Proposed Solution
+
+SwiftGen is a code generator which is able to generate all these types and properties automatically by running a script which would fix the two main problems that we have with the current approach.
+
+- Maintenance will be zero because everything will be mapped based on the localisable strings;
```
mr-v: We'll have third-party tool maintenance instead (as in integration/bootstrapping, updates, etc.). Not sure if we should have integration details in the proposal though.

Cookbook/Proposals/SwiftGen.md
```diff
@@ -0,0 +1,83 @@
...
+As an example, `chatbot_send` should become `chatbot.send` to ensure we get a namespace for chatbot. Ideally we would like to keep the same namespaces to avoid a huge refactor in the codebase.
+
+```swift
+enum L10n {
+  enum Chatbot {
+    static let send = L10n.tr("chatbot.send")
+  }
+```
+
+Because this involves a major rename of keys this needs to be properly analyzed and coordinated to avoid any side-effects.
```
mr-v: > Because this involves a major rename of keys this needs to be properly analyzed and coordinated to avoid any side-effects.

I'm not sure if we should accept this proposal without having a plan for migration. Even a high-level one.

mr-v (commented):
I like the idea and it would improve our workflow a little bit. I'd like to see some high-level plan for the migration.

Cookbook/Proposals/SwiftGen.md
dmcrodrigues: I think the last changes make this more clear, let me know if you think we should elaborate a bit more.

Cookbook/Proposals/SwiftGen.md
```diff
@@ -0,0 +1,122 @@
...
+
+#### Location
+
+All the symbols that we have at the moment are defined in `BabylonDependencies` and there's no reason to change this so everything can be kept as is.
+
+#### Workflow
+
+Our current workflow consists of pulling the strings manually and then only include the ones that are part of our current work. Adopting SwiftGen does not require any changes in our workflow so it should be kept unchanged, we should pull the strings and then run the script to generate the symbols manually. This file should be committed along with the strings to the repository to ensure we test the code without any external variables, e.g. if for some reason the generation fails we will get a failure while testing the PR on CI.
+
+Considering there's two steps involved, pulling the strings and then run SwiftGen to generate the symbols, and based on the fact that apply some filtering we should discard any strings unrelated with our work before running SwiftGen otherwise we will end with symbols that will need to be discarded too.
```
mr-v: We could have the fastlane script stop for dev input, e.g.:
```
✅ Strings pulled 
⚠️ Please  discard any strings unrelated with your work and hit Enter to continue.
✅ New string constants generated!
```

Cookbook/Proposals/SwiftGen.md
```diff
@@ -0,0 +1,122 @@
...
+
+#### Namespacing
+
+Based on the convention that SwiftGen follows, `module.any_key`, and how we have our keys defined, `module_any_key`, we won't be able to mangle them into namespaces directly, we will have a huge flat list with all the strings. There's two options that we can take.
+
+##### 1. Proposed: Rename all the keys
+
+An alternative is to manually rename all the keys to ensure we set up the module/namespace as expected. This give us more control and ensures we can name everything in a meaningful way but requires more work from us.
```
mr-v: I think we have third option. We can leverage reflection to generate our new keys which would match existing type structure, so in the end SwiftGen would generate exactly what we have now.

I did some quick test and added `localizedName` overload with `type` parameter:

```swift
        public static var ok: String { return localizedName(key: "general_ok", type: self) }
```

Implementation:

```swift
private func localizedName(key: String, type: Any, bundle: Bundle = LocalizationUI.bundle, function: String = #function) -> String {
    let newKey = [String(reflecting: type), function].joined(separator: ".").snakeCased()!.lowercased()
    print("current", key, "; rename to", newKey)
    return Localization.localizedName(key: key, bundle: bundle)
}
```

Which outputs:
`current general_ok; rename to babylon_dependencies.localization_ui.common.ok` (prefix can be easily removed).

This is just a proof of concept and I'm sure we can have something more sophisticated to create those nice mappings for renaming in one go.

Cookbook/Proposals/SwiftGen.md
mr-v: > Updating the keys will probably mean breaking Android too. And in Android, dots in keys are problematic (since R.java doesn't expect dots in key names)

> Android and iOS have different projects in Lokalise, so I don't see that there would be a problem.
  
It's possible to define different keys by platform, so it won't be an issue even if we use the same project.

Cookbook/Proposals/SwiftGen.md
mr-v: To clarify the example:

| Current constant | Current key | Generated key| SwiftGen constant |
|--|-|-|-|
| `LocalizationUI.Common.ok` | `general_ok` | `common.ok` |  `L10n.Common.ok` |

Cookbook/Proposals/SwiftGen.md
dmcrodrigues: Not familiar with this in specific but sounds a great option 👍

Cookbook/Proposals/SwiftGen.md
dmcrodrigues: But this still means a rename in Lokalise or not? I think this can extend the 1st option but not sure if it's a whole new option by itself.

mr-v: > But this still means a rename in Lokalise or not? I think this can extend the 1st option but not sure if it's a whole new option by itself.

Yes, there would still be be renaming on Lokalise. I can see how this can be just an extension of original option:

> 1. Proposed: Rename all the keys

I just thought that automating the keys renaming (generating from types) instead of renaming them manually would lessen the effort significantly and made it a separate option.

> An alternative is to manually rename all the keys to ensure we set up the module/namespace as expected. This give us more control and ensures we can name everything in a meaningful way but requires more work from us.

AliSoftware:
> Should we have an estimation/spike on the impact of these auto generated properties in terms of binary size — especially if we are going to SwiftGen every localised string we have? We have been doing manual localised string accessors for a very long time, but AFAIK haven’t really evaluated the approach.

@andersio I'm not sure I'm following your question: we already have those symbols in the app, we just wrote and maintain the constants them manually, and that change would automate the maintenance. We won't have more symbols, we'd just replace the symbols we already wrote manually by automatically-generated ones

AliSoftware:
> If I build and it includes all the changes (mine and all others) in the swift namespace, the work of removing others strings and swift is doubled.
>
> My suggestion: trigger it manually after removing others strings changes from git.

@daniloaliberti Not sure how the work would be doubled, since once you remove the extra other strings from your `.strings` file then SwiftGen will automatically also remove the constants next time you run it, it will not keep unused constants

dmcrodrigues:
I've updated the proposal with the agreements from the meeting. I also included a new section to clarify the migration and adoption which was something that we discussed during the meeting, we will apply the _Boy Scout Rule_ like @sergdort suggested.

MichalTKwiecien (approved):

sergdort (approved):

PR #197, opened 2019-07-24
[CNSMR-1890] Smarter diffs in view models unit tests

ilyapuchka:

### Review outcome

- explore if we can use snapshots library textual diffing instead 
- JIRA story CNSMR-2529


Cookbook/Proposals/SmarterDiffs.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+As can be seen from the diff it does not show in what exact array item the faulure happend. This is something that we will need to adjust along with [this issue](https://github.com/krzysztofzablocki/Difference/issues/1).
```
adiki:
```suggestion
As can be seen from the diff it does not show in what exact array item the failure happened. This is something that we will need to adjust along with [this issue](https://github.com/krzysztofzablocki/Difference/issues/1). 
```

Cookbook/Proposals/SmarterDiffs.md
```diff
@@ -0,0 +1,97 @@
...
+    makeState(status: .loading),
+    makeState(
+        shouldShowNHSBanner: false, // this value causes failure
+        shouldShowHealthcheckBanner: true,
+        shouldShowTestAndKitsBanner: true,
+        shouldShowGPAtHandStatusCard: true,
+        status: .loaded
+    )
+]))
+```
```
chitrakotwani: I am not convinced that writing all this boiler plate is worth it. If my test is checking a particular property in the state after an action of steps, I would like the test to be focussed on those subset of properties. It makes the reading and understanding the purpose of the test easier.
As to the problem of crashes when states array's index is out of bounds, maybe we can use guard statement
e.g. 
```
guard states.count > 10 else { XCTFail()} 
expect(states.first?.lastName).to(beNil())
```

ilyapuchka: There are benefits and tradeoffs in each of these options but regardless which one we use they both benefit from the proposal

Cookbook/Proposals/SmarterDiffs.md
```diff
@@ -0,0 +1,97 @@
...
+
+Proposed way to integrate this feature is to copy the implementation (with proper author attribution) and adjust it to our needs, optionally we can make pull requests to the original library.
+
+## Impact on existing codebase
+
+This approach can be rolled out gradually and used for new test (Xcode templates will use a new assert) or tests can be refactored in batches by replacing existing assertions with a new assert.
+
+## Alternatives considered
+
+As an alternative we can inforce asserting on individual states insteead of all states which will make diffs less of an issue. But as mentioned before it will not eliminate it.
```
brocoo:
```suggestion
As an alternative we can enforce asserting on individual states instead of all states which will make diffs less of an issue. But as mentioned before it will not eliminate it.
```

Cookbook/Proposals/SmarterDiffs.md
```diff
@@ -0,0 +1,97 @@
...
+
+Another alternative would be to use snapshot testing that we already use for UI tests. It will make our assert just a call to a single `assertSnapshot` function. But at the same time it will mean that all the snapshots will be writtent into the files rather than explicitly stated in the test code. This will decrease the boilerplate we will have to write in tests manually but will make it harder to understand test expectations as as we will need to search through text files to find values we need.
```
brocoo:
```suggestion
Another alternative would be to use snapshot testing that we already use for UI tests. It will make our assert just a call to a single `assertSnapshot` function. But at the same time it will mean that all the snapshots will be written into the files rather than explicitly stated in the test code. This will decrease the boilerplate we will have to write in tests manually but will make it harder to understand test expectations as as we will need to search through text files to find values we need.
```

Cookbook/Proposals/SmarterDiffs.md
```diff
@@ -0,0 +1,97 @@
...
+        shouldShowNHSBanner: true,
+        shouldShowHealthcheckBanner: true,
+        shouldShowTestAndKitsBanner: true,
+        shouldShowGPAtHandStatusCard: true,
+        status: .loaded
+    )
+]))
+```
+
+This eliminates the issue of the first approach when some state changes may be ignored by the test. It comes with a cost of actually creating all the intermediate states that might not be directly related to what test is testing which adds some amount of boilerplate. Another issue is that when this single assert fails it's very hard to extract what actually cuased this failure from the error message until we go and check each states separately. First option also suffers from this issue when each state consists of a lot of properties but it becomes much worse with a single assert.
```
brocoo:
```suggestion
This eliminates the issue of the first approach when some state changes may be ignored by the test. It comes with a cost of actually creating all the intermediate states that might not be directly related to what test is testing which adds some amount of boilerplate. Another issue is that when this single assert fails it's very hard to extract what actually caused this failure from the error message until we go and check each states separately. First option also suffers from this issue when each state consists of a lot of properties but it becomes much worse with a single assert.
```

Cookbook/Proposals/SmarterDiffs.md
anilputtabuddhi: @chitrakotwani But if you only assert on the subset of properties in question, then you will miss catching the corruption to state of properties outside the subset. Its best to assert the entire state equals your expected state. You could differentiate the setup of the expectation however with only the subset of properties (to understand what you are concerned about in the specific test case).

chitrakotwani: @anilputtabuddhi - This would make the tests more brittle because an unexpected change in one state property would break quite a few tests at once. This would make it difficult to pinpoint the blame of breaking tests.

But as @ilyapuchka  suggests, this is probably out of scope of discussion of this proposal anyway.

ilyapuchka: Agree with @anilputtabuddhi. We will always compromise between coverage and clarity of tests. I believe we should prefer coverage but have a clarity in mind (test should be steal readable and understandable). That's not what the proposal is about though =)

anilputtabuddhi: @chitrakotwani The tests would not be called brittle in the scenario you explained - An unexpected state change should cause failures and thats what tests are for. A brittle test is one that fails and succeeds intermittently on different runs. It is also possible to architect tests with proper fixture designs that help in test setup to be able to pinpoint breakages as you say.

Cookbook/Proposals/SmarterDiffs.md
```diff
@@ -0,0 +1,97 @@
+# Improving view models' unit tests
+
+* Author(s): Ilya Puchka
+* Review Manager: TBD
+
+## Introduction
+
+Our current approach to testing view models is described in [this doc](Cookbook/Technical-Documents/UnitTestingViewModels.md). This approach enforces the tests to follow "arrange-act-assert" pattern. If we forget about ergonomics of closures it works pretty well. But when it comes to assert section we have few options in how it can be implemented each with their own issues.
```
AliSoftware: "this doc" -> 404

Cookbook/Proposals/SmarterDiffs.md
chitrakotwani: @anilputtabuddhi - Brittle tests, as name implies, are the ones which break easily. If a test will begin to depend on too many state's properties, they will start breaking easily. A small change somewhere will have unintended effects in tests in other areas. 
Tests which fail and succeed intermittently on different runs could be called unreliable?

Cookbook/Proposals/SmarterDiffs.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Motivation
+
+As demonstrated our current approaches suffer from few issues which result in not ideal developer experience when it comes to debugging test failures.
+
+## Proposed solution
+
+To improve the experience with debugging test failures we propose to use a smarter diffing approach as in [this library](https://github.com/krzysztofzablocki/Difference) by Krzysztof Zabłocki. The idea is that when comparing two instances we will be using `Mirror` type and `dump` function to first get stored properties and then get their textual representation and this way combine a in-memory snapshot of the value. This way we can recursively compare each property and only include those that differ in the final failure message. 
+
+The fact that `Mirror` only provides access to stored properties is not an issue as computed properties are products of stored properties (or in case of enums of the enum cases and thier associated values) and for same stored properties values should return the same result.
```
AliSoftware:
```suggestion
The fact that `Mirror` only provides access to stored properties is not an issue as computed properties are products of stored properties (or in case of enums of the enum cases and their associated values) and for same stored properties values should return the same result.
```

Cookbook/Proposals/SmarterDiffs.md
AliSoftware: Taking @brocoo 's suggestion + also removing a duplicate `as`
```suggestion
Another alternative would be to use snapshot testing that we already use for UI tests. It will make our assert just a call to a single `assertSnapshot` function. But at the same time it will mean that all the snapshots will be written into the files rather than explicitly stated in the test code. This will decrease the boilerplate we will have to write in tests manually but will make it harder to understand test expectations as we will need to search through text files to find values we need.
```

karabatov:
Ah, I see why you're not keen on using the actual library, it's not very active. We could make a public fork though and work from that if necessary, with changes posted upstream?

sergdort (approved):
tenor-33512359

zzcgumn (changes requested):
I am tentatively positive towards the suggested changes, but would like to extend them a bit.

While refactoring and bug fixing I have in the last months come across a couple of examples of when states were updated as the developer expected, but events and routes were not emitted correctly.

I feel that if we update the template, we should at the same make the requirement to verify routes more explicit. If possible, we should also apply these ideas to make routes easier to diff.

ilyapuchka:
@zzcgumn the template already has a code needed to validate routes. But regardless of that I think these changes are not directly related to this suggestion, though I've started to bootstrap a separate proposal that would possible replace `perform(:::)` method

Cookbook/Proposals/SmarterDiffs.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+As can be seen from the diff it does not show in what exact array item the failure happened. This is something that we will need to adjust along with [this issue](https://github.com/krzysztofzablocki/Difference/issues/1). 
+
+Proposed way to integrate this feature is to copy the implementation (with proper author attribution) and adjust it to our needs, optionally we can make pull requests to the original library.
+
+## Impact on existing codebase
+
+This approach can be rolled out gradually and used for new test (Xcode templates will use a new assert) or tests can be refactored in batches by replacing existing assertions with a new assert.
```
mr-v: Given all issues with library, see my main review comment, I'd make it a non-default assertion. Something we can use when tests we're working on fails.
```swift
/// keep using `equal` for state tests
expect(states).to(equal([
    makeState(status: .idle),
    ...
)
// and if above fails we can switch to reflection based message for debugging or development
expect(states).to(equalWithDiff([
    makeState(status: .idle),
    ...
)
```

I'd keep it this way until we have more experience with it and have good idea in what areas we'd need to improve messages.

Cookbook/Proposals/SmarterDiffs.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+As can be seen from the diff it does not show in what exact array item the failure happened. This is something that we will need to adjust along with [this issue](https://github.com/krzysztofzablocki/Difference/issues/1). 
+
+Proposed way to integrate this feature is to copy the implementation (with proper author attribution) and adjust it to our needs, optionally we can make pull requests to the original library.
```
mr-v: Having a local copy might make it a bit easier to work on fixing the issues with library. Once we're confident in our fixes, we could submit PR to original library. Not sure how much of an overhead it would be to maintain our fork.

Cookbook/Proposals/SmarterDiffs.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Impact on existing codebase
+
+This approach can be rolled out gradually and used for new test (Xcode templates will use a new assert) or tests can be refactored in batches by replacing existing assertions with a new assert.
+
+## Alternatives considered
+
+As an alternative we can enforce asserting on individual states instead of all states which will make diffs less of an issue. But as mentioned before it will not eliminate it.
```
mr-v: I like the idea of seeing all the states. It makes it a little bit easier to figure out how given view model works. Plus, those tests are already a bit broad, so I'd say it's fine. Definitely agree with downsides you pointed out in "Option 1 - assert states one by one".

Cookbook/Proposals/SmarterDiffs.md
```diff
@@ -0,0 +1,97 @@
...
+
+Another alternative would be to use snapshot testing that we already use for UI tests. It will make our assert just a call to a single `assertSnapshot` function. But at the same time it will mean that all the snapshots will be written into the files rather than explicitly stated in the test code. This will decrease the boilerplate we will have to write in tests manually but will make it harder to understand test expectations as as we will need to search through text files to find values we need.
```
mr-v: I wonder if it would be possible to use snapshot library infrastructure to capture diffs on the fly in memory without storing them in files. It would work like `Difference` as in produce snapshots only on failure.
Would require some more usage to see how useful are the diffs it provides. I think we have less experience with this than with `Difference`, so it's hard to tell.

mr-v (commented):
I very much like the idea and I've been using this `Difference` library for ad-hoc debugging of tests with great effect. When it works, it's brilliant.

I'm just not sure whether we should use it as a default to provide error messages.
I previously also used it in a demo project and found a number of issues ([see PR](https://github.com/krzysztofzablocki/Difference/pull/6)).

From more recent experience I had following problems:
- Let's say property `a` is not equal in two states and you have property `b` with equal `Set<Int>`. In this case `Difference` produced misleading two error messages (for `a` and `b`) even though problem was only with `a`.
- It might not always produce reliable message for enums with associated values (can't recall specific scenario though).
- If we have different in states count it might also not produce helpful message. I think I added assertion on `count` to work around this.

Cookbook/Proposals/SmarterDiffs.md
ilyapuchka: The convenience of snapshots is that we don't have to specify the expected values since they are in written the file. If we don't write them then we have to construct them on every run so we end up with the same as regular asserts, maybe a bit different API.

Cookbook/Proposals/SmarterDiffs.md
ilyapuchka: Yes, probably.
The issues you highlighted seem all to relate to arrays to items, so we can try to mitigate them by applying diff algorithm after we apply reflection so that it will calculate mutations - if there are no mutations needed - assert passes, otherwise we see how things moved around.

MichalTKwiecien (approved):

brocoo (approved):

karabatov (approved):

adiki (approved):

dmcrodrigues (approved):
Good initiative 👍

Ben-Henshall (approved):
I really like this.

It doesn't mention nested properties in the library `README` at all. 
Say if I have `HealthCategory` that has a `HealthMetric` property and that `HealthMetric` property has a string property. If I compare two `HealthCategory`s, but the `HealthMetrics` objects have differing String properties, will the diff be flagged as the Strings being different? Or will it just flag that the two `HealthMetric` objects are different?

It would be great if it correctly flagged these nested differences, as HealthCheck has 3-4 layers deep of models so comparing `HealthCatgory`s is a nightmare.

@mr-v Any idea?

TheAdamBorek (approved):

inamiy (approved):

zzcgumn (approved):

AliSoftware:
Interesting very simple solution to implement that concept: https://twitter.com/mdiep/status/1224680641633181697

PR #199, opened 2019-07-26
Managing Assets

adiki:

### DIFF
--- a/Cookbook/Proposals/ManagingAssets.md
+++ b/Cookbook/Proposals/ManagingAssets.md
@@ -2,11 +2,11 @@
 # Managing Assets
 
 * Author(s): Adrian Śliwa
-* Review Manager: TBD
+* Review Manager: Danilo Aliberti
 
 ## Introduction
 
-Babylon's iOS application is a large and advanced software project. This project consisting of many frameworks. Each of them consists of many tools or features, most of which need icons or images to be fully implemented. The goal of this proposal is to standardise the way of adding, changing, removing and accessing assets in our codebase.
+Babylon's iOS application is a large and advanced software project. This project consists of many frameworks. Each of them consists of many tools or features, most of which need icons or images to be fully implemented. The goal of this proposal is to standardise the way of adding, changing, removing and accessing assets in our codebase.
 
 ## Motivation
 
@@ -40,6 +40,18 @@ Keeping every icon and image in one location potentially shouldn’t increase th
 1. We could try to systematize the way we include assets in specific feature frameworks but it can cause problems described in the motivation section.
 
 2. Instead of accessing icons or images by subscripts `designLibrary.tokens.icons.iconography[.close]` we could use new feature of Swift 5.1 `@dynamicMemberLookup` which could be combined with `KeyPath`. Then we could write just `designLibrary.tokens.icons.close`.
-
+To achieve that firstly we have to mark `struct Icons` with `@dynamicMemberLookup`. Then `ImageIdentifier` has to become `struct` with `String` properties with default value:
+```
+struct ImageIdentifier {
+	let close = "close"
+}
+```
+Later we need create property `let imageIdentifier = ImageIdentifier()` inside `struct Icons`. Having all of these we can finally write dynamicMember subscript:
+```
+subscript(dynamicMember keyPath: KeyPath<ImageIdentifier, String>) -> UIImage {
+    return Icons.image(for: imageIdentifier[keyPath: keyPath])
+}
+```
+And finally use it like: `designLibrary.tokens.icons.close`. The only drawback of this approach is the fact that for new icons we have to create new property and assign the default value to it compared to just create new enum case with the name matching asset name.
 3. We are going to use SwiftGen tool to auto-generate localizable strings identifiers. Having that tool in place we could also use it to generate assets identifiers.
 
--- a/Cookbook/Proposals/ManagingAssets.md
+++ b/Cookbook/Proposals/ManagingAssets.md
@@ -18,44 +18,11 @@ Managing assets is not a difficult engineering problem but without standardised 
 
 To make sure we won't have duplicated assets in the project and all of them will be accessible in all places where we implement features, all of us should agree on uniform rules about how and where we should add new assets and how we should manage them.
 
+To automate process of generating assets' identifiers we are going to use SwiftGen (agreed solution for Localization https://github.com/Babylonpartners/ios-playbook/pull/187). Having that tool in place we could also use it to generate assets' identifiers.
 
 ## Proposed solution
 
-The proposed solution contains a set of rules which should be followed during manipulation of our assets (this example covers case for icons and analogous solution should be applied for other kinds of assets):
-1. All new icons and images we are going to use should be added to `BabylonDependencies.framework` in `Assets.xcassets` catalog 
-2. `struct DesignLibrary.Tokens` should have another property for icons `public let icons: Icons`
-3. Then we have to add a new case to `enum ImageIdentifier` which will be embedded in `struct Icons`.
-4. To access e.g. close icon we could use `subscript`: `designLibrary.tokens.icons[.close]`
-5. If we are updating an icon we should check if any other place uses it. If so and that other place shouldn't be updated we should create a new icon with an updated design.
-6. If we are removing some code which is using some icons we should check if that code was the last place which was using given icon, if so icon should be deleted from `Assets.xcassets`.
-7. To support the overriding of standard icons in white label apps we should update `ImageCatalogueAware.image(for imageIdentifier:in bundle: compatibleWith traitCollection:) -> UIImage` to firstly access image from `Bundle.main` and then fallback to `BabylonDependencies` if image was not overridden. We can also add support to specify from which location we would like to access given image.
-
-## Impact on existing codebase
-
-Unfortunately, in our codebase assets are located in different places. If we will agree on above set of rules, they should be applied for newly created assets. We should also make an effort to eliminate technical debt and migrate all existing assets into `BabylonDependencies`. During that process, we should have in mind that eventually, we will use SwiftGen for assets and it is good opportunity to improve assets' names for SwiftGen integration. Keeping every asset in one location potentially shouldn’t increase the size of the target application on the condition that the app has all our frameworks linked.
-
-## Alternatives considered
-
-1. We could try to systematize the way we include assets in specific feature frameworks but it can cause problems described in the motivation section.
-	
-2. Instead of accessing icons or images by subscripts `designLibrary.tokens.icons[.close]` we could use new feature of Swift 5.1 `@dynamicMemberLookup` which could be combined with `KeyPath`. Then we could write just `designLibrary.tokens.icons.close`.
-To achieve that firstly we have to mark `struct Icons` with `@dynamicMemberLookup`. Then `ImageIdentifier` has to become `struct` with `String` properties with default value:
-```
-struct ImageIdentifier {
-    let close = "close"
-}
-```
-Later we need create property `let imageIdentifier = ImageIdentifier()` inside `struct Icons`. Having all of these we can finally write dynamicMember subscript:
-```
-subscript(dynamicMember keyPath: KeyPath<ImageIdentifier, String>) -> UIImage {
-    return Icons.image(for: imageIdentifier[keyPath: keyPath])
-}
-```
-And finally use it like: `designLibrary.tokens.icons.close`. The only drawback of this approach is the fact that for new icons we have to create new property and assign the default value to it, compared to just create new `enum` `case` with the name matching asset name. The small benefit here is to have a little bit better syntax while keeping adding new assets simple.
-
-3. We are going to use SwiftGen tool to auto-generate localizable strings identifiers (https://github.com/Babylonpartners/ios-playbook/pull/187). Having that tool in place we could also use it to generate assets identifiers.
-
-`SwiftGen` is generating 0-case enum called `Assets` with `static let` corresponding to a given asset:
+`SwiftGen` is generating 0-case enum called `Asset` with `static let`s corresponding to a given asset:
 ```
 enum Asset {
     static let close = ImageAsset(name: "Close")
@@ -69,9 +36,47 @@ or
 ```
 let sameCloseImage = Asset.close.image
 ```
-Ultimately we would like to limit SwiftGen impact on our codebase to absolute minimum. That's why I think we can use it alongside with `@dynamicMemberLookup`. There are couple of issues which has to be resolved first. If we want to use `KeyPath` we would have to be able to create an instance of `Asset`. It's not possible with 0-case enum that's why we should change it to `struct`. Another thing is to change `static let`s to just `let`s because `Dynamic key path member lookup cannot refer to static member 'close'`. To match this requirements we can just create custom templates in SwiftGen.
 
-Subscript for SwiftGen's `Asset` will look like this:
+Untimatelly we would like to have a solution which allows us just add new asset and be able to use it immediatelly:
+1. To add new asset e.g. `close icon` to the project it just need to be placed in `Assets.xcassets` catalog in `BabylonDependencies.framework`
+2. And then just use it like e.g: `designLibrary.tokens.icons[Asset.close]` (Xcode will report an error that will be fixed by recompiling).
+
+To achive this level of simplicity we have to:
+1. Extend `SwiftGen` configuration to create identifiers for assets.
+2. Add `struct`s to `DesignLibrary.Tokens`/`DesignLibrary` for each type of asset we are supproting e.g. `struct Icons`.
+3. Add `subscript` to particular `struct` responsible for accessing given asset e.g.:
+```
+subscript(imageAsset: ImageAsset) -> UIImage {
+    get {
+        return imageAsset.image
+    }
+}
+```
+This example uses implementation of accessing actual image provided by `SwiftGen`. To support the overriding of standard icons in white label apps we should update this implementation to firstly access image from `Bundle.main` and then fallback to `BabylonDependencies` if image was not overridden. We can also add support to specify from which location we would like to access given image.
+
+Besides standard rules of adding new assets we should also take care when:
+1. We are updating an icon we should check if any other place uses it. If so and that other place shouldn't be updated we should create a new icon with an updated design.
+2. If we are removing some code which is using some icons we should check if that code was the last place which was using given icon, if so icon should be deleted from `Assets.xcassets`. This step could be automated: if we are not using specific image its identifier won't be used across the project. We can write the script which is checking the usage of every asset and when it is not used anywhere the asset in `Assets.xcassets` catalog can be deleted.
+
+## Impact on existing codebase
+
+Unfortunately, in our codebase assets are located in different places. If we will agree on above set of rules, they should be applied for newly created assets. We should also make an effort to eliminate technical debt and migrate all existing assets into `BabylonDependencies`. During that process, we should have in mind it is good opportunity to improve assets' names for SwiftGen integration. Keeping every asset in one location potentially shouldn’t increase the size of the target application on the condition that the app has all our frameworks linked.
+
+## Alternatives considered
+
+1. We could try to systematize the way we include assets in specific feature frameworks but it can cause problems described in the motivation section.
+	
+2. Instead of accessing icons or images by subscripts `designLibrary.tokens.icons[Asset.close]` we could use new feature of Swift 5.1 `@dynamicMemberLookup` which could be combined with `KeyPath`. Then we could write just `designLibrary.tokens.icons.close`.
+To achieve that firstly we have to mark `struct Icons` with `@dynamicMemberLookup`. Then `enum Asset` has to become `struct`. Its `static let`s has to became just `let`s because `Dynamic key path member lookup cannot refer to static member 'close'`. To match this requirements we can just create custom templates in SwiftGen.
+`struct Asset` will look like this:
+```
+public struct Asset {
+    public let add = ImageAsset(name: "Add")
+    public let close = ImageAsset(name: "Close")
+  ...
+}
+```
+Then we have to write special subscript:
 ```
 @dynamicMemberLookup
 public struct Icons {
@@ -82,8 +87,4 @@ public struct Icons {
     }
 }
 ```
-
-Having all this pieces in place adding new asset will be very simple and will include two steps only:
-1. Adding assets to the catalog
-2. And just use it like e.g: `designLibrary.tokens.icons.close` (Xcode will report an error that will be fixed by recompiling).
-
+Having this this pieces in place call side will be really simple and clean: `designLibrary.tokens.icons.close`
--- a/Cookbook/Proposals/ManagingAssets.md
+++ b/Cookbook/Proposals/ManagingAssets.md
@@ -25,7 +25,7 @@ The proposed solution contains a set of rules which should be followed during ma
 1. All new icons and images we are going to use should be added to `BabylonDependencies.framework` in `Assets.xcassets` catalog 
 2. `struct DesignLibrary.Tokens` should have another property for icons `public let icons: Icons`
 3. Then we have to add a new case to `enum ImageIdentifier` which will be embedded in `struct Icons`.
-4. To access e.g. close icon we could use subscripts `designLibrary.tokens.icons[.close]`
+4. To access e.g. close icon we could use `subscript`: `designLibrary.tokens.icons[.close]`
 5. If we are updating an icon we should check if any other place uses it. If so and that other place shouldn't be updated we should create a new icon with an updated design.
 6. If we are removing some code which is using some icons we should check if that code was the last place which was using given icon, if so icon should be deleted from `Assets.xcassets`.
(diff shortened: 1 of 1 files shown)


Cookbook/Proposals/ManagingAssets.md
```diff
@@ -0,0 +1,54 @@
...
+	...
+	public var close: UIImage {
+	    return Iconography.image(for: "close")
+	}
+}
+```
+then use `designLibrary.tokens.icons.iconography.close`.
+That solution will require to write more code per each icon. There is no extra value here because autocomplition should also work with `ImageIdentifier`'s as subscript argument.
+
+3. We are going to use SwiftGen tool to auto-generate localizable strings identifiers. Having that tool in place we could also use it to generate assets identifiers.
```
sergdort: Yes please 🙏

inamiy (approved):

Cookbook/Proposals/ManagingAssets.md
```diff
@@ -0,0 +1,54 @@
...
+To make sure we won't have duplicated icons in the project and all icons will be accessible in all places where we implement the user interface, all of us should agree on uniform rules how and where we should add new assets and how we should manage them.
+
+
+## Proposed solution
+
+The proposed solution contains a set of rules which should be followed during manipulation of our assets:
+1. All new icons and images we are going to use should be added to `BabylonDependencies.framework` in `Assets.xcassets` catalog 
+2. `struct DesignLibrary.Tokens` should have another property for icons `public let icons: Icons`
+3. Icons should be namespaced and put into `struct` with common relation/meaning (e.g. `TabBarIcons`, `Iconography` etc.). If there is no existing `struct` we should create one. Then we have to add a new case to `enum ImageIdentifier` which is embedded in given `struct`.
+4. To access e.g. iconography's close icon we could use subscripts `designLibrary.tokens.icons.iconography[.close]`
```
ilyapuchka: this seems overly verbose, I'd suggest we have it reduced to `designLibrary.icons[.close]` and have all these details inside this subscript if we need them for some reason

Cookbook/Proposals/ManagingAssets.md
```diff
@@ -0,0 +1,54 @@
...
+6. If we are removing some code which is using some icons we should check if that code was the last place which was using given icon, if so icon should be deleted from `Assets.xcassets`.
+7. To support the overriding of standard icons in white label apps we should update `ImageCatalogueAware.image(for imageIdentifier:in bundle: compatibleWith traitCollection:) -> UIImage` to firstly access image from `Bundle.main` and then fallback to `BabylonDependencies` if image was not overridden. We can also add support to specify from which location we would like to access given image.
+
+## Impact on existing codebase
+
+Unfortunately, in our codebase icons and images are located in different places. If we will agree on above set of rules, they should be applied for newly created assets. We should also make an effort to eliminated technical debt and migrate all existing icons into `BabylonDependencies`.
+Keeping every icon and image in one location potentially shouldn’t increase the size of the target application on the condition that the app has all our frameworks linked.
+## Alternatives considered
```
ilyapuchka:
```suggestion

## Alternatives considered
```

Cookbook/Proposals/ManagingAssets.md
```diff
@@ -0,0 +1,54 @@
...
+4. To access e.g. iconography's close icon we could use subscripts `designLibrary.tokens.icons.iconography[.close]`
+5. If we are updating an icon we should check if any other place uses it. If so and that other place shouldn't be updated we should create a new icon with an updated design.
+6. If we are removing some code which is using some icons we should check if that code was the last place which was using given icon, if so icon should be deleted from `Assets.xcassets`.
+7. To support the overriding of standard icons in white label apps we should update `ImageCatalogueAware.image(for imageIdentifier:in bundle: compatibleWith traitCollection:) -> UIImage` to firstly access image from `Bundle.main` and then fallback to `BabylonDependencies` if image was not overridden. We can also add support to specify from which location we would like to access given image.
+
+## Impact on existing codebase
+
+Unfortunately, in our codebase icons and images are located in different places. If we will agree on above set of rules, they should be applied for newly created assets. We should also make an effort to eliminated technical debt and migrate all existing icons into `BabylonDependencies`.
```
ilyapuchka:
```suggestion
Unfortunately, in our codebase icons and images are located in different places. If we will agree on above set of rules, they should be applied for newly created assets. We should also make an effort to eliminate technical debt and migrate all existing icons into `BabylonDependencies`.
```

Cookbook/Proposals/ManagingAssets.md
```diff
@@ -0,0 +1,54 @@
...
+```
+public struct Iconography: ImageCatalogueAware
+	...
+	public var close: UIImage {
+	    return Iconography.image(for: "close")
+	}
+}
+```
+then use `designLibrary.tokens.icons.iconography.close`.
+That solution will require to write more code per each icon. There is no extra value here because autocomplition should also work with `ImageIdentifier`'s as subscript argument.
```
ilyapuchka: we won't need to write those properties in Swift 5.1 though as it supports key paths in `@dynamicMemberLookup`. Favourite feature of @andersio

ilyapuchka (approved):

Cookbook/Proposals/ManagingAssets.md
```diff
@@ -0,0 +1,54 @@
+
+# Managing Assets
+
+* Author(s): Adrian Śliwa
+* Review Manager: TBD
+
+## Introduction
+
+Babylon's iOS application is a large and advanced software project. This project consisting of many frameworks. Each of them consists of many tools or features, most of which need icons or images to be fully implemented. The goal of this proposal is to standardise the way of adding, changing, removing and accessing assets in our codebase.
```
daniloaliberti:
```suggestion
Babylon's iOS application is a large and advanced software project. This project consists of many frameworks. Each of them consists of many tools or features, most of which need icons or images to be fully implemented. The goal of this proposal is to standardise the way of adding, changing, removing and accessing assets in our codebase.
```

Cookbook/Proposals/ManagingAssets.md
```diff
@@ -0,0 +1,58 @@
...
+
+To make sure we won't have duplicated icons in the project and all icons will be accessible in all places where we implement the user interface, all of us should agree on uniform rules how and where we should add new assets and how we should manage them.
+
+
+## Proposed solution
+
+The proposed solution contains a set of rules which should be followed during manipulation of our assets:
+1. All new icons and images we are going to use should be added to `BabylonDependencies.framework` in `Assets.xcassets` catalog 
+2. `struct DesignLibrary.Tokens` should have another property for icons `public let icons: Icons`
+3. Then we have to add a new case to `enum ImageIdentifier` which will be embedded in `struct Icons`.
```
AliSoftware: Shouldn't we just have `struct Icons` + `static let close = Icon(...)` instead of having an `enum ImageIdentifier`? Or make `enum Icons` directly?

Cookbook/Proposals/ManagingAssets.md
```diff
@@ -0,0 +1,58 @@
...
+4. To access e.g. close icon we could use subscripts `designLibrary.tokens.icons[.close]`
+5. If we are updating an icon we should check if any other place uses it. If so and that other place shouldn't be updated we should create a new icon with an updated design.
+6. If we are removing some code which is using some icons we should check if that code was the last place which was using given icon, if so icon should be deleted from `Assets.xcassets`.
+7. To support the overriding of standard icons in white label apps we should update `ImageCatalogueAware.image(for imageIdentifier:in bundle: compatibleWith traitCollection:) -> UIImage` to firstly access image from `Bundle.main` and then fallback to `BabylonDependencies` if image was not overridden. We can also add support to specify from which location we would like to access given image.
+
+## Impact on existing codebase
+
+Unfortunately, in our codebase icons and images are located in different places. If we will agree on above set of rules, they should be applied for newly created assets. We should also make an effort to eliminate technical debt and migrate all existing icons into `BabylonDependencies`.
```
AliSoftware: Should we rename the icons (at least their constant in `DesignLibrary.Tokens`) at the same time as we migrate them in `BabylonDependencies`, in preparation for adopting the SwiftGen proposal presented by David recently?

Cookbook/Proposals/ManagingAssets.md
```diff
@@ -0,0 +1,58 @@
...
+    let close = "close"
+}
+```
+Later we need create property `let imageIdentifier = ImageIdentifier()` inside `struct Icons`. Having all of these we can finally write dynamicMember subscript:
+```
+subscript(dynamicMember keyPath: KeyPath<ImageIdentifier, String>) -> UIImage {
+    return Icons.image(for: imageIdentifier[keyPath: keyPath])
+}
+```
+And finally use it like: `designLibrary.tokens.icons.close`. The only drawback of this approach is the fact that for new icons we have to create new property and assign the default value to it, compared to just create new `enum` `case` with the name matching asset name.
```
AliSoftware: You describe the drawback here, but what are the benefits of doing that, especially compared to what you proposed in the previous section (line 28) that would directly allow `designLibrary.tokens.icons.close`?

Cookbook/Proposals/ManagingAssets.md
AliSoftware: ❤️ 
(Maybe link to David's proposal btw)

Cookbook/Proposals/ManagingAssets.md
```diff
@@ -0,0 +1,58 @@
+
+# Managing Assets
+
+* Author(s): Adrian Śliwa
+* Review Manager: Danilo Aliberti
+
+## Introduction
+
+Babylon's iOS application is a large and advanced software project. This project consists of many frameworks. Each of them consists of many tools or features, most of which need icons or images to be fully implemented. The goal of this proposal is to standardise the way of adding, changing, removing and accessing assets in our codebase.
```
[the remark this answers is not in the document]
adiki: Yes, I agree with you. I will update the proposal and provide an example for images that could be applied to other kinds of assets.

Cookbook/Proposals/ManagingAssets.md
adiki: Having `enum ImageIdentifier` allows us to write the smallest amount of code to add support for a new icon. This thing here is syntactic sugar and I'm open to discussing with the whole team which solution here will be the most convenient one. Today I'm going to check how SwiftGen will fit with @dynamicMemberLookup, maybe it will be the most convenient solution.

Cookbook/Proposals/ManagingAssets.md
adiki: Yes, I agree. I will add information about that.

Cookbook/Proposals/ManagingAssets.md
adiki: The benefit here a little better syntax (keeping addition simple) but it's a matter of taste.

AliSoftware: But you already have the smaller syntax on line 28 though

adiki: `designLibrary.tokens.icons[.close]`
`designLibrary.tokens.icons.close`
Second one is shorter and the second one uses `@dynamicMemberLookup`. 
If I misunderstand you we should find the same perspective here.

AliSoftware: Ohhh those square brackets were supposed to be real, literal, square brackets, of course 🤦‍♂️

Idk why but I read that `designlibrary.tokens.icons[.close]` on line 28 as some syntax supposed to mean "`designlibrary.tokens.icons.close` but replace `.close` here with whatever icon name you want", as if the square brackets were not part of the actual Swift expression but were there to explain the presence of a placeholder… so that's why I didn't read it as if it were literal square brackets here.

But in fact your suggestion is to implement `subscript` on `Icon` there, makes sense now, don't know how I didn't see that on first read, my bad (maybe mentioning `subscript` on the proposed solution around line 28 could help?)

Then maybe we could add my suggestion / what I thought it would be in the first place (make `struct Icon` define its icons using `static let close = …`) in the alternatives considered?

adiki: Sure I will add it :)

adiki: I would rather add alternative as computed static var `static var close: UIImage`. `static let` is lazy loaded and after that kept in memory. I don't think we would like to flood memory with many `UIImage`s

adiki: Hi @AliSoftware I extended SwiftGen section. Please let me know if you still would like me to add `struct Icon` and  `static let close = ...` I'm not sure if it could be a simpler approach then what is proposed is SwiftGen section but you can try to convince me :)

Cookbook/Proposals/ManagingAssets.md
karabatov: Which is how it's working now, despite images being defined in `Iconography`.

Cookbook/Proposals/ManagingAssets.md
```diff
@@ -0,0 +1,94 @@
...
+7. To support the overriding of standard icons in white label apps we should update `ImageCatalogueAware.image(for imageIdentifier:in bundle: compatibleWith traitCollection:) -> UIImage` to firstly access image from `Bundle.main` and then fallback to `BabylonDependencies` if image was not overridden. We can also add support to specify from which location we would like to access given image.
+
+## Impact on existing codebase
+
+Unfortunately, in our codebase assets are located in different places. If we will agree on above set of rules, they should be applied for newly created assets. We should also make an effort to eliminate technical debt and migrate all existing assets into `BabylonDependencies`. During that process, we should have in mind that eventually, we will use SwiftGen for assets and it is good opportunity to improve assets' names for SwiftGen integration. Keeping every asset in one location potentially shouldn’t increase the size of the target application on the condition that the app has all our frameworks linked.
```
AliSoftware: 👍

Cookbook/Proposals/ManagingAssets.md
```diff
@@ -0,0 +1,94 @@
...
+```
+Then you can use it like this:
+```
+let closeImage = UIImage(asset: Asset.close) 
+```
+or 
+```
+let sameCloseImage = Asset.close.image
+```
+Ultimately we would like to limit SwiftGen impact on our codebase to absolute minimum. That's why I think we can use it alongside with `@dynamicMemberLookup`. There are couple of issues which has to be resolved first. If we want use `KeyPath` we would have be able to create instance of Asset. It's not possible with 0-case enum that's why we should change it to struct. Another thing is to change `static let`s to just `let`s because `Dynamic key path member lookup cannot refer to static member 'close'`. To match this requirements we could add new Run Script phase similar to this:
```
AliSoftware:
```suggestion
Ultimately we would like to limit SwiftGen impact on our codebase to absolute minimum. That's why I think we can use it alongside with `@dynamicMemberLookup`. There are couple of issues which has to be resolved first. If we want to use `KeyPath` we would have to be able to create an instance of `Asset`. It's not possible with 0-case enum that's why we should change it to `struct`. Another thing is to change `static let`s to just `let`s because `Dynamic key path member lookup cannot refer to static member 'close'`. To match this requirements we can just create custom templates in SwiftGen.
```

Cookbook/Proposals/ManagingAssets.md
```diff
@@ -0,0 +1,94 @@
...
+```
+$SRCROOT/../swiftgen/bin/swiftgen
+sed -i -e 's/enum/struct/g' $SRCROOT/AssetsDependencies/assets-images.swift
+sed -i -e 's/static let/let/g' $SRCROOT/AssetsDependencies/assets-images.swift
+```
```
AliSoftware: This is not necessary. SwiftGen is based on templates which let you customise everything you need (that what makes it powerful and flexible as opposed to other similar tools 😉 ). No need for that kind of hack there!

adiki: That's great! Let me update that ;)

Cookbook/Proposals/ManagingAssets.md
```diff
@@ -0,0 +1,89 @@
...
+
+    public subscript(dynamicMember keyPath: KeyPath<Asset, ImageAsset>) -> UIImage {
+        return asset[keyPath: keyPath].image
+    }
+}
+```
+
+Having all this pieces in place adding new asset will be very simple and will include two steps only:
+1. Adding assets to the catalog
+2. And just use it like e.g: `designLibrary.tokens.icons.close` (it will cause the error which will be fixed during compliation).
```
zzcgumn:
```suggestion
2. And just use it like e.g: `designLibrary.tokens.icons.close` (Xcode will report an error that will be fixed by recompiling).
```

zzcgumn:
I am all in favour of centralising all assets to one place and standardise how they are accessed.

Since I am not at all familiar with SwiftGen I cannot comment on different technical solutions.

zzcgumn (approved):

Cookbook/Proposals/ManagingAssets.md
dmcrodrigues: Do we really need to have this as part of the design library? And why are we keeping this management a manual process? Why not relying on SwiftGen to avoid maintaining all of this ourselves?

Cookbook/Proposals/ManagingAssets.md
```diff
@@ -0,0 +1,89 @@
...
+```
+Then you can use it like this:
+```
+let closeImage = UIImage(asset: Asset.close) 
+```
+or 
+```
+let sameCloseImage = Asset.close.image
+```
+Ultimately we would like to limit SwiftGen impact on our codebase to absolute minimum. That's why I think we can use it alongside with `@dynamicMemberLookup`. There are couple of issues which has to be resolved first. If we want to use `KeyPath` we would have to be able to create an instance of `Asset`. It's not possible with 0-case enum that's why we should change it to `struct`. Another thing is to change `static let`s to just `let`s because `Dynamic key path member lookup cannot refer to static member 'close'`. To match this requirements we can just create custom templates in SwiftGen.
```
dmcrodrigues: > Ultimately we would like to limit SwiftGen impact on our codebase to absolute minimum.

Why? We already approved its usage for localisable strings and SwiftGen generates an enumeration so the impact is not that significant as we can remove it and keep the same enumeration or not?

dmcrodrigues (dismissed):
The proposed solution addresses the problem of centralisation but doesn't address the productivity of the team in managing them, keeping identifiers manually is something that takes development time from us and doesn't provide any benefit or type safety, considering we already accepted SwiftGen for localisable strings I don't see why we are not suggesting a solution where we used it to have the same benefits.

Cookbook/Proposals/ManagingAssets.md
adiki: When I started writing this proposal there was no agreement if we are going to use SwiftGen or not. That's why I decided to include SwiftGen in `Alternatives considered`. Because now we agreed to use SwiftGen I will move it as the proposed solution. We don't have to have it in DesignLibrary, but it's a place where you can access icons now, that's why I kept that solution. Would you like to propose to have it as a separated thing and give some pros of that approach?

Cookbook/Proposals/ManagingAssets.md
adiki: If we are using a flat structure of single-level enum this is minimal impact and I think that's the same thing we think about. What I meant is to not use nested levels of namespacing which potentially could spread out through codebase. Maybe sometime in the future, Apple will add a tool for this kind of purpose then the transition to it will be simple.

adiki:
> The proposed solution addresses the problem of centralisation but doesn't address the productivity of the team in managing them, keeping identifiers manually is something that takes development time from us and doesn't provide any benefit or type safety, considering we already accepted SwiftGen for localisable strings I don't see why we are not suggesting a solution where we used it to have the same benefits.

```
When I started writing this proposal there was no agreement if we are going to use SwiftGen or not. That's why I decided to include SwiftGen in Alternatives considered. Because now we agreed to use SwiftGen I will move it as the proposed solution.
```

adiki:
@David SwiftGen alternative was moved now to `Proposed Solution. There is still one improvement we can add when we will support Swift 5.1. Described in `Alternatives considered`.

karabatov (approved):

karabatov:
Maybe we can avoid the subscript altogether by returning the type of the generated enum from the design library?

``` swift
// Generated asset enum.
enum Asset {
    static let close = "Close"
}

struct DesignLibrary {
    // Return the type itself so we can use static members.
    var assets = Asset.self
}

let designLibrary = DesignLibrary()

// Use `close`…
print(designLibrary.assets.close)
```

adiki:
During the meeting we decided to:
1. Use a flat structure for assets which increases discoverability (the same approach was chosen for localization)
2. Access assets by abstraction generated by `SwiftGen` e.g. Assets.close.image. This abstraction is simple and shouldn't cause problems if we would like to move to another solution because we could keep it as it is.
3. We are not going to use @dynamicMemberLookup firstly because we are not supporting Swift 5.1 yet secondly because we agreed on point 2 and using @dynamicMemberLookup is not giving us any more value here.

The document will be updated to include those decisions.

I will also spend some more time to check how to solve overriding default assets in application-specific targets.

adiki:
> Maybe we can avoid the subscript altogether by returning the type of the generated enum from the design library?
> 
> ```swift
> // Generated asset enum.
> enum Asset {
>     static let close = "Close"
> }
> 
> struct DesignLibrary {
>     // Return the type itself so we can use static members.
>     var assets = Asset.self
> }
> 
> let designLibrary = DesignLibrary()
> 
> // Use `close`…
> print(designLibrary.assets.close)
> ```

That's really good apprach! but I think we decided to just use Asset.close.image.

emeseuk (approved):

Cookbook/Proposals/ManagingAssets.md
```diff
@@ -0,0 +1,111 @@
...
+}
+```
+
+We should also configure SwiftGen to generate identifiers for additional assets provided by application-specific targets. We can configure `SwiftGen` to name `enum`s like `enum TargetAsset` to not have conflicts with BabylonDependencies' `enum Asset`.
+
+### Other rules
+ 
+Besides standard rules of adding new assets we should also take care when: 
+1. We are updating an icon we should check if any other place uses it. If so and that other place shouldn't be updated we should create a new icon with an updated design. 
+2. If we are removing some code which is using some icons we should check if that code was the last place which was using a given icon, if so icon should be deleted from `Assets.xcassets`. This step could be automated: if we are not using a specific image its identifier won't be used across the project. We can write the script which is checking the usage of every asset and when it is not used anywhere the asset in `Assets.xcassets` the catalog can be deleted.
```
daniloaliberti: this part is a bit confused 

> the asset in `Assets.xcassets` the catalog can be deleted

daniloaliberti:
```suggestion
2. If we are removing some code which is using some icons we should check if that code was the last place which was using a given icon, if so icon should be deleted from `Assets.xcassets`. This step could be automated: if we are not using a specific image its identifier won't be used across the project. 
We can write a script which checks the usage of every asset and when it is not used anywhere, it can be deleted. 
```

Cookbook/Proposals/ManagingAssets.md
```diff
@@ -0,0 +1,111 @@
...
+2. If we are removing some code which is using some icons we should check if that code was the last place which was using a given icon, if so icon should be deleted from `Assets.xcassets`. This step could be automated: if we are not using a specific image its identifier won't be used across the project. We can write the script which is checking the usage of every asset and when it is not used anywhere the asset in `Assets.xcassets` the catalog can be deleted. 
+
+## Impact on existing codebase
+ 
+Unfortunately, in our codebase assets are located in different places. If we will agree on the above set of rules, they should be applied for newly created assets. We should also make an effort to eliminate technical debt and migrate all existing assets into `BabylonDependencies`. During that process, we should have in mind it is a good opportunity to improve assets' names for SwiftGen integration. Keeping every asset in one location potentially shouldn’t increase the size of the target application on the condition that the app has all our frameworks linked.
```
daniloaliberti:
```suggestion
Unfortunately, in our codebase assets are located in different places. If we will agree on the above set of rules, they should be applied for newly created assets. 

We should also make an effort to eliminate technical debt and migrate all existing assets into `BabylonDependencies`. During that process, we should have in mind it is a good opportunity to improve assets' names for SwiftGen integration. Keeping every asset in one location potentially shouldn’t increase the size of the target application on the condition that the app has all our frameworks linked. 
```

Cookbook/Proposals/ManagingAssets.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Alternatives considered 
+
+1. We could try to systematize the way we include assets in specific feature frameworks but it can cause problems described in the motivation section. 
+
+2. (`@deprecated - This point was discussed at the meeting, because we are not including assets in the design library and we are using SwiftGen abstraction to access assets, usage of @dynamicMemberLookup is not adding us any extra value.`) Instead of accessing icons or images by subscripts `designLibrary.tokens.icons[Asset.close]` we could use a new feature of Swift 5.1 `@dynamicMemberLookup` which could be combined with `KeyPath`. Then we could write just `designLibrary.tokens.icons.close`.
```
daniloaliberti: >  because we are not including assets in the design library

Is this true? Today the agreed way is in  a Renderer I'd use:
```appearance.tokens.icons.close```
Assuming `appearance` is the designLibrary.

daniloaliberti (commented):
I think we need to clarify the `DesignLibrary` part. Thanks! 👍

daniloaliberti (approved):
Great work here man! 🚀

PR #200, opened 2019-07-26
[CNSMR-1812] Feature Switches redesign

ilyapuchka:


Cookbook/Proposals/FeatureSwitchesRedesign.md
```diff
@@ -0,0 +1,218 @@
...
+
+
+## Proposed solution
+
+To address these issues we suggest the following changes in our feature switches architecture:
+
+1. Feature toggles and static configurations related to the particular feature should be defined in this feature module instead of shared module or in the app. This follows interface segregation principle and goes along with improving modularisation - one won't need to change common module with feature switches and recompile all of it (and potentially everything else) when they only need to recompile module that uses this feature switch. This way we also configure our frameworks the same way as we configure any other 3rd party framework and can use the similar infrastrauctore for that (i.e. dedicated instances responsibule for configuring each feature rather than doing that all in the app delegate).
```
karabatov: 🥇 I keep thinking of centralisation but this is a very valid approach, with the main benefit being quicker iteration and less recompiling.

I'm not sure about discoverability though (but that should be enforced with code generation).

Cookbook/Proposals/FeatureSwitchesRedesign.md
```diff
@@ -0,0 +1,218 @@
...
+
+2. Replace `AppConfigurationProtocol` and all related protocols with structs. This will decrease amount of boilerplate we have to write when adding new application configuration. Application configuration will be only concerned with application level configurations which are not directly related to any feature module.
+
+3. All feature switches access points should be moved to the `Current` and consequentially should be only used by builders - view models, flow controllers and renderers should have them injected through their constructors.
+
+4. Root.plist should be generated based on the code that declares feature switches.
```
karabatov: You mean using [this](https://github.com/apple/swift-syntax) for generation? Again, not sure right now how that might work.

We could possibly go the other way and use SwiftGen to transform plists into code instead :) But I'd prefer code first tbh.

karabatov (commented):
Overall this is very promising and seems to solve the current pitfalls, good job @ilyapuchka 👍

Cookbook/Proposals/FeatureSwitchesRedesign.md
```diff
@@ -0,0 +1,218 @@
...
+
+All feature specific configuration should be moved to the corresponding feature frameworks and should use concrete structs instead of protocols that then `AppConfiguration` extends. `AppConfiguration` should only be concerned with application level configurations (i.e. enabled tabs or privacy notices urls).
+
+3. All feature switches access points should be moved to the `Current` and consequentially should be only used in builders - view models, flow controllers and renderers should have them injected through their constructors. Currently we only have `Current.abTestingService` but we have global `LocalFeatureSwitches`
```
karabatov: I think there's something wrong with structure here, these are points from before. Probably `3` may be deleted.

And `4` below should be a header instead.

Cookbook/Proposals/FeatureSwitchesRedesign.md
```diff
@@ -0,0 +1,218 @@
...
+
+Without this syncing we would have to deal with optionals for all the switches, falling back to their defaults, defined in code. That also highlights another issue with `Root.plist`, that we need to repeat default value in the code and in the plist and make sure they allign to avoid confusion.
+
+Another issue caused by using `Root.plist` and overrides in this way is when we move from Local feature switch used during development to Remote feature switch used on a final stage of the feautre rollout we have to remember to move the plist entry for this switch from one section to another. Otherwise there will be no way to override it during testing.
```
Ben-Henshall:
```suggestion
Another issue caused by using `Root.plist` and overrides in this way is when we move from Local feature switch used during development to Remote feature switch used on a final stage of the feature rollout we have to remember to move the plist entry for this switch from one section to another. Otherwise there will be no way to override it during testing.
```

Cookbook/Proposals/FeatureSwitchesRedesign.md
```diff
@@ -0,0 +1,218 @@
...
+
+Having a wrong key name in the plist is also the issue that is very easy to miss.
+
+
+### Firebase switches semantics
+
+ATM we use Firebase Remote Config and A/B Tests features at the same time. While indeed for the client it does not matter if the value of a switch is set as a result of a running A/B test or is simply a remote config flag (in fact Firebase A/B Tests are built on top of Remote Config) there are situations when understanding the difference between flags being used for A/B tests or not when reading code instead of consulting with Firebase console is helpful. Right now all our Firebase switches are expressed as A/B test variants though. That created a lot of confusion when we had to consolidate and remove few feature switches related to the NVL and had to figure out if any of them are actually running A/B tests (as the names were not very helpful to identify if some configs were related to NVL or something else on the screens affected by NVL).
```
Ben-Henshall:
```suggestion
At the moment, we use Firebase Remote Config and A/B Tests features at the same time. While indeed for the client it does not matter if the value of a switch is set as a result of a running A/B test or is simply a remote config flag (in fact Firebase A/B Tests are built on top of Remote Config) there are situations when understanding the difference between flags being used for A/B tests or not when reading code instead of consulting with Firebase console is helpful. Right now all our Firebase switches are expressed as A/B test variants though. That created a lot of confusion when we had to consolidate and remove few feature switches related to the NVL and had to figure out if any of them are actually running A/B tests (as the names were not very helpful to identify if some configs were related to NVL or something else on the screens affected by NVL).
```

Cookbook/Proposals/FeatureSwitchesRedesign.md
```diff
@@ -0,0 +1,218 @@
...
+
+
+### Point of access
+
+There is no single way of accessing feature switches in code. Local feature switches are defined in one place, Firebase feature switches are defined in another place, we also have static app configurations and consumer network feature switches (in patient details). It does not create a lot of friction but can be cumbersome to replace local feature switch with remote when feature is due to release.
+At the same time all these places (local/Firebase feature switches, app configuration) became a dumping ground for the switchies for various features of the app, even when they have their own modules. And its not uncommon to have conflicts between PRs which both introduce Firebase feature switches for different issues.
```
Ben-Henshall:
```suggestion
At the same time all these places (local/Firebase feature switches, app configuration) became a dumping ground for the switches for various features of the app, even when they have their own modules. And its not uncommon to have conflicts between PRs which both introduce Firebase feature switches for different issues.
```

Cookbook/Proposals/FeatureSwitchesRedesign.md
```diff
@@ -0,0 +1,218 @@
...
+
+```swift
+@RemoteFeatureSwitch(
+    name: “New feature”,
+    key: “is_new_feature_enabled”
+)
+var isNewFeatureEnabled: Bool
+```
+
+Also we will be able to use `dynamicMemberLookup` with keypaths instead of custom getters or subscripts so that we wil be able to access dynamic feature switches or static configuration as `Current.someFeature.isNewFeatureEnabled` instead of `Current.someFeatures.get(\.isNewFeatureEnabled)`
```
Ben-Henshall:
```suggestion
Also we will be able to use `dynamicMemberLookup` with keypaths instead of custom getters or subscripts so that we will be able to access dynamic feature switches or static configuration as `Current.someFeature.isNewFeatureEnabled` instead of `Current.someFeatures.get(\.isNewFeatureEnabled)`
```

Cookbook/Proposals/FeatureSwitchesRedesign.md
```diff
@@ -0,0 +1,218 @@
...
+
+To solve these issues we propose to replace `protocol AppConfigurationProtocol` with concrete `struct AppConfiguration` defined in `BabylonDependencies` and initialise it with application specific values in a target specific code. This will reduce amount of boilerplate to write, but will still require us to change code multiple times in each target when something changes in this struct. I.e. when we add a new feature switch to `AppConfiguration` we will need to add it to the paramters that we pass to it's initializer in each target. We can use default values where possible to avoid that. For that we should agree that default values are always `false` so that we can definetely know what is the value if it is not passed to the constructor. Protocol and default protocol implementations achive the same result but with much more boilerplate (remember point-free issue about using value types instead of protocols)
```
Ben-Henshall:
```suggestion
To solve these issues we propose to replace `protocol AppConfigurationProtocol` with concrete `struct AppConfiguration` defined in `BabylonDependencies` and initialise it with application specific values in a target specific code. This will reduce amount of boilerplate to write, but will still require us to change code multiple times in each target when something changes in this struct. I.e. when we add a new feature switch to `AppConfiguration` we will need to add it to the parameters that we pass to it's initializer in each target. We can use default values where possible to avoid that. For that we should agree that default values are always `false` so that we can definetely know what is the value if it is not passed to the constructor. Protocol and default protocol implementations achive the same result but with much more boilerplate (remember point-free issue about using value types instead of protocols)
```

Cookbook/Proposals/FeatureSwitchesRedesign.md
```diff
@@ -0,0 +1,218 @@
...
+
+All feature specific configuration should be moved to the corresponding feature frameworks and should use concrete structs instead of protocols that then `AppConfiguration` extends. `AppConfiguration` should only be concerned with application level configurations (i.e. enabled tabs or privacy notices urls).
+
+3. All feature switches access points should be moved to the `Current` and consequentially should be only used in builders - view models, flow controllers and renderers should have them injected through their constructors. Currently we only have `Current.abTestingService` but we have global `LocalFeatureSwitches`
+
+4. Root.plist should be generated based on the code that declared feature switches.
+
+For that we can use SwiftSyntax to analyze the content of the files which define feature switches (based on the file naming convetion) and generate entry in the plist for each of them. This way we don't have to deal with plist manually and have a single source of thruth for defining feature switches. The plist will be ignored by git and will be generated as part of build step only in DEBUG configuration, so we don't have to strip it from release builds.
```
Ben-Henshall:
```suggestion
For that we can use SwiftSyntax to analyze the content of the files which define feature switches (based on the file naming convetion) and generate entry in the plist for each of them. This way we don't have to deal with plist manually and have a single source of truth for defining feature switches. The plist will be ignored by git and will be generated as part of build step only in DEBUG configuration, so we don't have to strip it from release builds.
```

Cookbook/Proposals/FeatureSwitchesRedesign.md
```diff
@@ -0,0 +1,218 @@
...
+
+
+## Impact on existing codebase
+
+Currently used types for feature switches should be deprecated and either replaced with a new approach in one go or gradually. We can keep `ABTestingService` related types but repurpose them to serve as providers of the feature switches rather than 
+
+## Alternatives considered
+
+- Keep all the feature switches in one place in the `BabylonDependencies`. The main downside of this approach if that it goes against modularity - we will need to change flags in the `BabylonDependencies` and rebuild it instead of changing it just in the feature framework and rebuilding only it. It can also trigger rebuild of all other frameworks depending on the `BabylonDependencies` as we know how Xcode incremental builds are unrelyiable.
```
Ben-Henshall:
```suggestion
- Keep all the feature switches in one place in the `BabylonDependencies`. The main downside of this approach if that it goes against modularity - we will need to change flags in the `BabylonDependencies` and rebuild it instead of changing it just in the feature framework and rebuilding only it. It can also trigger rebuild of all other frameworks depending on the `BabylonDependencies` as we know how Xcode incremental builds are unreliable.
```

Cookbook/Proposals/FeatureSwitchesRedesign.md
```diff
@@ -0,0 +1,218 @@
...
+
+```swift
+let isNewFeatureEnabled = FeatureSwitch(
+    name: “New feature”,
+    key: “is_new_feature_enabled”,
+    provider: Current.abTestingService,
+)
+```
+
+With that we loose the ability to explisitly specify flags for A/B tests unless we separate Firebase provider into separate providers for remote configs and for A/B tests.
```
Ben-Henshall:
```suggestion
With that we loose the ability to explicitly specify flags for A/B tests unless we separate Firebase provider into separate providers for remote configs and for A/B tests.
```

Ben-Henshall (approved):

zzcgumn:
@ilyapuchka I will review this more carefully early next week, I have a train to catch this afternoon.

At a quick glance I that the suggestion to add `Optimizely` makes the case for this proposal stronger. More options and uncertainty around how feature flags and configuration data will be handled makes a stronger business case for a facade architecture.

What we might want to consider is if the detailed API needs to change a bit. `Optimizely` has boolean flags enabled/disabled together with a set of attributes for each experiment. I am thinking we may want to add support for this in the value types. I am not sure if Firebase has a similar concept. We could extend our local feature switched to include it. I have no idea if the proposed in house solution will support it.

Cookbook/Proposals/FeatureSwitchesRedesign.md
```diff
@@ -0,0 +1,218 @@
...
+
+3. All feature switches access points should be moved to the `Current` and consequentially should be only used by builders - view models, flow controllers and renderers should have them injected through their constructors.
+
+4. Root.plist should be generated based on the code that declares feature switches.
+
+## Detailed design
+
+### Module specific feature switches
+
+Each type of feature switch will be implemented with it's own type to make its semantics clear. We will need `LocalFeatureSwitch` and `RemoteFeatureSwitch` types. When later we introduce a new produt configurator service we defined new type `ProductFeatureSwitch`. They all will have a common `value` property and each provider will have a `get(valueForKey: String)` method and should store their properties in a dictionary.
```
brocoo:
```suggestion
Each type of feature switch will be implemented with its own type to make its semantics clear. We will need `LocalFeatureSwitch` and `RemoteFeatureSwitch` types. When later we introduce a new product configurator service we define a new type: `ProductFeatureSwitch`. They all will have a common `value` property and each provider will have a `get(valueForKey: String)` method and should store their properties in a dictionary.
```

Cookbook/Proposals/FeatureSwitchesRedesign.md
```diff
@@ -0,0 +1,218 @@
...
+
+
+### Point of access
+
+There is no single way of accessing feature switches in code. Local feature switches are defined in one place, Firebase feature switches are defined in another place, we also have static app configurations and consumer network feature switches (in patient details). It does not create a lot of friction but can be cumbersome to replace local feature switch with remote when feature is due to release.
+At the same time all these places (local/Firebase feature switches, app configuration) became a dumping ground for the switchies for various features of the app, even when they have their own modules. And its not uncommon to have conflicts between PRs which both introduce Firebase feature switches for different issues.
+
+### App configuration boilerplate
+
+We use app configuration type (and types related to it) to define "static" feature switches things that are hardcoded in the app and are cuased by different requirements for the products. I.e. one app can define that it should use a new feature flow while another does not use it.
```
mr-v:
```suggestion
We use app configuration type (and types related to it) to define "static" feature switches things that are hardcoded in the app and are caused by different requirements for the products. I.e. one app can define that it should use a new feature flow while another does not use it.
```

Cookbook/Proposals/FeatureSwitchesRedesign.md
```diff
@@ -0,0 +1,218 @@
...
+
+
+## Proposed solution
+
+To address these issues we suggest the following changes in our feature switches architecture:
+
+1. Feature toggles and static configurations related to the particular feature should be defined in this feature module instead of shared module or in the app. This follows interface segregation principle and goes along with improving modularisation - one won't need to change common module with feature switches and recompile all of it (and potentially everything else) when they only need to recompile module that uses this feature switch. This way we also configure our frameworks the same way as we configure any other 3rd party framework and can use the similar infrastrauctore for that (i.e. dedicated instances responsibule for configuring each feature rather than doing that all in the app delegate).
+
+2. Replace `AppConfigurationProtocol` and all related protocols with structs. This will decrease amount of boilerplate we have to write when adding new application configuration. Application configuration will be only concerned with application level configurations which are not directly related to any feature module.
```
mr-v: > Replace `AppConfigurationProtocol` and all related protocols with structs.

👍

Cookbook/Proposals/FeatureSwitchesRedesign.md
mr-v: > 4. Root.plist should be generated based on the code that declares feature switches.
 
👌

Cookbook/Proposals/FeatureSwitchesRedesign.md
```diff
@@ -0,0 +1,218 @@
...
+    let name: String
+    let key: String
+}
+```
+
+
+Using dedicated switch types we will be also able to explicitly specify what remote feature switches are A/B tested and which not:
+
+```swift
+typealias ABTest = RemoteFeatureSwitch
```
mr-v: Could you elaborate on this? I'm not sure how it would be defined and based on description I'm thinking that not all `RemoteFeatureSwitch` are A/B tests.

Cookbook/Proposals/FeatureSwitchesRedesign.md
```diff
@@ -0,0 +1,218 @@
...
+## Detailed design
+
+### Module specific feature switches
+
+Each type of feature switch will be implemented with it's own type to make its semantics clear. We will need `LocalFeatureSwitch` and `RemoteFeatureSwitch` types. When later we introduce a new produt configurator service we defined new type `ProductFeatureSwitch`. They all will have a common `value` property and each provider will have a `get(valueForKey: String)` method and should store their properties in a dictionary.
+
+```swift
+// In BabylonDependencies
+
+public struct RemoteFeatureSwitch<T> {
```
mr-v: Those `*FeatureSwitch` types don't seem much different. Would it make sense to have a single type with extra generic parameter instead? I think this keeps readability and reduces boilerplate a little bit.

```swift
struct FeatureSwitch<T, Source> {
    let name: String
    let key: String
}

enum Remote {}
enum Local {}

extension FeatureSwitch where Source == Remote {
    var value: Bool {
        return Current.abTestingService.get(valueForKey: self.key)
    }
}

extension FeatureSwitch where Source == Local {
    var value: Bool {
        return Current.userDefaults.get(valueForKey: self.key)
    }
}

let remoteFeatureSwitch = FeatureSwitch<Bool, Remote>(name: "New feature", key: "is_new_feature_enabled")
let localFeatureSwitch = FeatureSwitch<Bool, Local>(name: "Another feature", key: "is_another_feature_enabled")
```

Cookbook/Proposals/FeatureSwitchesRedesign.md
```diff
@@ -0,0 +1,218 @@
...
+    
+    public struct SomeFeatureModule {
+       
+        public let featureSwitches: FeatureSwitches
+    
+        // dynamic feature switches
+        public struct FeatureSwitches {
+            public let isNewFeatureEnabled = RemoteFeatureSwitch<Bool>(
+                name: “New feature”,
+                key: “is_new_feature_enabled”
```
mr-v: Could you also give examples of how values are set in providers, e.g. local switches? Right now we only see how `name` and `key` would be defined.

Cookbook/Proposals/FeatureSwitchesRedesign.md
mr-v: >  They all will have a common `value` property and each provider will have a `get(valueForKey: String)` method and should store their properties in a dictionary.
```swift
   var value: T {
        return provider.get(valueForKey: self.key)
    }
```

Right now we have `value: (String) -> Value` defined to get value based on key. Wouldn't we still need something like this? I can't find anything like this in proposal examples.

mr-v (commented):
Left some initial thoughts. It's a lot to take in at once, so I'll possibly take another look later.

Cookbook/Proposals/FeatureSwitchesRedesign.md
ilyapuchka: yes, I think it will be pretty much the same what we have now in ab testing services implementations, just a bit refactored maybe

Cookbook/Proposals/FeatureSwitchesRedesign.md
ilyapuchka: The "phantom" type would work as well for that purpose, but IMO declaration becomes a little bit less nice. And I'm not sure how nice it will play with property wrappers (which we can easily convert to as they have exactly the same signature), I think they allow only one generic. So then it can be `@RemoteFeatureSwitch` or `@LocalFeatureSwitch`. I don't think it's possible to do `@FeatureSwitch<Local/Remote>`.

Cookbook/Proposals/FeatureSwitchesRedesign.md
ilyapuchka: yes, but that's true the other way around, all a/b test flags are remote feature switches. 
The alternative is to implement `ABTest` alike `RemoteFeatureSwitch` which will make them not interchangeable which is also fine because implementation is trivial and we are likely to use different services for remote flags and for at tests in future.

Cookbook/Proposals/FeatureSwitchesRedesign.md
ilyapuchka: value will be retrieved from the provider via commuted property as described above in the example implementation of remote feature switch. For other types of switches it will be the same, the only difference is a different type of provider, which can be probably defined as a generic type parameter instead of injected instance as you suggested above, though I think injecting instance allows better flexibility rather than type that we will need to reconstruct or have a singleton or something else.

Cookbook/Proposals/FeatureSwitchesRedesign.md
ilyapuchka: Yes, we can use SwiftSyntax for that or just roll out our custom parser. Using SwiftGen is a nice suggestion as we will have it used in the project already anyway and will probable require less effort (only if it supports parsing custom plist resources though cc @AliSoftware). But again, editing plist is much less friendly experience than editing code.

Cookbook/Proposals/FeatureSwitchesRedesign.md
ilyapuchka: Though we can use generic type alias for this purpose:

```
typealias RemoteFeatureSwitch<T> = FeatureSwitch<T, Remote>
typealias LocalFeatureSwitch<T> = FeatureSwitch<T, Local>
```

Not sure still if it will work with property wrappers, but they are not a requirement of this redesign, so we can as well go with one generic implementation as you suggest.

Cookbook/Proposals/FeatureSwitchesRedesign.md
ilyapuchka: looks like a draft leftover, removed

ilyapuchka:
@zzcgumn Optimizely arguments will make it trickier because they mean that feature switches instead of being a simple computed property will need to be functions. We still can have this in the generic feature switch structure as a function like `func value(with arguments: [String: Any]) -> T` and make computed property `var value: T` to call it with empty attributes.

zzcgumn:
I like the suggestion to have different types for local and remote feature switches. I am wondering whether it would in the fullness of time make sense to have four types of switches

```
LocalFeatureSwitch
RemoteFeatureSwitch
ProductFeatureSwitch
ABTestingFeatureSwitch
```

They are all booleans, but serve distinct purposes.

I appreciated that we don't need to decide exactly which switch types we are going to define as part of this proposal. We can always add more switch types after moving to these type definitions.

zzcgumn (approved):
This suggestion makes a a lot of sense to me.

karabatov (approved):

Cookbook/Proposals/FeatureSwitchesRedesign.md
AliSoftware: SwiftGen would be able to generate Swift code from arbitrary Plist (we have a PlistParser now so yes you can do this even for custom plist resources).
But if we want to do the opposite and generate Plist from code, we'd need to use either Sourcery or SwiftSyntax (at that point introducing Sourcery would make more sense to open ourselves to other uses of it in other situations in the future too while SwiftSyntax would require custom coding that would only work for this specific task)

adiki (approved):

daniloaliberti (commented):
As I mentioned in the talk, the `refresh` is important to have in place when implementing it. 
For example: Telus has features based on CN and considering that CN changes when user changes their state (province in address), it should refresh. 
In depth the tabs should change as `Book Appointment` tab would be present only in a few active regions in Canada.

Cookbook/Proposals/FeatureSwitchesRedesign.md
AliSoftware: FWIW Property wrappers support multiple generics, but the problem is inference. With one single generic parameter `T` that is used by the `wrappedValue: T`, the compiler can infer `T` from the type of the variable. But if one of the generic parameter is a phantom type, not used by the wrapped value, and not inferable by a parameter passed to the PW attribute, then you have to specify the full list of generic parameters to the PW attribute.

So we could totally do this:

```swift
@FeatureSwitch<Bool, FeatureSwitchLocal>(name: "Shiny Feature", key: "shiny_feature")
var shinyFeatureEnabled: Bool
```

But it would be a shame to have to specify the full generic parameters instead of relying on inference like we could do without the phantom type, letting the compiler infer the `Bool` here:

```swift
@LocalFeatureSwitch(name: "Shiny Feature", key: "shiny_feature")
var shinyFeatureEnabled: Bool
```

So not so sure a phantom type is worth the trouble anyway indeed. And having separate types could also allow us to expose other methods on the `projectedValue` if we wanted to, like `$shinyFeatureEnabled.toggle()` or `$shinyFeatureEnabled.someAdditionalMetaDataThatWouldOnlyMakeSenseForLocalFS`

Cookbook/Proposals/FeatureSwitchesRedesign.md
```diff
@@ -0,0 +1,222 @@
...
+
+With patient details situation is a bit different because this data is associated with a `PatientDTO`. We can refactor it so that `PatientDTO` stores abstract key-value pairs and pass it to the frameworks after patient details are retrieved similarly with Firebase remote config (or similarly access it via `Current` when these patient details will be moved there). But we suggest to address this type of feature switches separately as this work depends on other work related to how the app works with patient details. In future these feature switches can become just a new type of switch like `PatientFeatureSwitch` which will use `Current.patient.details` as their provider.
+
+Inside the module we can extend `World` with a module specific namespace and keep its feature switches and configurations separate from the rest of `World` properties.
+
+```swift
+// In the feature framework
+extension Current {
+    private static private(set) var someFeature: SomeFeatureModule!
```
AliSoftware: Just realized: we can't declare stored properties on an extension though 🤔

Cookbook/Proposals/FeatureSwitchesRedesign.md
```diff
@@ -0,0 +1,222 @@
...
+
+These feature switches should be separate from static configuration (implemented ATM via `AppConfiguration`) but both should be defined and stored inside the module. This way we break up the monolythic `AppConfiguration` and allow static configuration be configured separately from dynamic feature switches which get their values at runtime.
+
+With patient details situation is a bit different because this data is associated with a `PatientDTO`. We can refactor it so that `PatientDTO` stores abstract key-value pairs and pass it to the frameworks after patient details are retrieved similarly with Firebase remote config (or similarly access it via `Current` when these patient details will be moved there). But we suggest to address this type of feature switches separately as this work depends on other work related to how the app works with patient details. In future these feature switches can become just a new type of switch like `PatientFeatureSwitch` which will use `Current.patient.details` as their provider.
```
ilyapuchka:
```suggestion
With patient details situation is a bit different because this data is associated with a `PatientDTO`. We can refactor it so that `PatientDTO` stores abstract key-value pairs and pass it to the frameworks after patient details are retrieved similarly with Firebase remote config (or similarly access it via `Current` when these patient details will be moved there). But we suggest to address this type of feature switches separately/later as this work depends on other work related to how the app works with patient details. In future these feature switches can become just a new type of switch like `PatientFeatureSwitch` which will use `Current.patient.details` as their provider and will expose `Property<Bool>` rather than just `Bool` in its interface.
```

ilyapuchka: @daniloaliberti this is actually mentioned in the doc already, just added a small detail.

ilyapuchka:
Proposal meeting outcome: no concerns were raised regarding this design except having support for asynchronous feature switches (for patient details) which will be addressed in implementation.

inamiy:
(This is slack conversation with Ilya, just for memo)

A bit of issue in the proposal though is that:

```swift
extension World {
    private static var maps: MapsUI! // NOTE: static var isn't nice
}
```

doesn’t look nice to me, and instead we simply want to create another World and Current i.e.:

```swift
// in BabylonMapsUI
struct World: FeatureModule {
    public let featureSwitches: FeatureSwitches = FeatureSwitches()
    public let config: Configuration
}
#if DEBUG
public var Current: World = .production
#else
public let Current: World = .production
#endif
```

Then in Octopus:

```swift
// in App target
BabylonDependencies.Current.setXXX(...)
BabylonMapsUI.Current.setYYY(...)  // prefixed by framework name, which is simpler than `Current.mapsUI.-` prefix.
```

This is slightly simpler and consistent across multiple frameworks, even including BabylonDependencies.

---

> We also will need to disambiguate BabylonDependencies.Current inside each vertical

PR #201, opened 2019-07-29
Add PullAssigner doc in README + mention limitation about team auto-removal

AliSoftware:
Currently on our bot repos, we have "Require approval from code owners" so the PA team isn't removed from reviewers (while it is in playbook and in main repo)

### DIFF
--- a/Cookbook/Technical-Documents/PullAssigners.md
+++ b/Cookbook/Technical-Documents/PullAssigners.md
@@ -41,7 +41,9 @@ But if you have created new teams (team listing reviewers + proxy team) you'll n
 
  - Click "Add Team"
  - Select the real (not proxy) team you've created before in the dropdown menu
- - Configure the team in the next screen, especially the number of reviewers to assign, the algorithm, but also the proxy team you previously created, and select to delete that (proxy) team review request after assigning reviewers.
+ - Configure the team in the next screen, especially the number of reviewers to assign, the algorithm, but also the proxy team you previously created, and select to delete that (proxy) team review request after assigning reviewers (†).
+
+ (†) Note that if your repository's protected branch is configured with "Require review from Code Owners" checked, then any team in your `CODEOWNERS` will stay assigned as reviewer – which means that your proxy team, being typically part of your `CODEOWNERS` itself, won't be removed from your PR even if you chose "Delete after assigning reviewer(s)" for the "Team review request" setting in your https://pullreminders.com setup. It will only be able to be deleted if the "Require review from Code Owners" setting is unchecked in your repo.
 
 ## Update GitHub's PullAssigner app settings
 
--- a/Cookbook/Technical-Documents/PullAssigners.md
+++ b/Cookbook/Technical-Documents/PullAssigners.md
@@ -43,7 +43,7 @@ But if you have created new teams (team listing reviewers + proxy team) you'll n
  - Select the real (not proxy) team you've created before in the dropdown menu
  - Configure the team in the next screen, especially the number of reviewers to assign, the algorithm, but also the proxy team you previously created, and select to delete that (proxy) team review request after assigning reviewers (†).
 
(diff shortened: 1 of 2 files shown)


Cookbook/Technical-Documents/PullAssigners.md
```diff
@@ -41,7 +41,9 @@ But if you have created new teams (team listing reviewers + proxy team) you'll n
 
  - Click "Add Team"
  - Select the real (not proxy) team you've created before in the dropdown menu
- - Configure the team in the next screen, especially the number of reviewers to assign, the algorithm, but also the proxy team you previously created, and select to delete that (proxy) team review request after assigning reviewers.
+ - Configure the team in the next screen, especially the number of reviewers to assign, the algorithm, but also the proxy team you previously created, and select to delete that (proxy) team review request after assigning reviewers (†).
+
+ (†) Note that if your repository's protected branch is configured with "Require review from Code Owners" checked, then any team in your `CODEOWNERS` will stay assigned as reviewer – which means that your proxy team, being typically part of your `CODEOWNERS` itself, won't be removed from your PR even if you chose "Delete after assigning reviewer(s)" for the "Team review request" setting in your https://pullreminders.com setup. It will only be able to be deleted if the "Require review from Code Owners" setting is unchecked in your repo.
```
zzcgumn: I find this sentence confusing. Is not `CODEOWNERS` a set of individual members?

AliSoftware: @zzcgumn No, `CODEOWNERS` can contain any type of GitHub mentions, either members or teams.

So for example our current `CODEOWNERS` (as you can see elsewhere in the same doc) is this for the main repo and the playbook:
```
* @Babylonpartners/ios-pullassigner
```
And this for the bot repos:
```
* @Babylonpartners/ios-pullassigner @Babylonpartners/ios-bot-owners
```

As you can see, it's saying "whatever part of the code this touches" (`*` matcher), the owners are `@Babylonpartners/iOS-PullAssigner` and `@Babylonpartners/ios-bot-owners` – which happen to be GitHub teams, not individual members.

You could totally have individual members in your `CODEOWNERS` files as well, if you want one specific person to be specified as the owner of (a part of) the code. But teams work as well, and that's actually what we use.


The main different with using teams in `CODEOWNERS` like we do here, is that when GitHub requests a review from a team T, even if this team T contains 5 members (A,B,C,D,E), it will be happy once any one person in that team T approves the PR. As opposed to requiring an approval of all of the 5 members A,B,C,D,E if you mentioned those members explicitly on `CODEOWNERS`.

AliSoftware:
```suggestion
 (†) Note that if your repository's protected branch is configured with "Require review from Code Owners" checked, then any member or team mentioned in your `CODEOWNERS` file will stay assigned as reviewer. Which means that in that case, your proxy team, being typically mentioned in your `CODEOWNERS` to always be auto-assigned, won't be removed from your PR even if you chose "Delete after assigning reviewer(s)" for the "Team review request" setting in your https://pullreminders.com setup. It will only be able to be deleted if the "Require review from Code Owners" setting is unchecked in your repo.
```
@zzcgumn better?

ilyapuchka: what's the problem with this team not being removed from list of reviewers?

ilyapuchka (approved):

Cookbook/Technical-Documents/PullAssigners.md
AliSoftware: I'm not sure there's a problem with it not being removed tbh. It's just that it was always configured to "Remove team after assigning reviewers" in our pullreminders.com settings, but didn't actually remove it in bot repos (while it did in babylon and playbook), so we didn't _understand_ why it didn't work while it was configured to.

But in fact I don't think it's a problem if it doesn't remove the team.
Probably the next iteration of this document once we migrate all our repos to that setup would instead just tell to NOT configure pullreminders.com settings to "Delete once it assigned reviewers" anymore (since it won't work if we have "Require Code Owners approval" checked anyway). We'll see how that goes once we reconfigured our main repos (but since main repos will also soon probably use code owners per vertical / squad, we'll re-update the doc at that point when we know more what we'll decide about it)

gtsiap (approved):

zzcgumn (commented):
Looks good to me, thanks for clarifying the sentence around hoe code owner membership works.

PR #213, opened 2019-08-12
Bento-as-SwiftUI manifesto

andersio:


andersio:
Let's just get the parts about modifiers in for now. We would then continue on component definition & screen composition.

sergdort (approved):

Cookbook/Technical-Documents/Bento-SwiftUI-manifesto.md
```diff
@@ -0,0 +1,42 @@
+# Bento-as-SwiftUI manifesto
+## Preface
+As part of our commitment to eventually migrate to SwiftUI, it was made clear that our infrastructure needs to be gradually made aligned with SwiftUI to both (1) prepare developers for SwiftUI paradigms and (2) reduce the cost and impact of the migration with source compatibility at best effort.
+This document outlines the planned changes to Bento.
+## Component composition
+SwiftUI view composites are encoded into the type system, say (in pseudo code) a modifier view is expressed as `Modified<Original, U>` and the modified version of it being `Modified<Modified<Original, U>, T>`, similar to lazy collections in the Standard Library. Since different compositions always result in a unique type, it can take advantage of the implied non-substitutibility to reuse view hierarchies and reconstruct them /only/ when the component type differs.
```
j531:
```suggestion
SwiftUI view composites are encoded into the type system, say (in pseudo code) a modifier view is expressed as `Modified<Original, U>` and the modified version of it being `Modified<Modified<Original, U>, T>`, similar to lazy collections in the Standard Library. Since different compositions always result in a unique type, it can take advantage of the implied non-substitutibility to reuse view hierarchies and reconstruct them _only_ when the component type differs.
```

Cookbook/Technical-Documents/Bento-SwiftUI-manifesto.md
```diff
@@ -0,0 +1,42 @@
...
+	// Before
+	public func width(_ value: CGFloat) -> AnyRenderable
+
+	// After
+	public func width(_ value: CGFloat) -> Self._ConstraintModified<ExactWidth>
+}
+```
+The unfortunate side effect is that spelling the type out in code might become an annoyance, as we cannot use opaque result types while still having to support iOS 12. Note that type erasure with `AnyRenderable` still remains an option.
+Here is a list of existing operators that would be affected by this change:
+| Group | Operators |
```
j531: This table doesn't render in the preview. Might be different when live?

j531 (approved):

PR #222, opened 2019-08-27
Proposal: Split Octopus

gtsiap:

### DIFF
--- a/Cookbook/Proposals/SplitOctopus.md
+++ b/Cookbook/Proposals/SplitOctopus.md
@@ -5,7 +5,7 @@
 
 ## Introduction
 
-From the early versions of the codebase Octopus was the place were  features were added.
+From the early versions of the codebase Octopus was the place were features were added.
 Unfortunately Octopus hasn't adapted as the rest of the codebase has.
 In the iOS codebase we use frameworks in order to organize the code and Octopus isn't one.
 
@@ -25,6 +25,16 @@ The first framework will be used as a container for small features and it should
 The second framework will contain anything else which was in Octopus.
 This framework can depend on any other framework.
 
+Converting Octopus into two frameworks
+- should improve the build time of the app
+- will reduce the time which is required for unit tests to be executed
+- will allow better code organization
+- engineers won't be limited in any way
+
+In the initial split `BabylonFeatures` will host all the features which satisfy the dependencies requirement.
+After the initial split it will be up to each squad to decide the appropriate place for their code.
+
+By having these two frameworks we will improve the developer experience and provide the same level of flexibility as before. 
 
 ## Impact on existing codebase
 
--- a/Cookbook/Proposals/SplitOctopus.md
+++ b/Cookbook/Proposals/SplitOctopus.md
@@ -51,6 +51,10 @@ Option 2 will be easier for the team to review the changes but it will require b
 
 Personally I prefer option 2.
 
+## Octopus Analysis
+
+Octopus analysis can be found [here.](https://docs.google.com/spreadsheets/d/1nu7L6vwQcET1fYas1Ccj6vlqHJRtsdaXQ94LaXjvXkQ/edit#gid=0)
+
 ## Impact on existing codebase
 
 Although this is a breaking change, most likely it won't affect the enginners much.
--- a/Cookbook/Proposals/SplitOctopus.md
+++ b/Cookbook/Proposals/SplitOctopus.md
@@ -1,7 +1,7 @@
 # Split Octopus codebase
 
 * Author(s): Giorgos Tsiapaliokas
-* Review Manager: TBD
+* Review Managers: David Rodrigues, Yasuhiro Inami, Olivier Halligon
 
 ## Introduction
 
--- a/Cookbook/Proposals/SplitOctopus.md
+++ b/Cookbook/Proposals/SplitOctopus.md
@@ -11,50 +11,21 @@ In the iOS codebase we use frameworks in order to organize the code and Octopus 
 
 ## Motivation
 
-While there is still the need for having a place for small features Octopus has become a massive container which contains both features and services.
-Also from a tech POV Octopus is just a folder which is being reused in different targets.
-I am sure we can do better than this :)
+In it's current state Octopus is collection of services and features which is consumed
+by each target this requires some manual work for each target.
 
 ## Proposed solution
 
-Split Octopus into two different frameworks.
-The proposed names are `BabylonFeatures` and `BabylonApp`.
-
-The first framework will be used as a container for features which
-- depend only on `BabylonCore`, `BabylonUI`, `BabylonDepedencies` and `BabylonSDK`
-- don't depend on any third party SDKs
-
-The second framework will be more flexibile with less restrictions, features in this framework
-- can depend on other Babylon framework
-- can depend on any third party SDK
-
-In the initial split Octopus will be splitted in the new two frameworks and then it Octopus will be removed from the codebase.
-After the initial split it will be up to each squad to decide the appropriate place for their code.
-
-Converting Octopus into two frameworks
-- should improve the build time of the app
-- will reduce the time which is required for unit tests to be executed
-- will allow better code organization
-- engineers won't be limited in any way, since two frameworks will exist in order to host features depending on their dependencies
+We propose to convert Octopus into a framework with the name `BabylonOctopus`.
+By having Octopus as a framework it will be easier for each target to make use of Octopus' features and services.
+With this proposal there is still the need for splitting Octopus even further which will be addressed by other proposals
+in the future.
 
 ### Transition Period
 
-Splitting Octopus will require some time since it's not trivial.
-There are two approaches 
-
-1. Split everything at once
-2. Initially create empty frameworks and then move each subdirectory of Octopus piece by piece.
-
-Option 1 has the drawback of keeping everything in a large PR and updating this PR constantly.
-Also IMO having such a massive PR will make it more error prone.
-
-Option 2 will be easier for the team to review the changes but it will require better coordination in order to make sure that the two new frameworks will be used instead of Octopus.
-
-Personally I prefer option 2.
-
-## Octopus Analysis
-
-Octopus analysis can be found [here.](https://docs.google.com/spreadsheets/d/1nu7L6vwQcET1fYas1Ccj6vlqHJRtsdaXQ94LaXjvXkQ/edit#gid=0)
+Converting Octopus will require some time since it's not trivial.
+We propose to  initially create an empty framework and then move each subdirectory of Octopus piece by piece.
+This way it will be easier for the team to review the changes.
 
 ## Impact on existing codebase
 
@@ -62,4 +33,8 @@ Although this is a breaking change, most likely it won't affect the enginners mu
 
 ## Alternatives considered
 
-Leave it as is.
+1. Leave it as is.
+2. Convert Octopus to framework at once.
+The drawback of this alternative is that the team will have to review a large PR which
+is more error prone. Also we will need to maintain a branch with a lot of changes for a
+longer period of time.
--- a/Cookbook/Proposals/SplitOctopus.md
+++ b/Cookbook/Proposals/SplitOctopus.md
@@ -1,0 +1,35 @@
+# Split Octopus codebase
+
+* Author(s): Giorgos Tsiapaliokas
+* Review Manager: TBD
+
+## Introduction
+
+From the early versions of the codebase Octopus was the place were  features were added.
+Unfortunately Octopus hasn't adapted as the rest of the codebase has.
+In the iOS codebase we use frameworks in order to organize the code and Octopus isn't one.
+
+## Motivation
+
+While there is still the need for having a place for small features Octopus has become a massive container which contains both features and services.
+Also from a tech POV Octopus is just a folder which is being reused in different targets.
+I am sure we can do better than this :)
+
+## Proposed solution
+
+Split Octopus into two different frameworks.
+The proposed names are `BabylonFeatures` and `BabylonApp`.
+
+The first framework will be used as a container for small features and it should depend only on `BabylonCore`, `BabylonUI`, `BabylonDepedencies` and `BabylonSDK`.
+
+The second framework will contain anything else which was in Octopus.
+This framework can depend on any other framework.
+
+
+## Impact on existing codebase
+
+Although this is a breaking change, most likely it won't affect the enginners much.
+
+## Alternatives considered
+
+Leave it as is.
--- a/Cookbook/Proposals/SplitOctopus.md
+++ b/Cookbook/Proposals/SplitOctopus.md
@@ -20,21 +20,22 @@ I am sure we can do better than this :)
 Split Octopus into two different frameworks.
 The proposed names are `BabylonFeatures` and `BabylonApp`.
 
-The first framework will be used as a container for small features and it should depend only on `BabylonCore`, `BabylonUI`, `BabylonDepedencies` and `BabylonSDK`.
+The first framework will be used as a container for features which
+- depend only on `BabylonCore`, `BabylonUI`, `BabylonDepedencies` and `BabylonSDK`
+- don't depend on any third party SDKs
 
-The second framework will contain anything else which was in Octopus.
-This framework can depend on any other framework.
+The second framework will be more flexibile with less restrictions, features in this framework
+- can depend on other Babylon framework
+- can depend on any third party SDK
+
+In the initial split Octopus will be splitted in the new two frameworks and then it Octopus will be removed from the codebase.
+After the initial split it will be up to each squad to decide the appropriate place for their code.
 
 Converting Octopus into two frameworks
 - should improve the build time of the app
 - will reduce the time which is required for unit tests to be executed
 - will allow better code organization
-- engineers won't be limited in any way
-
-In the initial split `BabylonFeatures` will host all the features which satisfy the dependencies requirement.
-After the initial split it will be up to each squad to decide the appropriate place for their code.
-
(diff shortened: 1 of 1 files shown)


inamiy (changes requested):
This will be related to `CNSMR-2646 iOS Modularisation` epic, and I believe this PR will be a good starter proposal.

However at the moment, documentation doesn't provide enough discussion of which subdirectories to actually move into `BabylonFeatures`, so @gtsiap you might want to add more detailed actions.

Cookbook/Proposals/SplitOctopus.md
```diff
@@ -0,0 +1,45 @@
...
+While there is still the need for having a place for small features Octopus has become a massive container which contains both features and services.
+Also from a tech POV Octopus is just a folder which is being reused in different targets.
+I am sure we can do better than this :)
+
+## Proposed solution
+
+Split Octopus into two different frameworks.
+The proposed names are `BabylonFeatures` and `BabylonApp`.
+
+The first framework will be used as a container for small features and it should depend only on `BabylonCore`, `BabylonUI`, `BabylonDepedencies` and `BabylonSDK`.
```
ilyapuchka: can you add some examples?

Cookbook/Proposals/SplitOctopus.md
```diff
@@ -0,0 +1,45 @@
...
+
+The second framework will contain anything else which was in Octopus.
```
ilyapuchka: ditto

Cookbook/Proposals/SplitOctopus.md
```diff
@@ -0,0 +1,45 @@
...
+This framework can depend on any other framework.
+
+Converting Octopus into two frameworks
+- should improve the build time of the app
+- will reduce the time which is required for unit tests to be executed
+- will allow better code organization
+- engineers won't be limited in any way
```
ilyapuchka: how they are limited now and how exactly splitting into two frameworks solves it?

Cookbook/Proposals/SplitOctopus.md
```diff
@@ -0,0 +1,45 @@
...
+
+In the initial split `BabylonFeatures` will host all the features which satisfy the dependencies requirement.
```
ilyapuchka: > which satisfy the dependencies requirement

can we rephrase it? what requirements?

Cookbook/Proposals/SplitOctopus.md
```diff
@@ -0,0 +1,45 @@
...
+After the initial split it will be up to each squad to decide the appropriate place for their code.
```
ilyapuchka: any guidance on how this should be decided?

Cookbook/Proposals/SplitOctopus.md
```diff
@@ -0,0 +1,45 @@
...
+
+By having these two frameworks we will improve the developer experience and provide the same level of flexibility as before.
```
ilyapuchka: Can you describe how exactly?

Cookbook/Proposals/SplitOctopus.md
```diff
@@ -0,0 +1,45 @@
...
+## Motivation
+
+While there is still the need for having a place for small features Octopus has become a massive container which contains both features and services.
+Also from a tech POV Octopus is just a folder which is being reused in different targets.
+I am sure we can do better than this :)
+
+## Proposed solution
+
+Split Octopus into two different frameworks.
+The proposed names are `BabylonFeatures` and `BabylonApp`.
```
ilyapuchka: We already have a `Babylon` target for the app, which `Octopus` folder is part of (as well as all other app targets)  is it the same?

ilyapuchka (changes requested):
It would be nice to see some at least minimum diagram of dependencies before/after (at least at the high level). We have such diagram already here https://github.com/Babylonpartners/ios-playbook/blob/master/Cookbook/Proposals/ControlTheWorld.md#where-will-current-live

Cookbook/Proposals/SplitOctopus.md
AliSoftware: I'm not clear how you'd want to organise that split. I think it warrants more details

 - Which current subfolders of Octopus will go in which framework?
 - "BabylonFeatures will be use as a container for small features"
   - How do you define big vs small feature?
   - Would that mean that every feature and screen and renderer and builder would move into that framework? I guess not, so how would we decide which feature should go in one of our other Babylon* modules vs inside BabylonFeatures?
 - "BabylonApp will contain anything else" sounds like we plan to keep that framework (however we plan to name it) as a "dumping ground for everything we don't know where to put"?
    - Is that temporary (just the time to do a first pass extraction and triage the remaining code better later) or a plan for the long term?
    - In any case, not sure that the name `BabylonApp` is very representative here, could you give more examples of what you'd imagine would end up in that second framework?
 - How does that work with other flavors (Bupa, Telus, Ascension, ...)
    - question actually applies for both frameworks described here. Will BabylonFeatures contain only Babylon UK features or could it also end up containing Telus ones? Same for BabylonApp

AliSoftware (changes requested):
Nice first draft but I don't think it's detailed enough. It's not clear to me what is currently in Octopus that would go in which of the new frameworks

Also, in addition to better describe how things will distribute, we need to:

* Describe the transition period: will we migrate everything at once, or create the (almost empty frameworks) then migrate parts little by little, and will we end up completely deleting the `Octopus` folder once everything is done or will _some_ rare stuff still stay in the main app target?
* Describe the impacts and anticipated difficulties of the transition if possible. This goes alongside with a better description of which subfolder of Octopus goes into which framework actually, as it will give us an idea of how tangled each feature in Octopus is with other frameworks but also with other features. So are there any features already identified that we know would be trickier to untangle while extracting, or is each subfolder of Octopus already independent from the other Octopus folders and easy to extract?

Cookbook/Proposals/SplitOctopus.md
```diff
@@ -0,0 +1,65 @@
...
```
sergdort: Can I propose to make ability to make other apps more independent?
To the extend that we would just have something like

```swift
final class AppDelegate {
   let app: App
   init() {
     app = BabylonApp()
   }

open func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey : Any]? = nil) -> Bool {
    app.run()
}
}
```

Current approach of feature switches is not perfect as there may be some features that are only available on specific app

gtsiap: I see the benefits of having a new `AppDelegate` for each app.

This would remove some of the configuration values/tokens which are needed, but feature switches would still be required. Right?

ilyapuchka: @sergdort how this is different than having its won app delegate implemented in each app? we still have common tasks to perform for all the apps

Cookbook/Proposals/SplitOctopus.md
```diff
@@ -0,0 +1,65 @@
...
+Option 1 has the drawback of keeping everything in a large PR and updating this PR constantly.
+Also IMO having such a massive PR will make it more error prone.
+
+Option 2 will be easier for the team to review the changes but it will require better coordination in order to make sure that the two new frameworks will be used instead of Octopus.
+
+Personally I prefer option 2.
+
+## Octopus Analysis
+
+Octopus analysis can be found [here.](https://docs.google.com/spreadsheets/d/1nu7L6vwQcET1fYas1Ccj6vlqHJRtsdaXQ94LaXjvXkQ/edit#gid=0)
```
ilyapuchka: no permissions. can you please move it to shared drive?

Cookbook/Proposals/SplitOctopus.md
```diff
@@ -0,0 +1,65 @@
...
+While there is still the need for having a place for small features Octopus has become a massive container which contains both features and services.
+Also from a tech POV Octopus is just a folder which is being reused in different targets.
+I am sure we can do better than this :)
+
+## Proposed solution
+
+Split Octopus into two different frameworks.
+The proposed names are `BabylonFeatures` and `BabylonApp`.
+
+The first framework will be used as a container for features which
```
ilyapuchka: any examples?

Cookbook/Proposals/SplitOctopus.md
sergdort: The idea I head in mind is really separate app for each client so if they would have their own Flow they can implement them freely without worrying that it gonna break other apps. 
Essentially what other apps would use are common Frameworks + Their features. So we would have a separate project without adding files to the multiple targets like we do now

ilyapuchka: Yes, that would be good. But we still will need to have common things to perform for each app as they are configured identically using the environment values. So it will be like extracting shared app delegate code to the framework but then we don't solve the problem of code sharing, or having this shared code duplicated which also have tradeoffs. So probably we would still have some kind of BaseApp 🤷‍♂ 
If what you are talking about is for example how these apps will configure let's say their tab bar, then I guess the architecture we could have is to have a tab bar provided by a framework and provide an interface for apps to configure it, rather than how we have now using feature switches in the tab bar. This way the configuration point will be in these app instances. That also intersects with feature switches proposal then.

Cookbook/Proposals/SplitOctopus.md
```diff
@@ -0,0 +1,65 @@
...
+Splitting Octopus will require some time since it's not trivial.
+There are two approaches 
+
+1. Split everything at once
+2. Initially create empty frameworks and then move each subdirectory of Octopus piece by piece.
+
+Option 1 has the drawback of keeping everything in a large PR and updating this PR constantly.
+Also IMO having such a massive PR will make it more error prone.
+
+Option 2 will be easier for the team to review the changes but it will require better coordination in order to make sure that the two new frameworks will be used instead of Octopus.
```
AliSoftware: I agree with Option 2. I think you could reword this paragraph to only mention this Option 2 for "Transition Period" (be affirmative in the proposal, don't leave questions), and move option 1 to "Alternatives Considered"

Cookbook/Proposals/SplitOctopus.md
AliSoftware: same.

Cookbook/Proposals/SplitOctopus.md
AliSoftware: I still don't understand what `BabylonApp` framework will be. That naming seems too generic.
(Maybe the Google Document linked below explains that better, but since I don't have access and can't view it, all this seems pretty vague to me)

AliSoftware (commented):
tbh I still feel that the proposal is not very detailed at which parts of Octopus will go where. Maybe the spreadsheet is supposed to shed some light on this but I can't view it.

inamiy:
After discussion with Platform team, we probably want to: 

1. **Start by focusing on only 1 new framework** to decouple from current multi-app targets

In current proposal's case, it will be `BabylonApp`, but I also find this name confusing, so we probably want to keep the naming as is, i.e. `BabylonOctopus`.

In this couter-proposal, `BabylonOctopus` will be a **temporal dumping ground** that migrates most of `Octopus` directory to the new framework, while some subdirectories may require decouple-refactoring before migration.
This migration will be done **one by one**, so we can evaluate how many dependencies are actually there, and once we could migrate, that means there is a higher chance of making it as an another smaller micro framework, e.g. `BabylonFeatures`, `BabylonPayments`.

But during this practice, we don't separate to `BabylonFeatures` yet because of Rule 1 and because we probably need a discussion for how much micro-management we really want to do, so in any micro-frameworking process it will probably go to each proposal first.

### Example

At this point, I find some candidates to make micro-framework:

- Notifications / RoutingEvent
- Payments
- Analytics (core + event definitions per each app)
- Each screen flow

But when applying Rule 1, **all of them will first go to `BabylonOctopus`** , then decide if we should move to micro-framework or not.

Cookbook/Proposals/SplitOctopus.md
```diff
@@ -0,0 +1,65 @@
...
+I am sure we can do better than this :)
+
+## Proposed solution
+
+Split Octopus into two different frameworks.
+The proposed names are `BabylonFeatures` and `BabylonApp`.
+
+The first framework will be used as a container for features which
+- depend only on `BabylonCore`, `BabylonUI`, `BabylonDepedencies` and `BabylonSDK`
+- don't depend on any third party SDKs
```
ilyapuchka: What about first party dependencies, like StoreKit, or other other frameworks? This way most of the staff will still live in the "app" framework where the name of it suggests that it should contain only stuff related to the "app level" code, like app delegate related code and app lifecycle, or things like tab bar and MainUserSession. How depending on StoreKit for instance justifies keeping payments functionality in this "app" framework when it is clearly a "feature" that should be actually consumed by other frameworks, i.e. test and kits should be able to use payments functionality (it does not have to depend on the payments framework for that though, but it could), but test and kits are moving to features but payments stay at the app level. What is the dependency diagram then?

Imo we need to discuss a long term plan for the split, even if we are doing it in stages like first splitting into smaller number of frameworks and them split these frameworks in smaller pieces.

Cookbook/Proposals/SplitOctopus.md
```diff
@@ -0,0 +1,40 @@
...
+We propose to convert Octopus into a framework with the name `BabylonOctopus`.
+By having Octopus as a framework it will be easier for each target to make use of Octopus' features and services.
+With this proposal there is still the need for splitting Octopus even further which will be addressed by other proposals
+in the future.
+
+### Transition Period
+
+Converting Octopus will require some time since it's not trivial.
+We propose to  initially create an empty framework and then move each subdirectory of Octopus piece by piece.
+This way it will be easier for the team to review the changes.
```
AliSoftware: If we choose to only have one framework at first (and keep the splitting of that framework into smaller pieces for a later proposal as you suggest), then it would be nice to at least ensure that each piece of code moved into the new framework is independent of each other. There's no infallible way to do that and make the compiler ensure that for us if we keep a single framework though (unlike if we create multiple independent frameworks right away), but we could still mention this "keep the various features of Octopus independent from each other" as a hard goal to keep in mind here during those step-by-step migration.

Also we could mention that we could try to organise the `BabylonOctopus` framework in folders that would likely correspond to independent features so that when we decide to later split that framework into smaller ones in a future proposal, we already have pieces of code mainly grouped appropriately and ready to be split. i.e. that proposal could suggest an initial list of groups and how we could organise that framework in independent folders – even if that split isn't final and makes no guarantee and will evolve as we move the code piece by piece into that framework and realise we might need to tweak this folder hierarchy as we progress on that migration, at least that would be a good guideline to start with.

Cookbook/Proposals/SplitOctopus.md
```diff
@@ -0,0 +1,40 @@
...
+In the iOS codebase we use frameworks in order to organize the code and Octopus isn't one.
+
+## Motivation
+
+In it's current state Octopus is collection of services and features which is consumed
+by each target this requires some manual work for each target.
+
+## Proposed solution
+
+We propose to convert Octopus into a framework with the name `BabylonOctopus`.
```
zzcgumn: Should we not call it `BabylonAppServices` instead of `BabylonOctopus`?

The original motivation for the name, because it has no back bone, is funny but not very relevant anymore.

Cookbook/Proposals/SplitOctopus.md
zzcgumn: I am not convinced that we will be able to move all subdirectories from the main target to a new framework. I don't think it will be feasible to move all UI code.

Also, GP @ Hand related code should be moved to `BabylonNHS`. It might make sense to start with this refactoring.

AliSoftware: Yeah that's also why I think it would be beneficial to already have a view of what will go where (in the new `BabylonOctopus/BabylonAppServices` framework and if so in which subfolder? Or in another framework like `BabylonNHS`?), that work of moving the code from the main target should be the occasion to also reorganise how the files are split in the main target today to dispatch them in what might end up being a different folder hierarchy and organisation.

My point (and I think yours too) being that since it will likely not be a 1-to-1 exact match from "folder X of main target goes into same folder X of new framework", but instead will likely be "some files of folder X of main targets goes into folder Y of new framework, some other files of folder X will go into a separate folder Z of new framework, and other files will move to BabylonNHS".

And I think we need to make the exercise and agree to such a mapping (even if it's not a definitive one, at least to have a rough idea) before approving the proposal, both to understand if such a mapping makes sense and if keeping the proposed solution (only create one framework now and split later) is the best choice or not, but also to have a sense of the amount of work that will be needed to do that migration and split, in order to vote if the proposed plan makes sense

AliSoftware: I wonder if it would be possible to declare a custom `module.modulemap` file that would automatically create one submodules per top-level folder we'd put in that Octopus framework.
That way we'd still only have one single `BabylonOctopus` target and one framework, but the content of each folder would be forced to be independent (and we could even require to explicitly `import BabylonOctopus.FeatureA` to only import the features/swift files that are in the `FeatureA` folder), that would offer us the guarantee that each folder would have to be independent – Xcode would tell us if we need to declare some explicit dependencies between those folders – without having to create separate frameworks right away, and with the advantage that if we see a file that needs to be moved from one folder to another while we work on this refactoring, it's just a matter of moving a file between folder (and let the compiler tell us if that would accidentally create unwanted dependencies between those submodules), without having to commit to a particular fixed and frozen folder organisation but letting us experiment with various splits as we advance on the piece-by-piece move...

Not even sure if that's totally possible – we'd need to look at blog posts and documentation about `module.modulemap` to be sure – but might be worth looking into the idea?

inamiy:
### Note

Final outcome of this proposal was to not replace all Octopus code into a new framework at once, but extract small pieces by pieces into a new framework done by each responsible squad.

PR #228, opened 2019-09-02
[Proposal] - Base class for FeedbackLoops systems.md

sergdort:


Cookbook/Proposals/Base class for FeedbackLoops systems.md
```diff
@@ -0,0 +1,157 @@
+# Base class for FeedbackLoops systems
+
+* Author(s): Sergey Shulga
+* Review Manager: Anders Ha
+
+## Introduction
+
+Almost for all our screens, we implement ViewModels which are powered by the [ReactiveFeeedback](https://github.com/Babylonpartners/ReactiveFeedback).
+
+The current interface of VM looks like this
```
ilyapuchka: so how will it change? Will `state.mutate` be the only change with this approach or the way we initialise state will also change? I suppose it will because the renderer signature changes now.

ilyapuchka (commented):
I'm +1 for this proposal, great idea. Though I'd like to see what other impact it will make on the view models, as I expect it to change state initialisation.

Cookbook/Proposals/Base class for FeedbackLoops systems.md
```diff
@@ -0,0 +1,157 @@
...
+1. We need to implement `input` feedback in pretty much every VM which may lead to errors if do not schedule input correctly, for this we have `Feedback.input()` convenience extension
+
+2. If the screen contains a lot of input field we end up having a lot of event's that are just mutate one property of the `State` e.g
+`Action.didChangeUserName(String)` `Action.didPickBirthDate(Date)`
+ 
+## Proposed solution
+1. Implementation of the feedback:
+
+```swift
+open class FeedbackLoop<State, Event, Action>: FeedbackLoopType {
```
ilyapuchka: can you add example of how the subclass of this will look like and how it will be used in a some view model?

Cookbook/Proposals/Base class for FeedbackLoops systems.md
```diff
@@ -0,0 +1,157 @@
...
+  open func mutate<V>(keyPath: WritableKeyPath<State, V>, value: V) {
+    self.input.observer(.mutation(Mutation(keyPath: keyPath, value: value)))
+  }
+
+  private enum Update {
+    case event(Event)
+    case mutation(Mutation)
+  }
+
+  private struct Mutation {
```
ilyapuchka: this will probably benefit a little bit from @dynamicCallable and this new proposal https://github.com/apple/swift-evolution/blob/master/proposals/0253-callable.md This way instead of `mutation.mutate(state)` we can just have `mutation(state)`.

mluisbrown:
My main issue with this approach is that it means that all state mutations would no longer all be handled explicitly in the reducer. Although the `.ui(.didChangeEmail(String))` type events can become a bit verbose, it is very clear how they change the state and they are all explicitly enumerated in the reducer.

Now you have mutations that are sprinkled all over the renderer and so are harder to find. My initial feeling would be to be against this change 😞

Cookbook/Proposals/Base class for FeedbackLoops systems.md
```diff
@@ -0,0 +1,157 @@
...
+      scheduler: scheduler,
+      reduce: { (state: State, update: Update) -> State in
+        switch update {
+        case let .event(event):
+          return reducer(state, event)
+        case let .mutation(mutation):
+          return mutation.mutate(state)
+        }
+      },
+      feedbacks: feedbacks.map { $0.mapEvent(Update.event) }
```
ilyapuchka: what is this `mapEvent`?

mluisbrown: See here: https://github.com/sergdort/CombineFeedback/blob/master/Sources/CombineFeedbackUI/ViewModel.swift#L62-L66

Cookbook/Proposals/Base class for FeedbackLoops systems.md
```diff
@@ -0,0 +1,157 @@
+# Base class for FeedbackLoops systems
+
+* Author(s): Sergey Shulga
+* Review Manager: Anders Ha
+
+## Introduction
+
+Almost for all our screens, we implement ViewModels which are powered by the [ReactiveFeeedback](https://github.com/Babylonpartners/ReactiveFeedback).
```
NSMyself:
```suggestion
Almost for all our screens, we implement ViewModels which are powered by [ReactiveFeeedback](https://github.com/Babylonpartners/ReactiveFeedback).
```

Cookbook/Proposals/Base class for FeedbackLoops systems.md
```diff
@@ -0,0 +1,157 @@
...
+      initial: initial,
+      scheduler: scheduler,
+      reduce: { (state: State, update: Update) -> State in
+        switch update {
+        case let .event(event):
+          return reducer(state, event)
+        case let .mutation(mutation):
+          return mutation.mutate(state)
+        }
+      },
```
RuiAAPeres: This feels like a higher order reducer.

```
func mutationReducer<State, Event>(_ reducer: (State, Event) -> State) -> (State, Event) -> State {
    return { state, event in 
       switch event {
              case let .event(anEvent):
                     return reducer(state, anEvent)
              case let .mutation(mutation):
                     return mutation.mutate(state)
              }
    }
}
```

And then:

```swift
self.state = Property(
	      initial: initial,
	      scheduler: scheduler,
	      reduce: mutationReducer(reducer),
              feedbacks: feedbacks.map { $0.mapEvent(Update.event) } 
```

Cookbook/Proposals/Base class for FeedbackLoops systems.md
```diff
@@ -0,0 +1,157 @@
...
+}
+```
+
+## Motivation
+
+One of the main issues with this approach
+
+1. We need to implement `input` feedback in pretty much every VM which may lead to errors if do not schedule input correctly, for this we have `Feedback.input()` convenience extension
+
+2. If the screen contains a lot of input field we end up having a lot of event's that are just mutate one property of the `State` e.g
```
NSMyself:
```suggestion
2. If the screen contains a lot of input fields we end up having a lot of events that just mutate one property of the `State` e.g
```

Cookbook/Proposals/Base class for FeedbackLoops systems.md
```diff
@@ -0,0 +1,157 @@
...
+
+  func render(snapshot: StateSnapshot<FeedbackLoop>) -> AnyRenderable
+}
+```
+
+This will allow us easier bind VM to a component e.g
+
+```swift
+return Component(textDidChange: {
+   //This will still couse rerendering but we would not need to create an Event for it
```
NSMyself:
```suggestion
   // this will still cause rerendering but we would not need to create an Event for it
```

Cookbook/Proposals/Base class for FeedbackLoops systems.md
```diff
@@ -0,0 +1,157 @@
...
+    snapshot.mutate(\.userName, value: $0)
+})
+```
+
+
+## Impact on the existing codebase
+
+This is additive no impact can be bundled together with new `RenderableViewController`
```
NSMyself: Can we rephrase this? Sounds confusing imho

Cookbook/Proposals/Base class for FeedbackLoops systems.md
inamiy: Yes, it's a lift from `Event` to `Update`:

```swift
func mutationReducer<State, Event>(_ reducer: (State, Event) -> State) 
    -> (State, Update) -> State
```

But what `(State, Update) -> State` basically means is that we will have a **direct way to mutating sub-state at anytime**, which is against the concept of (well-designed) state machine that has the minimal state transition arrows.
If we allow `Update` to be passed in, it means we are going to draw arbitrary arrows between two states, which will be just like **mutable state `class` with `var`s**.

RuiAAPeres: Ah yes, I missed the state lifting from the reducer passed in the init, to the reducer passed to the state machine.

Cookbook/Proposals/Base class for FeedbackLoops systems.md
sergdort: I guess in terms of point free it's a form of pull back 🤔

Cookbook/Proposals/Base class for FeedbackLoops systems.md
sergdort:
```
init(...) {
super.init(
    initial: State(),
    reducer: VM.reduce,
    scheduler: scheduler,
    feedback: [feedbacks]
)
}
```

Cookbook/Proposals/Base class for FeedbackLoops systems.md
mluisbrown: > If we allow `Update` to be passed in, it means we are going to draw arbitrary arrows between two states, which will be just like **mutable state `class` with `var`s**.

That's exactly my concern too.

Cookbook/Proposals/Base class for FeedbackLoops systems.md
sergdort: We have a `RenderableViewController` which is suppose to replace `BoxViewController`

Cookbook/Proposals/Base class for FeedbackLoops systems.md
sergdort: > mutable state class with vars

That's not entirely correct, because mutation will emit new State

sergdort: Well I guess it's a trade off of dropping some actions that do not give much value. To some extent we just bind a component to particular sub state if only thing it does it updating its value ¯\_(ツ)_/¯

sergdort: But if the emitting event does something else e.g triggering another feedback or does some validation of input than we still are going to do it via `Action.someChange(String)`.

Cookbook/Proposals/Base class for FeedbackLoops systems.md
sergdort: Could be very likely that we won't have `Renderer` any more

Cookbook/Proposals/Base class for FeedbackLoops systems.md
RuiAAPeres: @sergdort is it the goal to update the state via the KeyPath conveniences?

Cookbook/Proposals/Base class for FeedbackLoops systems.md
ilyapuchka: So the view model will be a subclass of `FeedbackLoop`?

Cookbook/Proposals/Base class for FeedbackLoops systems.md
RuiAAPeres: A form of pull back would be on the state (versus the event):

```
func mapEvent<U>(_ f: (Event -> U) -> Feedback<State, U> {

}
```

versus

```
func contraMapEvent<U>(_ f: (State -> U) -> Feedback<U, Event> {

}
```

Cookbook/Proposals/Base class for FeedbackLoops systems.md
ilyapuchka: can we address @mluisbrown and @inamiy concern by leaving handling of mutation in each view model rather than doing that in the base class? essentially mutation is a pair of key path and value, so why not to bundle it in one generic even? this seems to be the main idea of the suggestion, isn't it?

ilyapuchka: It also will hurt debugging if handling of all mutations is done in one central place, wouldn't it?

Cookbook/Proposals/Base class for FeedbackLoops systems.md
ilyapuchka: what we will have instead then?

Cookbook/Proposals/Base class for FeedbackLoops systems.md
inamiy: My two cents for this solution is to add **`SubState`** instead, and **disallow `Mutation` that receives current state as input** (since current-state referencing i.e. `Reducer` is a complicated mechanism that should only be handled by state machine, not each caller's sending).

```swift
struct State {
    var subState: SubState // Mutation-aware
    var others: ...  // not Mutation-aware
    
    // Explicitly define sub-state for mutation.
    struct SubState {
    	...
    }
}

enum Update {
    case event(Event)
    
    // NOTE: Directly passing a new sub-state WITHOUT referring current sub-state.
    // Behaviour is mostly the same as SwiftUI's direct state binding.
    case syncSubState(SubState)
}
```

This literally means **`SubState` will also behave as sub-action**,
which then allows us to not draw arbitrary state transition arrows, so it's safer.

inamiy: Ah pls forget above.
`SubState` won't be fully able to behave as sub-action, since struct and enum are not the same but dual.

Cookbook/Proposals/Base class for FeedbackLoops systems.md
sergdort:
```
protocol View: Renderable where Self.View == Never {
    func body() -> AnyRenderable
}

Cookbook/Proposals/Base class for FeedbackLoops systems.md
sergdort: yes

Cookbook/Proposals/Base class for FeedbackLoops systems.md
sergdort: @ilyapuchka 
> It also will hurt debugging if handling of all mutations is done in one central place, wouldn't it?

You would need to add it in every view model

Cookbook/Proposals/Base class for FeedbackLoops systems.md
ilyapuchka: I don't think this is related to this proposal

Cookbook/Proposals/Base class for FeedbackLoops systems.md
ilyapuchka: I think these kind of questions should be addressed with extensive code snippets in the document so that we can see the whole picture

Cookbook/Proposals/Base class for FeedbackLoops systems.md
ilyapuchka: but all the mutations are handled here in the base class, no code to handle key path mutations would be in the view models, which seems to be the point of proposal. so considering debugging the only point where we will be able to set breakpoint will be this switch and it will hit it on all view models, there will be no way to debug a single view model conveniently.

sdaniel55:
Closed due to inactivity

PR #232, opened 2019-09-04
Root & TabBar Framework

inamiy:
Successor of #145 .

This proposal introduces a new **Root & TabBar Framework (`BabylonRootTab.framework`)** that is aimed to **decouple Root and Tab from Octopus**.

> Slide and Demo:
> https://drive.google.com/drive/u/1/folders/1mzaBL8f_vwFwFRYjOHaIDB2gTwNOdmLq
>
> JIRA: CNSMR-2646

NOTE: Above new architecture proposal was once a bundle of:

1. Global FlowController, Dispatcher, and Arbitrary Navigation mechanism. [#145](https://github.com/Babylonpartners/ios-playbook/pull/145)
2. Unified Root & TabBar (related: flash-launch)

But **this proposal only supports 2**.

### DIFF
--- a/Cookbook/Proposals/RootTabBarFramework.md
+++ b/Cookbook/Proposals/RootTabBarFramework.md
@@ -5,7 +5,7 @@
 
 ## Introduction
 
-This proposal introduces a new **Root & TabBar Framework (`BabylonRootTab.framework`)** that is first presented in a private team meeting on 15 Aug 2019.
+This proposal introduces a new **Root & TabBar Framework (`BabylonRootTab.framework`)** that is aimed to **decouple Root and Tab from `Octopus`**.
 
 > Slide and Demo:
 > https://drive.google.com/drive/u/1/folders/1mzaBL8f_vwFwFRYjOHaIDB2gTwNOdmLq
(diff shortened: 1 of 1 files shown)


andersio:
If we are to effectively rearchitect the root layer, is it possible to consider Flash Launch, Product Config & SDK Config requirements as part of the design goal here?

For example:

1. We can say that tab construction will no longer wait until `MainUserSession` is ready. They should all be instantiated effectively instantly.
2. To support (1), the app would then need to remember the last known tab combination, so that it can recover the tab bar from app termination without network request.
3. Since the fresh copy of feature switches may affect the tab bar at any time, we would still need the tab bar to react to that.
4. While with (1) we want things to be effectively instantaneous, we might need to consider keeping the whole thing an asynchronous control flow (which is likely to happen as it is RAF based) for Product/SDK/Remote config to have a chance to deter the tab construction if they have no cached state to recover from.

inamiy:
@andersio 
Tab caching and retrieving mechanism can be injected in this RootTab system.
It's up to how we define Session(ViewModel)'s reducer and feedback, and it can be done without having this proposal, then after that, we can migrate to this proposal's approach.

inamiy:
Closing this proposal for now as priority is not high.

PR #245, opened 2019-09-27
[CNSMR-2922] aka [IOSP-151] Report on Build Time Spikes

AliSoftware:
This PR provides a report and some history on the various spikes we tried to improve our build times.

It mainly reports on what we recently tried since I joined. We could consider adding some notes about what we also tried before too (like our past experiments with `Carthage` for example) but not sure we still have details of what was tried on those at that time (Which is also why we're creating this document, to keep track of what we tried at the time in case we want to revisit in the future)


Cookbook/README.md
```diff
@@ -41,15 +41,20 @@ iOS Cookbook 👩‍🍳
 * [Introduce "fish" operator (Kleisli composition)](./Proposals/Fish_Operator.md)
 * [Learn about our functional operators](./Technical-Documents/FunctionalOperators.md)
 
+##### ... learn about our tooling 🛠
```
AliSoftware: The list of articles about tooling that were previously all under "other topics" is starting to grow more and more, so I figured it was about time we moved them to a dedicated section, while also adding the entry for my new document to that list 😉

Cookbook/Technical-Documents/BuildTime-Spikes-Report.md
```diff
@@ -0,0 +1,190 @@
...
+   [ [CNSMR-2847](https://babylonpartners.atlassian.net/browse/CNSMR-2847) ]
+
+
+
+## Helping the type-checker
+
+We've installed two Xcode warnings on all our projects to warn us if the type-checker took:
+
+ - More than 150ms on type-checking single expressions (`-Xfrontend -warn-long-expression-type-checking=150`)
+ - More than 200ms on type-checking entire functions (`-Xfrontend -warn-long-function-bodies=200`
```
AliSoftware:
```suggestion
 - More than 200ms on type-checking entire functions (`-Xfrontend -warn-long-function-bodies=200`)
```

ilyapuchka:
TBH not sure if we need it as part of playbook. This may better belong to internal Jira/Wiki for documentation purposes (we should have epic/spice for this work so this report should be part of it)

Cookbook/Technical-Documents/BuildTime-Spikes-Report.md
```diff
@@ -0,0 +1,190 @@
+# Improving our Build Times
+
+Build Times in Swift are known to be quite slow, but we could still try to find ways to reduce them. So far we've tried the following different tactics to reduce our build times especially on CI and clean builds:
+
+ - [➥](#helping-the-type-checker)
```
ilyapuchka: I mean all these tickets can be part of epic and in the epic or individual tickets we have report with results

Cookbook/Technical-Documents/BuildTime-Spikes-Report.md
```diff
@@ -0,0 +1,190 @@
...
+
+We found a lot of those across the codebase. Most common causes were mainly:
+
+* **Long chains of expressions**, due to our heavily-functional codebase, like `x.map(...).flatMap(...).filter(...)` and similar. Those were in fact essentially coming from:
+  * our use of `Bento` to render screens by composing various rows via `|---+` and `|-+` operators
+  * our use of ReactiveSwift/ReactiveCocoa which tend to create long chains like those
+* **Custom operators, especially heavily-overloaded ones and ones that rely on generics**
+  * Especially `Bento` and `Functional.swift` declare a lot of operators that could lead to these usages at point of use, when we have expressions composed with those operators
+  * But also uses of `+` in many places (yes, just `+`). For example `["a","b","c"].joined()` compiles way faster than `"a" + "b" + "c"`. This is probably both caused by `+` being heavily overloaded, by generic implementations nonetheless, but also by the existence of types conforming to `ExpressibleByStringLiteral` that become candidates that the type-checker has to consider during its analysis in these kind of expressions
```
ilyapuchka: this seems to be more like general advices on how writing code affects compile times and how we agree to do that in the team

ilyapuchka:
@dmcrodrigues what do you think about that? Personally I don't see much point in this being in the playbook. As a post on our findings about cocoa pods-binary limitations and speeding up builds in general - yes, we have a company tech blog for that 🤷‍♂  I'd only leave instructions on what we need to look at encode day to day to ensure that we don't degrade compile time and how to improve it if we do.

Cookbook/Technical-Documents/BuildTime-Spikes-Report.md
AliSoftware: This report and its exposition to the Playbook has been originally requested by @dmcrodrigues , and I agree with him too that we should have that exposed as a document, and not as individual reports in the Epic.

- The goal is to have clear reports of all the things we try in an *accessible* document in case we want to try one of these solutions later and want to remember what we tried, what went wrong, what might have evolved since last time. And I'm pretty sure we'll never think about searching JIRA to try to remember which solution we tried (and all solutions we tried) and what to search. So a single document is better for the future in that regard
 - Then it could indeed be a Jira/Wiki article instead, but I'm not sure I see the point of not exposing that in our playbook. After all, it's open information and valuable experience that we have no reason to hide in private wikis anyway.

Cookbook/Technical-Documents/BuildTime-Spikes-Report.md
AliSoftware: That's also exactly one of the point of the report. While reporting on what we tried to reduce the build time, here's what worked and here is what didn't work, so now that this report is on the playground people can take advantage of that experience and feedback to know what stuff to try rather than others even in future code.

ilyapuchka: But it is hard to extract these bits of something that worked from things that didn't when it all in one huge doc :/

ilyapuchka: I'd say things that didn't work could be restructured into FAQ like `did we try cocoapods-binary?` with short results in form of bullet points of what was the outcome. If anyone would be interested in more in depth explanation we can referent them to some other place, blog post, wiki, whatever
🤷‍♂

j531:
@AliSoftware Nice article (type-checking section was especially interesting). Was there a conclusion around where this was going to be posted? Do we want to keep this open?

AliSoftware:
> @AliSoftware Nice article (type-checking section was especially interesting). Was there a conclusion around where this was going to be posted? Do we want to keep this open?

\cc @AnnKatF

Cookbook/Technical-Documents/BuildTime-Spikes-Report.md
AnnKatF: I totally agree with this. We should definitely expose our findings. It will be beneficial to the whole community 🙂

AnnKatF (approved):
This is an amazing article!! 🙌
Very well thought out! 
Well done!!
🚀

j531 (approved):

PR #247, opened 2019-09-30
BabylonSDK factory for main API interfaces 

viorel15:


Cookbook/Proposals/BabylonSDK.md
```diff
@@ -0,0 +1,77 @@
...
+var healthcheckAPI: HealthcheckAPI
+var healthManagementAPI: HealthManagementAPI
+var healthRecordsAPI: HealthRecordsAPI just V2
+var mapsAPI: MapsAPI
+var notificationsAPI: NotificationsAPI
+var patientMetricsAPI: PatientMetricsAPI
+var userAPI: UserAPI
+}
+```
+`authAPI` is the only constant parameter since an app should only have one instance of that. The rest of API's can be computed variables and it's up to the developer to mange their lifetime.
```
ilyapuchka: does this mean that all the properties will be computed and new instances will be created every time so developers have to retain than themselves?

ilyapuchka: In this case I think having them as a function will be better. Users won't be able to distinguish computed property from stored so can incorrectly assume that the instances are the same, even if it is documented. With functions it is more clear.

mluisbrown (approved):

Cookbook/Proposals/BabylonSDK.md
```diff
@@ -0,0 +1,77 @@
...
+
+The `BabylonSDK` will be the main entry point for any app that uses the Babylon apis.
+`DigitalTwinView` will not change the way it is created and will not be part of the `BabylonSDK`
+
+This interface will be responsible for creating any api.
+`BabylonSDK`will need a configuration object and after that an user can create any set of API's.
+
+```swift
+class BabylonSDK {
+let authAPI: AuthAPI
```
gtsiap: Will all these APIs be `class`es, `struct`s, `protocol`s?

Cookbook/Proposals/BabylonSDK.md
```diff
@@ -0,0 +1,81 @@
...
+var clinicalRecordsAPI: clinicalRecordsAPI
+var appointmentAPI: AppointmentAPI
```
chitrakotwani: Since the context might be nil, shouldn't all these properties be optional? 
I am a bit worried that it does not follow open-closed principle, to extend the SDK, we are going ahead and modifying it. 
Maybe we should think of an alternative

chitrakotwani: Sorry for a very late review!

PR #253, opened 2019-10-08
Make new hire access list easier to find

danspndl:
- Added "access to services" section to make the list easier to find
- Added new iOS Slack channel to the list
- Changed Ruby version to current we use
- Replaced Rui in access list

PR #257, opened 2019-10-14
Update slack channels list

AnnKatF:
**Why?**
We've added two important channels that every iOS Engineers should be aware of.

**How**
By updating the list of channels with
#ios-pedia
#ios-meeting-outcomes


sdaniel55:
@AnnKatF already added these in #253 😊

AnnKatF:
This is being addressed in [#253](https://github.com/Babylonpartners/ios-playbook/pull/253) hence closing it.

PR #267, opened 2019-10-31
Add implicit returns to styleguide

danspndl:
### Why?

Swift 5.1 allows implicit returns

 ### How?

- Added example for implicit returns for functions and computed properties

 ### PR checklist
* [x] I've assigned this PR to myself
* [x] I've set all the necessary PR's labels

 With :heart:


Jimmybee:
So this implies using `return` on single expressions is wrong when it is *clear* (which is left up to who to decide?). I am concerned that this will result in lots of nits, when either is perfectly acceptable and should be left up the code author IMHO.

sdaniel55:
> @Jimmybee: So this implies using `return` on single expressions is wrong when it is _clear_ (which is left up to who to decide?). I am concerned that this will result in lots of nits, when either is perfectly acceptable and should be left up the code author IMHO.

Both are acceptable, the goal would be to allow implicit returns only for single line expressions. I've made the changes more clear

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,14 @@
+### Why?
```
diegopetrucci: Not sure if the template applies for playbook

diegopetrucci (dismissed):
let's wait for banter

Cookbook/Style-guide/README.md
```diff
@@ -381,10 +381,16 @@ Use self only when required by the compiler (in `@escaping` closures, or in init
 
 ## Computed Properties
 
-For conciseness, if a computed property is read-only, omit the get clause. The get clause is required only when a set clause is provided.
+For conciseness, if a computed property is read-only, omit the get clause. The get clause is required only when a set clause is provided. For single-expressions where the context is clear, both explicit and implicit returns can be used:
 
 **Preferred**:
 ```swift
+var diameter: Double {
```
ilyapuchka: how about `var diameter: Double { radius * 2 }`? should we allow this or discourage?

.github/PULL_REQUEST_TEMPLATE.md
inamiy: PR template is unrelated to this PR's title, so pls propose in different PR.

Cookbook/Style-guide/README.md
```diff
@@ -460,7 +468,21 @@ func updateConstraints() -> () {
 typealias CompletionHandler = (result) -> ()
 ```
 
-## Function Calls
+For single-expression functions where the context is clear, both explicit and implicit returns can be used:
```
ilyapuchka: This is similar to closures, do we have a preference to prefer no-return over return in closures? maybe then we should have the same preference here?

inamiy (changes requested):
- [x] Resolves https://github.com/babylonhealth/ios-playbook/pull/267#discussion_r341116843

ilyapuchka:
@Jimmybee we already have this situation with single-expression closures, but IIRC we prefer no-return over return in closures where it is possible. I think with should be consistent here with properties and method. Actually compiler will not compile if there is no return in the closure with more than one expression, at least it was like, not sure if anything changed with this swift proposal implemented and if it's the same with properties and methods

Cookbook/Style-guide/README.md
mluisbrown: AFAIK we have a preference for not using `return` in single expression closures.

mluisbrown (approved):
I think it's enough that the styleguide *allows* not using `return`. However IMO it should not be a rule. So the current changes to the guide are good 👍

mluisbrown:
> let's wait for banter

@diegopetrucci why? This is not a valid reason to block the PR IMO 👎

Jimmybee (approved):
For me, the style guide should be about improving readability of code. Allowing both doesn't detract from that.

Cookbook/Style-guide/README.md
Jimmybee: It looks readable and easily understandable to me. I'd vote 👍

mluisbrown:
@inamiy what changes are you requesting?

inamiy (approved):

Cookbook/Style-guide/README.md
MichalTKwiecien: +1 for adding one-liner

sdaniel55: Added in f91870b

MichalTKwiecien (approved):

diegopetrucci (approved):

sdaniel55:
Any objections? Or can I merge it? /cc @karabatov @ilyapuchka

ilyapuchka (approved):

karabatov (approved):
👍

PR #276, opened 2019-11-08
Change Ruby version instruction to 2.6.4

mluisbrown:


DeFrenZ (approved):
I had to do that afterwards anyway, so makes sense to make it direct 👌

mluisbrown:
Unsure how often this changes. Maybe we should make the instruction version agnostic, and put something like "Find what is the required Ruby version from the `.ruby-version` file in the root of the project" 🤔

danspndl (approved):
🚢

Ben-Henshall:
@mluisbrown Has this requirement changed recently? Might be worth announcing in #ios-pedia to ensure everyone is on the correct version, so this sort of thing isn't always immediately recognisable as a version issue

Ben-Henshall (approved):

sdaniel55:
> @Ben-Henshall: Has this requirement changed recently? Might be worth announcing in #ios-pedia to ensure everyone is on the correct version, so this sort of thing isn't always immediately recognisable as a version issue

It was changed with the Xcode 11 migration

PR #278, opened 2019-11-13
Tech Test - Change test scope

anilputtabuddhi:
As part of improving the Tech test, @sergdort @konrad-em and I have come up with a slightly modified test. This is done to allow for some broader scope in demonstrating technical abilities for a candidate submitting a test to us. 

The primary change has been to move from a list of posts to lists of photos. We hope this allows for better demonstration of technical skills in handling various aspects like proper cell reuse, cancellation of network requests, caching, prefetching etc.

### DIFF
--- a/Interview/demo.md
+++ b/Interview/demo.md
@@ -22,7 +22,7 @@ Before proceeding to the section of your choice, please consider the following t
 
 * We like code that is simple, but [simple is different from easy](https://www.infoq.com/presentations/Simple-Made-Easy).
 * Keep in mind the [SOLID principles](https://en.wikipedia.org/wiki/SOLID_(object-oriented_design)) when doing the project.
-* We left out of the requirements whether or not you should try to download everything (posts + each post detail). This is up to you to decide and to justify.
+* We left out of the requirements whether or not you should try to download everything (photos + each photo detail). This is up to you to decide and to justify.
 * Testing is very important for us. Even if you don't write a single test (for instance, because of time constraints), your app should be easy to test (we love [Dependency injection](https://en.wikipedia.org/wiki/Dependency_injection)).
 * Error scenarios should be taken into consideration and it should be easy to add them, even if you don't explicitly handle them (e.g. using an `UIAlertController`).
 * Although UI and UX are important, we are more concerned in this demo with your thought process and with how you architect your application. Your demo should take into consideration features that might be added in the future.
@@ -34,50 +34,62 @@ Before proceeding to the section of your choice, please consider the following t
 
 ### 1. The babylon demo project
 
-From a high level point of view the demo consists of a list of posts, where each post has its own detail.
+From a high level point of view the demo consists of a list of photos. Upon tapping any photo, the user is taken to the photo detail screen.
 
-#### Posts Screen
+#### Photos List Screen
 
-A post has a title and it's up to you how to display it. To retrieve the posts, you can use the following API:
+A Photo has a thumbnail image and a title and you can decide how best to display it. To retrieve the photos, you can use the following API:
 
-* http://jsonplaceholder.typicode.com/posts
+* http://jsonplaceholder.typicode.com/photos
 
-When a post is tapped, you should go to the detail screen.
+When a photo is tapped, you should go to the Photo detail screen.
+Also, a photo that is marked favourite shows up first in the list followed by other photos.
 
-#### Detail screen
+#### Photo Detail screen
 
-A post detail screen will have:
+A Photo detail screen will have:
 
-* Its author.
-* Its description (via the `body`).
-* The number of comments it has.
+* The user the photo belongs to.
+    * Get the user id from the album(http://jsonplaceholder.typicode.com/albums/{:id})
+    * Get the User from the user detail (http://jsonplaceholder.typicode.com/user/{:id})
+* The title of the photo.
+* The number of comments on the photo:
+    * (https://jsonplaceholder.typicode.com/photos/{:id}/comments)
+* The ability to mark/unmark a photo as favourite.
 
-You can retrieve the remaining information from these API:
 
-* http://jsonplaceholder.typicode.com/users
-* http://jsonplaceholder.typicode.com/comments
 
 #### Requirements
 
 The following requirements should be met:
 
 * Use Swift 4.0 or above.
-* The information (posts and post details) should be available offline. It's assumed that, if it's the first time you are accessing the app and you are offline, you shouldn't see any data.
-* Add reloading mechanism e.g. button or `UIRefreshControl`. `Reachability` is not required.
+* The information (photos and photo details) should be available offline. It's assumed that, if it's the first time you are accessing the app and you are offline, you shouldn't see any data.
+* A reloading mechanism e.g. button or `UIRefreshControl`. `Reachability` is not required.
 * **Have a point of synchronization (e.g. making two concurrent requests and waiting for both of them to finish).**
 * The code should be production grade.
 * It should compile and run.
+* Stutter free, crash free UI.
+* If you are applying for Senior, there are additional Requirements:
+    * You should use a reactive framework
+    * UI should be written in code
+    * Self sizing Cells in TableView / CollectionView
 
 ### 2. Previously written project
 
 We would be happy if you would submit a project you already have (for instance, a demo project for another company). Still, the project **must**:
 
 * Use Swift 4.0 or above.
+* Should have clear instructions on how to run the project.
 * Have at least two distinct network calls.
 * Parse the network response and present the information to the user.
 * Have some sort of persistence mechanism.
 * Compile and run.
 * Have a point of synchronization (e.g. making two concurrent requests and waiting for both of them to finish).
+* If you are applying for Senior, there are additional Requirements:
+    * You should use a reactive framework
+    * UI should be written in code
+    * Self sizing Cells in TableView / CollectionView
 
 If you have a project with these requirements, then perfect! Please ensure that you also have a description of what the project does, in order to give us some context.
 
--- a/Interview/demo.md
+++ b/Interview/demo.md
@@ -26,7 +26,7 @@ Before proceeding to the section of your choice, please consider the following t
 * Testing is very important for us. Even if you don't write a single test (for instance, because of time constraints), your app should be easy to test (we love [Dependency injection](https://en.wikipedia.org/wiki/Dependency_injection)).
 * Error scenarios should be taken into consideration and it should be easy to add them, even if you don't explicitly handle them (e.g. using an `UIAlertController`).
 * Although UI and UX are important, we are more concerned in this demo with your thought process and with how you architect your application. Your demo should take into consideration features that might be added in the future.
-* You can use any 3rd party libraries (except for networking) if you wish ( `ReactiveCocoa`, `PromiseKit`, `Realm`, etc) but be prepared to justify why you did so. Feel free to use package managers to handle them.
+* You can use any 3rd party libraries if you wish ( `Alamofire`, `ReactiveCocoa`, `PromiseKit`, `Realm`, etc) but be prepared to justify why you did so. Feel free to use package managers to handle them.
 * **Be consistent in your code**. We advise using something like [raywenderlich's swift style guide](https://github.com/raywenderlich/swift-style-guide) while doing the demo. It's absolutly fine to use any other style, as long as you are consistent.
 * Clean the file project structure and remove any unused methods (e.g., from `AppDelegate`). This shows attention to detail.
 * Be opinionated regarding any architecture you use and take your time to make it a reflection of your thought process.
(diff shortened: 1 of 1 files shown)


Interview/demo.md
```diff
@@ -34,50 +34,62 @@ Before proceeding to the section of your choice, please consider the following t
...
+* The information (photos and photo details) should be available offline. It's assumed that, if it's the first time you are accessing the app and you are offline, you shouldn't see any data.
+* A reloading mechanism e.g. button or `UIRefreshControl`. `Reachability` is not required.
 * **Have a point of synchronization (e.g. making two concurrent requests and waiting for both of them to finish).**
 * The code should be production grade.
 * It should compile and run.
+* Stutter free, crash free UI.
+* If you are applying for Senior, there are additional Requirements:
+    * You should use a reactive framework
+    * UI should be written in code
+    * Self sizing Cells in TableView / CollectionView
```
sergdort: I'm afraid there is nothing to self-size anymore =) Title of the photo is very short

Interview/demo.md
```diff
@@ -34,50 +34,62 @@ Before proceeding to the section of your choice, please consider the following t
...
 
 * Use Swift 4.0 or above.
+* Should have clear instructions on how to run the project.
 * Have at least two distinct network calls.
 * Parse the network response and present the information to the user.
 * Have some sort of persistence mechanism.
 * Compile and run.
 * Have a point of synchronization (e.g. making two concurrent requests and waiting for both of them to finish).
+* If you are applying for Senior, there are additional Requirements:
+    * You should use a reactive framework
```
sergdort: Should we also require to not use 3rd party lib for Networking?

anilputtabuddhi: Oops!!! Absolutely. Forgot that. Should that be only for Seniors or for everyone?

RuiAAPeres:
I have a feeling that less people will apply, or that some of them will miss the requirements, which translates to an overall less valid candidates.

Interview/demo.md
sergdort: Networking requests in this test is pretty simple, there is no Auth, no Multipart upload no files upload download ¯\_(ツ)_/¯

anilputtabuddhi: ok will add for everyone then

Interview/demo.md
```diff
@@ -34,50 +34,62 @@ Before proceeding to the section of your choice, please consider the following t
...
-* http://jsonplaceholder.typicode.com/comments
 
 #### Requirements
 
 The following requirements should be met:
 
 * Use Swift 4.0 or above.
-* The information (posts and post details) should be available offline. It's assumed that, if it's the first time you are accessing the app and you are offline, you shouldn't see any data.
-* Add reloading mechanism e.g. button or `UIRefreshControl`. `Reachability` is not required.
+* The information (photos and photo details) should be available offline. It's assumed that, if it's the first time you are accessing the app and you are offline, you shouldn't see any data.
```
sergdort: We are thinking that having to display images in the `TableView` or `CollectionView` has a lot more interesting problems to solve then just displaying `Text`.

Interview/demo.md
```diff
@@ -34,50 +34,62 @@ Before proceeding to the section of your choice, please consider the following t
...
+* If you are applying for Senior, there are additional Requirements:
+    * You should use a reactive framework
+    * UI should be written in code
+    * Self sizing Cells in TableView / CollectionView
 
 ### 2. Previously written project
 
 We would be happy if you would submit a project you already have (for instance, a demo project for another company). Still, the project **must**:
 
 * Use Swift 4.0 or above.
```
mluisbrown: Should we change this to Swift 5 or above?

Interview/demo.md
```diff
@@ -34,50 +34,62 @@ Before proceeding to the section of your choice, please consider the following t
...
-You can retrieve the remaining information from these API:
 
-* http://jsonplaceholder.typicode.com/users
-* http://jsonplaceholder.typicode.com/comments
 
 #### Requirements
 
 The following requirements should be met:
 
 * Use Swift 4.0 or above.
```
mluisbrown: Swift 5.0 or above?

Interview/demo.md
mluisbrown: In that case we should remove `Alamofire` from the list of example 3rd party dependencies.

mluisbrown:
> I have a feeling that less people will apply

What in these changes makes you think less people will apply @RuiAAPeres ?

RuiAAPeres:
There have been several rejections because the small list of requirements this test demanded weren't met. On a different dimension, a subset of people complained that the test took them too much time. With this test, the candidates need to code more, not only because the project requirements is broader (e.g. images), but because they are forced into a specific way of doing things.

RuiAAPeres:
I can be absolutely wrong on my assumptions, but this can be easily measure by the amount of people that used to pass the test phase versus with this test.

anilputtabuddhi:
> I can be absolutely wrong on my assumptions, but this can be easily measure by the amount of people that used to pass the test phase versus with this test.

Fair point. But in my experience, a lot of medium to large companies do have such assessments of similar complexity.  We could off-course change the way the recruitment process works by probably having a short chat with the candidate as a first step after CV evaluation. That is another discussion though.

RuiAAPeres:
> But in my experience, a lot of medium to large companies do have such assessments of similar complexity.

The iOS team never benchmarked itself against other companies. It set the standard themselves.

> We could off-course change the way the recruitment process works by probably having a short chat with the candidate as a first step after CV evaluation.

CV evaluation was never important: It what the candidate performance that dictated their level. 

Times change. 😄

RuiAAPeres:
A more realistic test, aligned with what they would have to do, was to ask for the candidate to deliver the project using [Carbon](https://github.com/ra1028/Carbon).

Interview/demo.md
```diff
@@ -22,62 +22,74 @@ Before proceeding to the section of your choice, please consider the following t
...
 * Testing is very important for us. Even if you don't write a single test (for instance, because of time constraints), your app should be easy to test (we love [Dependency injection](https://en.wikipedia.org/wiki/Dependency_injection)).
 * Error scenarios should be taken into consideration and it should be easy to add them, even if you don't explicitly handle them (e.g. using an `UIAlertController`).
 * Although UI and UX are important, we are more concerned in this demo with your thought process and with how you architect your application. Your demo should take into consideration features that might be added in the future.
-* You can use any 3rd party libraries you wish (`Alamofire`, `ReactiveCocoa`, `PromiseKit`, `Realm`, etc) but be prepared to justify why you did so. Feel free to use package managers to handle them.
+* You can use any 3rd party libraries (except for networking) if you wish ( `ReactiveCocoa`, `PromiseKit`, `Realm`, etc) but be prepared to justify why you did so. Feel free to use package managers to handle them.
```
dmcrodrigues: > but be prepared to justify why you did so

Considering this why do we need to exclude networking? Realm is fine but using a network library is not? Because persistence is also one of the requirements of the project so I don't see why we are ok in allowing Realm but then we don't allow anything for networking even if the candidate knows the library quite well and he/she is able to explain what it's happening behind the scenes. Would be great to have a justification about the reasoning for this direction and what's the main benefit, not only for us but for the candidates too.

mluisbrown: I agree with @dmcrodrigues. Why exclude 3rd party networking libs? It's not as if we do any network coding anyway.

anilputtabuddhi: The networking API provided by Apple is powerful and easy enough to setup. It is easy to meet the needs of our Test as well with that simple setup.  Core data on the other hand, which is the Apple preferred mode of handling persistence is not exactly known for easy setup especially for a small tech test. There are other libraries which vastly reduce effort in implementing various aspects - could be powerful animation, automatic diffing etc. This is why we went with that exception. Again if the team feels strongly against it, we can remove the exception. We were just looking to expand scope of test in simple ways to allow candidate some space to exhibit more technical knowledge

anilputtabuddhi:
> A more realistic test, aligned with what they would have to do, was to ask the candidate to use [Carbon](https://github.com/ra1028/Carbon).

Wouldn't asking for implementing a test in a specific library also limit the possibility of a candidate already having done a similar test, or the likelihood of the candidate putting the effort to learn a new framework/library just to submit a test?  

Instead, we are giving the candidate the freedom to choose a reactive framework of their choice, and adding some image downloading and caching requirements.  This kind of tech test is something that a lot of candidates would have done at some point if they are interviewing.  

Agreed this could potentially result in greater tech test rejections. But another measurement that is very valuable to the company is "how many of the selected candidates clear the probation". We have tried to strike a balance between "our way of working" with "general tech tests".  If the candidates self select themselves by choosing not to submit a test that is slightly more aligned with our way of working, surely there are positives to it as well? Off-course this whole exercise could also be revisited again if this change results in catastrophic results (which we hope it doesn't :)).

Interview/demo.md
dmcrodrigues: I personally disagree with the option taken and I don't buy the reasoning behind it, we can always extend the same thought to Core Data, the test is simple enough that shouldn't be too much of a hassle to use Core Data to store the images or anything needed, the APIs have been improving in the last couple of years, Xcode generates a stack for you, ...

To me feels like we are restricting something based on a personal conviction, using a 3rd party project doesn't necessarily mean the candidate doesn't know about the native APIs **but he/she may decide to use a different project to avoid re-inventing the wheel all over again** or are we expecting to have URLs being created directly without any kind of abstraction? We have our own `Resource` type and a couple of abstractions, for a new project something like Moya, as an example, could save the candidate from doing everything from scratch at least for networking, for the rest we say it's ok as long as you are able to justify the usage. To me sounds a bit incoherent and, as I said, there's nothing in the description that explains why we don't allow any 3rd party dependency just for networking.

anilputtabuddhi: Fair enough, lets get rid of this then. @sergdort  and @konrad-em thoughts?

Interview/demo.md
```diff
@@ -22,62 +22,74 @@ Before proceeding to the section of your choice, please consider the following t
...
-#### Detail screen
+#### Photo Detail screen
 
-A post detail screen will have:
+A Photo detail screen will have:
 
-* Its author.
-* Its description (via the `body`).
-* The number of comments it has.
+* The user the photo belongs to.
```
dmcrodrigues: I feel like this one is a breeze of fresh air and may bring new technical challenges, 💯 about changing it 👍

Interview/demo.md
sergdort: What we thought is that the test requires to make simple GET network calls. But on the other hand implementing networking without 3rd party lib give an opportunity for a candidate to show have they would architect certain aspects of Networking, which spawns more interesting conversations during interview.

Interview/demo.md
```diff
@@ -22,62 +22,74 @@ Before proceeding to the section of your choice, please consider the following t
...
+* No 3rd party libraries for networking
+* The information (photos and photo details) should be available offline. It's assumed that, if it's the first time you are accessing the app and you are offline, you shouldn't see any data.
+* A reloading mechanism e.g. button or `UIRefreshControl`. `Reachability` is not required.
 * **Have a point of synchronization (e.g. making two concurrent requests and waiting for both of them to finish).**
 * The code should be production grade.
 * It should compile and run.
+* Stutter free, crash free UI.
+* If you are applying for Senior, there are additional Requirements:
+    * You should use a reactive framework
+    * UI should be written in code
```
mluisbrown: Why was this requirement added, both in this section and the test section? As with the 3rd party networking question, I don't see why we should specify this. We could still have a good test using storyboards and/or XIBs.

Interview/demo.md
sergdort: Sometimes candidates use 3rd party libs for everything and interview becomes very boring.

Interview/demo.md
```diff
@@ -22,62 +22,74 @@ Before proceeding to the section of your choice, please consider the following t
...
+* No 3rd party libraries for networking
+* Should have clear instructions on how to run the project.
 * Have at least two distinct network calls.
 * Parse the network response and present the information to the user.
 * Have some sort of persistence mechanism.
 * Compile and run.
 * Have a point of synchronization (e.g. making two concurrent requests and waiting for both of them to finish).
+* If you are applying for Senior, there are additional Requirements:
+    * You should use a reactive framework
+    * UI should be written in code
```
mluisbrown: As I mentioned above. This is particularly harsh for a pre-existing project.

Interview/demo.md
sergdort: This is a proposal

Interview/demo.md
sergdort: Yeah fair enough 👍

Interview/demo.md
AnnKatF: I understand the reasoning behind this choice. However, IMO I don't think we should have this restriction. We should allow the user to use the libraries they prefer and then we can question that choice in the interview. It can also be interesting to see how the candidate can isolate their code from the library.

Regarding the interview, it is our job to make the interview engaging and interesting regardless of the technical choices of the candidate.

AnnKatF (commented):
Well done @anilputtabuddhi , @sergdort and @konrad-em !! 👏
This is a good improvement to our current tech test.

Just have a concern regarding the restriction of not using 3rd party libraries for networking. As I added in the comment, I think we should give the freedom to the candidate to use the libraries they are comfortable with.

AnnKatF (approved):
Good job! 👏  🚀

mluisbrown (approved):

PR #280, opened 2019-11-13
[Technical test] Add question about backpressure

NSMyself:
Just another question I remembered: the concept of backpressure.

### DIFF
--- a/Interview/questions.md
+++ b/Interview/questions.md
@@ -48,6 +48,7 @@ Please keep in mind that this document is in a continuous state of flux, with ne
  1. Do you know what Curried Functions are? How do they differ from Partial Application?
  1. What is the `pipe` (`|>`) operator? Is it different from `map`? How do you propagate errors with it? Would a `flatPipe` operator make sense?
  1. What is a monad? Have you ever used one? (Related: why do you think FP has so many academic sounding terms?)
+ 1. Can you explain the concept of backpressure?
 
 ### SWIFT
 
--- a/Interview/questions.md
+++ b/Interview/questions.md
@@ -48,6 +48,8 @@ Please keep in mind that this document is in a continuous state of flux, with ne
  1. Do you know what Curried Functions are? How do they differ from Partial Application?
  1. What is the `pipe` (`|>`) operator? Is it different from `map`? How do you propagate errors with it? Would a `flatPipe` operator make sense?
  1. What is a monad? Have you ever used one? (Related: why do you think FP has so many academic sounding terms?)
+ 1. Are you familiar with the Combine framework?
+ 1. In what ways does it differ from ReactiveSwift or RxSwift?
  1. Can you explain the concept of backpressure?
 
 ### SWIFT


karabatov (approved):

Interview/questions.md
```diff
@@ -48,6 +48,7 @@ Please keep in mind that this document is in a continuous state of flux, with ne
  1. Do you know what Curried Functions are? How do they differ from Partial Application?
  1. What is the `pipe` (`|>`) operator? Is it different from `map`? How do you propagate errors with it? Would a `flatPipe` operator make sense?
  1. What is a monad? Have you ever used one? (Related: why do you think FP has so many academic sounding terms?)
+ 1. Can you explain the concept of backpressure?
```
mluisbrown: I would be against adding that question. Until Combine came out none of us had heard about back pressure. It's a fairly obscure topic that isn't relevant to anything that we do at Babylon.

mluisbrown (changes requested):
I don't see a good reason to add a question about back pressure.

Interview/questions.md
NSMyself: But Combine _is_ out and will be around for a while. I'm not saying we should always ask this question but if the interaction between candidate and interviewer happens to address Combine, couldn't this be asked? Combine _is_ backed by Apple, after all.

In any case we need to be able to ascertain which questions to ask based on our interaction with the candidate, that's why we never just follow the list blindly.

I've no strong feelings about this, if you want I can delete this PR.

mluisbrown: If you are talking about Combine, you can always ask about back pressure anyway, even if it's not in the list. I would maybe rephrase this into 2 questions:

1. Are you familiar with the Combine framework?
1. In what ways does it differ from ReactiveSwift or RxSwift?

NSMyself: I can refactor and add those two questions, good call.

Backpressure isn't something specific to Combine though.

mluisbrown: > Backpressure isn't something specific to Combine though.

It doesn't exist in any other reactive frameworks that you might use on iOS. It's part of the [Reactive Streams](https://www.reactive-streams.org) initiative, mostly implemented in Java and Scala AFAIK.

NSMyself: If you're referring to the ability of controlling or handling backpressure then I agree. 
However, the concept itself is framework-independent.

mluisbrown: Yes, the concept is framework independent. But neither RAS nor Rx support it, so it's a concept that most iOS devs have never had a need to be familiar with.

NSMyself: Agreed. That's why I mentioned it shouldn't be always asked. 

After all, we've questions in the playbook that most iOS devs may or may not have had the need to be familiar with (eg: knowledge about sorting algorithms; curried functions and partial application)

NSMyself: Just added your Combine questions @mluisbrown :)

mluisbrown (approved):

PR #289, opened 2019-11-22
[CNSMR-3203] Add task review guidelines

konrad-em:
Ticket: [CNSMR-3203](https://babylonpartners.atlassian.net/browse/CNSMR-3203)

### Why?
Right now there is no consistency in the reviews. We are not aligned in which technical decision is considered good, bad or improvement. We tend to be bias when reviewing the tests. There is no clarity on what we require the engineer to be experienced with.

### How?
Adding a doc page with guidelines.

With :heart:

### DIFF
--- a/Cookbook/Technical-Documents/TaskReviewGuidelines.md
+++ b/Cookbook/Technical-Documents/TaskReviewGuidelines.md
@@ -1,0 +1,60 @@
+
+# Technical test reviewer guidelines
+
+This document describes the guidelines for reviewing technical task assignments that candidates work on during the recruitment process.
+Every question is attributed with **S/M/J** or a combination of those, defining what level it relates to.
+
+- **S** (expected of a senior-level task)
+- **M** (expected of a mid-level task)
+- **J** (expected of a junior-level task)
+
+## General:
+
+### Basics
+- Does it have clear instructions on how to run the project? (SMJ)
+- Are there any compile time issues? (SMJ)
+- Does it meet functional requirements? (SMJ)
+- Are there any run time issues? (SM)
+- Does it have descriptive README? (SM)
+
+### Good practices
+
+- Is it testable? (S, some for M)
+- Does it conform to the SOLID principles? (S, some for M)
+- Does it use dependency injection? (S, some for M)
+- Does it use descriptive naming? (SM)
+- Is the code clean? (no dead code) (SM)
+- Does it use encapsulation? (SM)
+- Are there any warnings in the codebase? (SM)
+- Are modules loosely coupled? (S)
+- Does it use higher order functions? (bonus point)
+- Side effects pushed to the edge? (bonus point)
+- Is it modularized? (bonus point)
+
+## Architecture:
+
+### Networking
+- Does it use `Codable`? (SMJ)
+- Is it well abstracted? (S, some for M)
+- Is it open for extensions? (like authorization, different routes and HTTP methods) (SM)
+- Are errors propagated? (SM)
+- Are requests cancelable? (S)
+
+### Persistence
+- Is it well abstracted? (it does not leak implementation details to other layers) (S)
+
+### UI-Business logic
+- Does `UIViewController` has only its clear responsibilities? (SM, some for J)
+- Is the navigation abstracted? (SM)
+- Is business logic state management done within UI state/actions? (S)
+- Is there a layer in between (Persistence & Networking) and ViewModel/Presenter/ViewController? (S)
+- Does it use unidirectional data flow? (bonus point)
+- Is there an abstraction of `UITableViewDataSource`/`UICollectionViewDataSource` (bonus point)
+
+### UI
+- Is the Auto Layout used? (SMJ)
+- If Auto Layout is done in InterfaceBuilder, are constraints correct and not ambiguous? (SM)
+- If Auto Layout is done in code, is it readable? (S)
+- Are there any Auto Layout warnings in the console? (S)
+- Are views composable? (S)
+
--- a/Cookbook/Technical-Documents/TaskReviewGuidelines.md
+++ b/Cookbook/Technical-Documents/TaskReviewGuidelines.md
@@ -30,8 +30,9 @@ What we mean by `some` is that the requirement does not need to be completely sa
 - Does it use encapsulation? (SM)
 - Are there any warnings in the codebase? (SM)
 - Are modules loosely coupled? (S)
+- Is it performing operations on the right threads? (SM)
+- Does it utilize reference and value semantics? (S)
 - Does it use higher order functions? (bonus point)
-- Side effects pushed to the edge? (bonus point)
 - Is it modularized? (bonus point)
 
 ## Architecture:
@@ -48,15 +49,14 @@ What we mean by `some` is that the requirement does not need to be completely sa
 
 ### UI-Business logic
 - Does `UIViewController` has only its clear responsibilities? (SM, some for J)
-- Is the navigation abstracted? (SM)
+- Is there an abstraction of `UITableViewDataSource`/`UICollectionViewDataSource` (SM)
+- Is navigation from one module to another not tightly coupled? (SM)
 - Is business logic state management done within UI state/actions? (bonus point)
-- Is there a layer in between (Persistence & Networking) and ViewModel/Presenter/ViewController? (S)
+- Is there a data access layer in between (Persistence & Networking) and ViewModel/Presenter/ViewController? (S)
 - Does it use unidirectional data flow? (bonus point)
-- Is there an abstraction of `UITableViewDataSource`/`UICollectionViewDataSource` (bonus point)
 
 ### UI
 - Is the Auto Layout used? (SMJ)
 - If Auto Layout is done in InterfaceBuilder, are constraints correct and not ambiguous? (SM)
 - If Auto Layout is done in code, is it readable? (S)
 - Are there any Auto Layout warnings in the console? (S)
-- Are views composable? (S)
--- a/Cookbook/Technical-Documents/TaskReviewGuidelines.md
+++ b/Cookbook/Technical-Documents/TaskReviewGuidelines.md
@@ -51,9 +51,9 @@ What we mean by `some` is that the requirement does not need to be completely sa
 - Does `UIViewController` has only its clear responsibilities? (SM, some for J)
 - Is there an abstraction of `UITableViewDataSource`/`UICollectionViewDataSource` (SM)
 - Is navigation from one module to another not tightly coupled? (SM)
-- Is business logic state management done within UI state/actions? (bonus point)
 - Is there a data access layer in between (Persistence & Networking) and ViewModel/Presenter/ViewController? (S)
 - Does it use unidirectional data flow? (bonus point)
+- Is business logic state management done within UI state/actions? (bonus point)
 
 ### UI
 - Is the Auto Layout used? (SMJ)
--- a/Cookbook/Technical-Documents/TaskReviewGuidelines.md
+++ b/Cookbook/Technical-Documents/TaskReviewGuidelines.md
@@ -8,6 +8,9 @@ Every question is attributed with **S/M/J** or a combination of those, defining 
 - **M** (expected of a mid-level task)
 - **J** (expected of a junior-level task)
 
+> Note:
+What we mean by `some` is that the requirement does not need to be completely satisfied. For example, an ideal senior level task should have all parts of the app designed for testability, but for a mid level test, we can consider if they exhibit this in a few parts.
+
 ## General:
 
 ### Basics
--- a/Cookbook/Technical-Documents/TaskReviewGuidelines.md
+++ b/Cookbook/Technical-Documents/TaskReviewGuidelines.md
@@ -46,7 +46,7 @@ Every question is attributed with **S/M/J** or a combination of those, defining 
 ### UI-Business logic
 - Does `UIViewController` has only its clear responsibilities? (SM, some for J)
 - Is the navigation abstracted? (SM)
-- Is business logic state management done within UI state/actions? (S)
+- Is business logic state management done within UI state/actions? (bonus point)
 - Is there a layer in between (Persistence & Networking) and ViewModel/Presenter/ViewController? (S)
 - Does it use unidirectional data flow? (bonus point)
 - Is there an abstraction of `UITableViewDataSource`/`UICollectionViewDataSource` (bonus point)
@@ -57,4 +57,3 @@ Every question is attributed with **S/M/J** or a combination of those, defining 
 - If Auto Layout is done in code, is it readable? (S)
 - Are there any Auto Layout warnings in the console? (S)
 - Are views composable? (S)
-


laurenyew-babylon:
Thank you for adding these guidelines. It might be also helpful to include notes about priority for these items. For example, if someone missed making the views composable, is that a deal breaker for an interviewee?

We might also include a brief description on why an item is important (ex: we prefer using composable views b/c ...)

laurenyew-babylon (approved):

Cookbook/Technical-Documents/TaskReviewGuidelines.md
```diff
@@ -0,0 +1,60 @@
...
+- Does it have clear instructions on how to run the project? (SMJ)
+- Are there any compile time issues? (SMJ)
+- Does it meet functional requirements? (SMJ)
+- Are there any run time issues? (SM)
+- Does it have descriptive README? (SM)
+
+### Good practices
+
+- Is it testable? (S, some for M)
+- Does it conform to the SOLID principles? (S, some for M)
```
scass91: How do we quantify `some` here? Is there a minimum amount expected, and are there varying amounts of this quantity for other points throughout this document?

anilputtabuddhi: @scass91  What we mean by `Some` is that an ideal Senior candidate's tech test should have all parts of the app designed for testability, but for a Mid level candidate, we can consider if they exhibit this in a few parts. It is hard to make every part of the app testable, which is why this differentiation. If there are better ways to word this suggestions welcome.

scass91: Ahhh ok so there's no real quantity expected, just a usage in *some* cases where it can be used? That makes sense - thanks.

anilputtabuddhi: I think its better we add a description of what `some` means in the beginning to make it clearer. @konrad-em Thoughts?

anilputtabuddhi (commented):
> Thank you for adding these guidelines. It might be also helpful to include notes about priority for these items. For example, if someone missed making the views composable, is that a deal breaker for an interviewee?

None of the points individually are deal breakers. But we would expect a Senior to address a lot of items marked `S` and similarly for Mid level and Juniors. The assessment will still be a bit subjective because of this.  We wanted to move away from being an overtly objective guideline so that there is some scope for creativity.   

> We might also include a brief description on why an item is important (ex: we prefer using composable views b/c ...)
  
  We assumed the guidelines we are proposing are more or less the expected norms at the said levels - but clearly we are wrong if some things require explanation for their need. Feel free to challenge some or all of the guidelines and we can edit, delete as applicable once we agree :)

Cookbook/Technical-Documents/TaskReviewGuidelines.md
konrad-em: I agree, on second thought I find `some` in this particular case (SOLID) a bit vague. But does it need more explanation in other questions? Let's keep in mind these are rather guidelines than strict step by step instructions and we've intentionally left some space for interpretation :)

Cookbook/Technical-Documents/TaskReviewGuidelines.md
```diff
@@ -0,0 +1,60 @@
...
+- Are errors propagated? (SM)
+- Are requests cancelable? (S)
+
+### Persistence
+- Is it well abstracted? (it does not leak implementation details to other layers) (S)
+
+### UI-Business logic
+- Does `UIViewController` has only its clear responsibilities? (SM, some for J)
+- Is the navigation abstracted? (SM)
+- Is business logic state management done within UI state/actions? (S)
```
sergdort: I would make it as a bonus point because, someone would do that probably only if the do some kind of unidirectional data flow

danspndl (approved):
Looks good to start with! We can always add more points or clarify if needed

viorel15 (approved):
👍

TheAdamBorek (approved):

Cookbook/Technical-Documents/TaskReviewGuidelines.md
```diff
@@ -0,0 +1,62 @@
...
+- Is it testable? (S, some for M)
+- Does it conform to the SOLID principles? (S, some for M)
+- Does it use dependency injection? (S, some for M)
+- Does it use descriptive naming? (SM)
+- Is the code clean? (no dead code) (SM)
+- Does it use encapsulation? (SM)
+- Are there any warnings in the codebase? (SM)
+- Are modules loosely coupled? (S)
+- Does it use higher order functions? (bonus point)
+- Side effects pushed to the edge? (bonus point)
```
ilyapuchka: this needs to be clarified what this means

Cookbook/Technical-Documents/TaskReviewGuidelines.md
```diff
@@ -0,0 +1,62 @@
...
+- Does it use `Codable`? (SMJ)
+- Is it well abstracted? (S, some for M)
+- Is it open for extensions? (like authorization, different routes and HTTP methods) (SM)
+- Are errors propagated? (SM)
+- Are requests cancelable? (S)
+
+### Persistence
+- Is it well abstracted? (it does not leak implementation details to other layers) (S)
+
+### UI-Business logic
```
ilyapuchka: this section except first and last point feels to be like "you get bonus points for doing things the way we want to have them"... And they are pretty vague. IMO both first an last should be required for S and M and so should go first

Cookbook/Technical-Documents/TaskReviewGuidelines.md
```diff
@@ -0,0 +1,62 @@
...
+- Is there a layer in between (Persistence & Networking) and ViewModel/Presenter/ViewController? (S)
+- Does it use unidirectional data flow? (bonus point)
+- Is there an abstraction of `UITableViewDataSource`/`UICollectionViewDataSource` (bonus point)
+
+### UI
+- Is the Auto Layout used? (SMJ)
+- If Auto Layout is done in InterfaceBuilder, are constraints correct and not ambiguous? (SM)
+- If Auto Layout is done in code, is it readable? (S)
+- Are there any Auto Layout warnings in the console? (S)
+- Are views composable? (S)
```
ilyapuchka: what does this mean exactly? and in contrast how "uncompossable" views look like?

ilyapuchka: Also I don't think this should be required of S (depending on what it means), bonus point maybe

Cookbook/Technical-Documents/TaskReviewGuidelines.md
```diff
@@ -0,0 +1,62 @@
...
+## General:
+
+### Basics
+- Does it have clear instructions on how to run the project? (SMJ)
+- Are there any compile time issues? (SMJ)
+- Does it meet functional requirements? (SMJ)
+- Are there any run time issues? (SM)
+- Does it have descriptive README? (SM)
+
+### Good practices
```
ilyapuchka: Should we add something about immutability, value types, multithreading?

Cookbook/Technical-Documents/TaskReviewGuidelines.md
```diff
@@ -0,0 +1,62 @@
...
+- Is it open for extensions? (like authorization, different routes and HTTP methods) (SM)
+- Are errors propagated? (SM)
+- Are requests cancelable? (S)
+
+### Persistence
+- Is it well abstracted? (it does not leak implementation details to other layers) (S)
+
+### UI-Business logic
+- Does `UIViewController` has only its clear responsibilities? (SM, some for J)
+- Is the navigation abstracted? (SM)
```
ilyapuchka: This is a bit vague and one can assume something is a good abstraction when another will consider it bad. Is having a subclass of UINavigationController and interacting it as with abstract `UIViewController` via it's `show` method, that is already performing adaptive presentation, is good abstraction enough or not?

Cookbook/Technical-Documents/TaskReviewGuidelines.md
```diff
@@ -0,0 +1,62 @@
...
+- Is it well abstracted? (S, some for M)
+- Is it open for extensions? (like authorization, different routes and HTTP methods) (SM)
+- Are errors propagated? (SM)
+- Are requests cancelable? (S)
+
+### Persistence
+- Is it well abstracted? (it does not leak implementation details to other layers) (S)
+
+### UI-Business logic
+- Does `UIViewController` has only its clear responsibilities? (SM, some for J)
```
ilyapuchka:
```suggestion
- Does `UIViewController` has only its clear responsibilities? (SM, some for J)
- Is there an abstraction of `UITableViewDataSource`/`UICollectionViewDataSource` (SM)
```

Cookbook/Technical-Documents/TaskReviewGuidelines.md
```diff
@@ -0,0 +1,62 @@
...
+- Is the navigation abstracted? (SM)
+- Is business logic state management done within UI state/actions? (bonus point)
+- Is there a layer in between (Persistence & Networking) and ViewModel/Presenter/ViewController? (S)
+- Does it use unidirectional data flow? (bonus point)
+- Is there an abstraction of `UITableViewDataSource`/`UICollectionViewDataSource` (bonus point)
```
ilyapuchka:
```suggestion
```

Cookbook/Technical-Documents/TaskReviewGuidelines.md
```diff
@@ -0,0 +1,62 @@
...
+- Are requests cancelable? (S)
+
+### Persistence
+- Is it well abstracted? (it does not leak implementation details to other layers) (S)
+
+### UI-Business logic
+- Does `UIViewController` has only its clear responsibilities? (SM, some for J)
+- Is the navigation abstracted? (SM)
+- Is business logic state management done within UI state/actions? (bonus point)
+- Is there a layer in between (Persistence & Networking) and ViewModel/Presenter/ViewController? (S)
```
ilyapuchka: Not sure I understand what we mean by this layer, what's its responsibility?

ilyapuchka (changes requested):
UI-Business logic section related comments are blockers in my opinion

Cookbook/Technical-Documents/TaskReviewGuidelines.md
anilputtabuddhi: Yes. It is true we have said bonus points for unidirectional data flow and patterns that help in making an architecture move towards that. While it is true there are other alternate ways of architecting an app, we thought someone already comfortable with unidirectional flow will better appreciate the approach we take at Babylon and by extension - likely to enjoy the work, pass probation and win-win :). A bonus point is not an expectation - we would not reject a candidate if he/she hasn't got a single bonus point. But getting them means they are more aligned to our thinking.

Cookbook/Technical-Documents/TaskReviewGuidelines.md
anilputtabuddhi: Probably we can change the wording to "Is navigation from module to another not tightly coupled?" @sergdort & @konrad-em ?

Cookbook/Technical-Documents/TaskReviewGuidelines.md
anilputtabuddhi: This is a layer which abstracts how persistence/networking works from the presentation layer objects like VM/VCs. Typically like BusinessControllers in Babylon, or other alternate names I have seen are UseCases, Interactors etc. Should we elaborate more on this?. Any suggestion of alternate wording would be welcome.

ilyapuchka: data access layer?

Cookbook/Technical-Documents/TaskReviewGuidelines.md
konrad-em: Agreed, I removed this one. It's not clear enough and it does not makes sense to require this from a demo project.

Cookbook/Technical-Documents/TaskReviewGuidelines.md
konrad-em: I've removed this one as it is to vague and it's not a requirement anyway.

Cookbook/Technical-Documents/TaskReviewGuidelines.md
```diff
@@ -0,0 +1,62 @@
...
+- Does it have clear instructions on how to run the project? (SMJ)
+- Are there any compile time issues? (SMJ)
+- Does it meet functional requirements? (SMJ)
+- Are there any run time issues? (SM)
+- Does it have descriptive README? (SM)
+
+### Good practices
+
+- Is it testable? (S, some for M)
+- Does it conform to the SOLID principles? (S, some for M)
```
bh-pwestmeyer: SOLID is kind of old school. I'd be surprised if anyone knew about this.

Cookbook/Technical-Documents/TaskReviewGuidelines.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Networking
+- Does it use `Codable`? (SMJ)
+- Is it well abstracted? (S, some for M)
+- Is it open for extensions? (like authorization, different routes and HTTP methods) (SM)
+- Are errors propagated? (SM)
+- Are requests cancelable? (S)
+
+### Persistence
+- Is it well abstracted? (it does not leak implementation details to other layers) (S)
```
bh-pwestmeyer: This and dependency injection are opposing guidelines.

Cookbook/Technical-Documents/TaskReviewGuidelines.md
bh-pwestmeyer: I think you're underestimating how rare this pattern is. It's been hard for us to find anyone doing RFP and unidirectional with pro experience.

Cookbook/Technical-Documents/TaskReviewGuidelines.md
```diff
@@ -0,0 +1,62 @@
...
+
+### UI-Business logic
+- Does `UIViewController` has only its clear responsibilities? (SM, some for J)
+- Is there an abstraction of `UITableViewDataSource`/`UICollectionViewDataSource` (SM)
+- Is navigation from one module to another not tightly coupled? (SM)
+- Is business logic state management done within UI state/actions? (bonus point)
+- Is there a data access layer in between (Persistence & Networking) and ViewModel/Presenter/ViewController? (S)
+- Does it use unidirectional data flow? (bonus point)
+
+### UI
```
bh-pwestmeyer: Given that our UIs are supposed to be built with our custom UIKit wrapper, is this even relevant?

konrad-em: Well, sometimes we need to build some new component. Following this logic we don't need to know networking because we have stack we already use. Sure, reactive programming, unidirectional data flow etc is. something we deal with everyday, more frequently than UIKit, but I think we should expect some basic UI to be done properly.

Cookbook/Technical-Documents/TaskReviewGuidelines.md
konrad-em: Not sure I follow, can you elaborate on this? How is this opposing guidelines?

Cookbook/Technical-Documents/TaskReviewGuidelines.md
konrad-em: Is it? IMO, it's principles still apply. But maybe we should not expect task to be fully compliant given it's just a demo task and make it a bonus point? What do you think?

konrad-em:
> UI-Business logic section related comments are blockers in my opinion

@ilyapuchka I made some changes in this section. Do you still consider it to be a blocker?

Cookbook/Technical-Documents/TaskReviewGuidelines.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Persistence
+- Is it well abstracted? (it does not leak implementation details to other layers) (S)
+
+### UI-Business logic
+- Does `UIViewController` has only its clear responsibilities? (SM, some for J)
+- Is there an abstraction of `UITableViewDataSource`/`UICollectionViewDataSource` (SM)
+- Is navigation from one module to another not tightly coupled? (SM)
+- Is business logic state management done within UI state/actions? (bonus point)
+- Is there a data access layer in between (Persistence & Networking) and ViewModel/Presenter/ViewController? (S)
```
ilyapuchka: [nit] I'd move what is required before everything that is just for bonus points

ilyapuchka (approved):
Looks good to me now 👍

PR #292, opened 2019-11-26
[CNSMR-3180] Proposal: Decouple snapshot tests from unit tests

TheAdamBorek:
On our H2 goal list there is a goal about decoupling snapshot tests from unit tests so we are able to run only unit tests which are much more faster than snapshots.

This is a proposal how this could be implemented. This is a continuation of a quick survey I did on our slack: https://babylonhealth.slack.com/archives/G22EWKXQT/p1574330951269300

### DIFF
--- a/Cookbook/Proposals/DecoupleSnapshotTestsFromUnitTests.md
+++ b/Cookbook/Proposals/DecoupleSnapshotTestsFromUnitTests.md
@@ -28,4 +28,4 @@ We would also need to create a new scheme like `BabylonUniTestsOnly` which would
 
 ## Alternatives considered
 
-- **Xcode's 11 Test plans** - Unfortunately, I haven't found a way to select test files per configuration. I'm afraid we cannot use test plans.
+- **Xcode's 11 Test plans** - TestPlans won't help us in context of this proposal. We would still need to have 2 tests targets, even if we use test plans. Since TestPlans are not necessary and doesn't make this proposal any simpler, I don't think we should enable them as part of the proposal.
--- a/Cookbook/Proposals/DecoupleSnapshotTestsFromUnitTests.md
+++ b/Cookbook/Proposals/DecoupleSnapshotTestsFromUnitTests.md
@@ -4,13 +4,15 @@
 ## Introduction
 The purpose of this proposal is to decouple snapshot tests from unit tests to be able to run only unit-tests which takes much less time to execute than snapshot tests.
 
+The goal of this proposal is to make it possible to run only unit-tests **locally** in Xcode. Improving CI run time is not a goal of this proposal.
+
 ## Motivation
-Right now when we run all unit-tests (by pressing cmd+U) we also run snapshot tests. Right now it takes **14 minutes**(😱) to run all unit+snapshot tests which is very long.\
+Right now when we run all unit-tests (by pressing cmd+U) we also run snapshot tests. Right now it takes **14 minutes**(😱) to run all unit+snapshot tests which is very long. Unit-tests should run as quickly as possible so we can have a quick feedback about stability of our changes.\
 \
-Ideally, it would be nice to be able to run only snapshot tests when we don't work on UI.
+Ideally, it would be nice to be able to run only unit-tests when we don't work on UI.
 
 ## Proposed solution
-Decouple snapshot tests from unit test by creating SnapshotTest targets in every framework which has snapshots.
+Decouple snapshot tests from unit test by creating `*SnapshotTests` targets in every framework which has snapshots.
 
 We would also need to create a new scheme like `BabylonUniTestsOnly` which would run only tests targets without snapshot tests targets.
 
@@ -24,7 +26,7 @@ We would also need to create a new scheme like `BabylonUniTestsOnly` which would
 
 ## Alternatives considered
 
-- **Xcode's 11 Test plans** - Unfortunately, I haven't found a way to select test files per a configurations. I'm afraid we cannot use test plans.
+- **Xcode's 11 Test plans** - Unfortunately, I haven't found a way to select test files per configuration. I'm afraid we cannot use test plans.
 
 
 
--- a/Cookbook/Proposals/DecoupleSnapshotTestsFromUnitTests.md
+++ b/Cookbook/Proposals/DecoupleSnapshotTestsFromUnitTests.md
@@ -28,4 +28,4 @@ We would also need to create a new scheme like `BabylonUnitTests` which would ru
 
 ## Alternatives considered
 
-- **Xcode's 11 Test plans** - TestPlans won't help us in context of this proposal. We would still need to have 2 tests targets, even if we use test plans. Since TestPlans are not necessary and doesn't make this proposal any simpler, I don't think we should enable them as part of the proposal.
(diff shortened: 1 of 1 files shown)


Cookbook/Proposals/DecoupleSnapshotTestsFromUnitTests.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Impact on existing codebase
+
+1. We would have more tests targets
+2. Once we create any test file we would need to be cautious to add it to the correct test target. Unit tests into `*Tests`, snapshot tests into `*SnapshotTests`
+3. We would need to have additional scheme like `BabylonUnitTestsOnly` which would run only unit-tests on `cmd+u`.
+
+## Alternatives considered
+
+- **Xcode's 11 Test plans** - Unfortunately, I haven't found a way to select test files per configuration. I'm afraid we cannot use test plans.
```
ilyapuchka: can you elaborate on that? Test plans can be created from scheme, so if we create BabylonUnitTests scheme it can be as well converted to a test plan. And schemes can have multiple test plans assigned to them. 

https://shashikantjagtap.net/wwdc19-getting-started-with-test-plan-for-xctest/

> You can run the test using the Xcode ->Product -> Test for your scheme as usual which executes the attached test plan for the scheme. In our case, it will execute both smoke and regression test configured in the functional test plan.

It's not necessary that we should use test plans as they might not bring enough value for us, but the conclusion that we can't use them at all seems premature

The other thing we would need to explore in regards of test plans is using them with fastlane

Cookbook/Proposals/DecoupleSnapshotTestsFromUnitTests.md
```diff
@@ -0,0 +1,31 @@
...
+Right now when we run all unit-tests (by pressing cmd+U) we also run snapshot tests. Right now it takes **14 minutes** to run all unit+snapshot tests which is very long. Those 14 minutes doesn't include the build time 😱.
+
+Unit-tests should run as quickly as possible so we can have a quick feedback about stability of our changes.\
+\
+Ideally, it would be nice to be able to run only unit-tests when we don't work on UI. Unit-tests in separation take about ~2% of the execution time.
+
+## Proposed solution
+Decouple snapshot tests from unit test by creating `*SnapshotTests` targets in every framework which has snapshots.
+
+We would also need to create a new scheme like `BabylonUniTestsOnly` which would run only tests targets without snapshot tests targets.
```
ilyapuchka:
```suggestion
We would also need to create a new scheme like `BabylonUnitTests` which would run only tests targets without snapshot tests targets.
```

Cookbook/Proposals/DecoupleSnapshotTestsFromUnitTests.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Introduction
+The purpose of this proposal is to decouple snapshot tests from unit tests to be able to run only unit-tests which takes much less time to execute than snapshot tests.
+
+The goal of this proposal is to make it possible to run only unit-tests **locally** in Xcode. Improving CI run time is not a goal of this proposal.
+
+## Motivation
+Right now when we run all unit-tests (by pressing cmd+U) we also run snapshot tests. Right now it takes **14 minutes** to run all unit+snapshot tests which is very long. Those 14 minutes doesn't include the build time 😱.
+
+Unit-tests should run as quickly as possible so we can have a quick feedback about stability of our changes.\
```
anilputtabuddhi:
```suggestion
Unit-tests should run as quickly as possible, so we can have a quick feedback about our changes that break tests.\
```

Cookbook/Proposals/DecoupleSnapshotTestsFromUnitTests.md
```diff
@@ -0,0 +1,31 @@
+# Decouple Snapshot tests from unit tests
+* Author(s): Adam Borek
+
+## Introduction
+The purpose of this proposal is to decouple snapshot tests from unit tests to be able to run only unit-tests which takes much less time to execute than snapshot tests.
+
+The goal of this proposal is to make it possible to run only unit-tests **locally** in Xcode. Improving CI run time is not a goal of this proposal.
+
+## Motivation
+Right now when we run all unit-tests (by pressing cmd+U) we also run snapshot tests. Right now it takes **14 minutes** to run all unit+snapshot tests which is very long. Those 14 minutes doesn't include the build time 😱.
```
anilputtabuddhi:
```suggestion
Right now, when we trigger all unit-tests (by pressing cmd+U), the snapshot tests are also run.  It takes **14 minutes** to run all unit+snapshot tests which is very long. Those 14 minutes doesn't include the build time 😱.
```

anilputtabuddhi (approved):

Cookbook/Proposals/DecoupleSnapshotTestsFromUnitTests.md
TheAdamBorek: Yes, I used a wrong word here. We can use test plans. TestPlans just won't help in context of this proposal. We would still need 2 targets, one for unit and one for snapshot tests.

I'll change the explanation above to be more clear.

ilyapuchka: don't they allow to include different test cases without defining different targets?

TheAdamBorek: It's tricky. They do, however, if you add a new test file then they are automatically enabled in every TestPlan which uses the target. That means, if I add a `NewFeatureSnapshotTest.swift` it would be enabled in both TestPlans for unit-tests and snapshot-tests.

ilyapuchka: Did I hear someone saying "Danger"?...
But ok, got your point. This can be included to the doc imo

AnnKatF (approved):

Cookbook/Proposals/DecoupleSnapshotTestsFromUnitTests.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Motivation
+Right now, when we trigger all unit-tests (by pressing cmd+U), the snapshot tests are also run.  It takes **14 minutes** to run all unit+snapshot tests which is very long. Those 14 minutes doesn't include the build time 😱.
+
+Unit-tests should run as quickly as possible, so we can have a quick feedback about our changes that break tests.\
+\
+Ideally, it would be nice to be able to run only unit-tests when we don't work on UI. Unit-tests in separation take about ~2% of the execution time.
+
+## Proposed solution
+Decouple snapshot tests from unit test by creating `*SnapshotTests` targets in every framework which has snapshots.
```
AliSoftware: Side note be I think we could use that occasion to rename our test targets too, i.e. use a dash to separate main name from suffix. That way we'd have `BabylonFoo-Tests` (or `BabylonFoo-UnitTests`, we can bikeshed), `BabylonFoo-SnapshotTests` and `BabylonFoo-UITests` – avoiding the ambiguity of names like `BabylonFooUITests` that are not clear if they are about Unit `Tests` for `BabylonFooUI`, or `UITests` for `BabylonFoo`

Sure it's not the direct topic of this proposal, but since this proposal will add more test targets to the project, I think it's the right time to make sure the name of each test target are clear in intent.

AliSoftware (approved):
👍

Cookbook/Proposals/DecoupleSnapshotTestsFromUnitTests.md
TheAdamBorek: Yeah, I agree this ambiguity is irritating. I could also change this :)

AliSoftware:
@TheAdamBorek While grooming our platform backlog I found this old issue: https://babylonpartners.atlassian.net/browse/IOSP-203 which seem to describe exactly the same plan that you presented in the proposal.

Probably worth using this ticket instead of creating a new one (if it's not too late ; or link the two tickets together if you already created a new one) to be consistent in our JIRA boards.

PR #294, opened 2019-11-28
Update OOO steps for calendar event

AliSoftware:
Now that @AnnKatF has changed the process in #272 from creating a meeting with invites to creating an event directly on the calendar instead, the step about having to turn "Request responses" off is not relevant anymore


danspndl (approved):
🚢

konrad-em (approved):
👍

PR #300, opened 2019-12-09
Update PR labels documention as per decision in the latest iOS retro

diegopetrucci:
It was decided that labels for vertical should be deleted. Only `US` and `platform` will stay, as they are special cases.


NSMyself (approved):
😢

diegopetrucci:
Should the `Telus` label stay?

DeFrenZ (approved):
That was quick! 👏

emeseuk (approved):

PR #310, opened 2019-12-31
Update AddingFrameworks.md

MichalTKwiecien:
While following the doc, I found out that one character is missing and causing cocoapods to generate an error message.


Daniel1of1 (approved):
Nice catch 🚢

danspndl (approved):
👍

PR #334, opened 2020-01-30
PR automatically assigned to their creators

NSMyself:
@ilyapuchka added a GitHub Action to automatically assign a PR to its creator. This action runs exactly once (when the PR itself is created) and should take about 30s for it to do its thing


mluisbrown (approved):

danspndl (approved):
🎉

PR #336, opened 2020-02-04
Adds documentation for icons assets SwiftGen integration

adiki:
Image assets SwiftGen integration is already merged to develop. This document explains how to add new icon and what's the logic with retrieving images.


Cookbook/Technical-Documents/SwiftGen.md
```diff
@@ -0,0 +1,32 @@
...
+	}
+	return result
+}
+```
+The method without specifying bundle automatically chooses `Bundle.main`. The algorithm is following:
+1. Method checks if the icon is added in assets catalog in the main bundle. In most cases it will be white label app bundle (this allows overriding the icon for specific app).
+2. If the icon is not there it takes icon from DesignLibrary assets catalog.
+3. In other case `fatalError` is called, but this situation is impossible because icon has to be in DesignLibrary assets catalog because SwiftGen created identifier for it. (If icon is deleted SwiftGen will regenerate file without identifier for that icon).
+
+When some app overrides icon with the same name in its assets catalog but it still would like to use icon from DesignLibrary in other case it can specify bundle: `Bundle(for: IconsBundleToken.self)` and it will enforce that image from DesignLibrary assets catalog will retrieved.
```
AliSoftware:
```suggestion
When an app overrides an icon with the same name in its assets catalog but you would still like to use the icon from DesignLibrary in some cases, you can explicitly pass `DesignLibrary.bundle` to enforce that the image is loaded from the DesignLibrary assets catalog.
```

Note: Would actually have `DesignLibrary.bundle` (see `DesignLibrary.swift` line 35) which feels easier to understand and read than `Bundle(for: IconsBundleToken.self)`.

It would actually be nice to update the main repo to get rid of `IconsBundleToken` in our custom SwiftGen template, and use `DesignLibrary.bundle` instead.

AliSoftware (approved):
Some wording/typos suggestions, and a note about existence of `DesignLibrary.bundle`.

Otherwise LGTM 👍

Cookbook/Technical-Documents/SwiftGen.md
```diff
@@ -0,0 +1,34 @@
+## SwiftGen
+
+[SwiftGen](https://github.com/SwiftGen/SwiftGen) allows plenty of possibilities to automatically generate code based on resources we store in our project. Following sections will describe which types of resources are currently integrated with SwiftGen and how to use them.
+
+### Image assets
+
+Using icons and images with our SwiftGen integration is really easy. The only thing you have to do is to add the image to `BabylonDesignLibrary/DesignLibrary/Assets.xcassets` and then you can just use it by calling `appearance.tokens.icons.image(.myNewIcon)`. Before compilation of `BabylonDesignLibrary`, SwiftGen is executed and regenerates the identifiers.
```
ilyapuchka: | then you can just use it...
doesn't it require first to build the app? We might want to have this as a Xcode extension in the future so that the build is not required, or at least to mention how to call SwiftGen manually from command line

ilyapuchka (approved):

Cookbook/Technical-Documents/SwiftGen.md
adiki: If you will use identifier without compiling first you will probably get an error from the Xcode but it will be removed during compilation and it won't fail.

adiki: What you won't have will be autocompletion

adiki: I can add that note

daniloaliberti (approved):
👍great work man 👏

PR #342, opened 2020-02-10
Property wrappers for custom decoding

ilyapuchka:
With property wrappers we can drastically decrease amount of boilerplate we have for decoding/encoding without changing how we access properties that require custom decoding, which was not fully possible before property wrappers.

* [ ] I will send a meeting invitation, using this [template](https://raw.githubusercontent.com/babylonhealth/ios-playbook/master/Cookbook/Proposals/Template_Proposal_Meeting_Invitation.md), scheduled for 2 weeks after this proposal is made, so an agreement can be reached.


Cookbook/Proposals/PropertyWrappers.md
```diff
@@ -0,0 +1,155 @@
...
+
+ `decodeIfPresent` still fails when value is of a wrong type (i.e. unknown enum case) so we use custom `decodeSafelyIfPresent` helper method that returns `nil` in case of error
+ 
+ - dates decoding
+
+ usually dates are decoded as strings and then flat-mapped using some date formatting helper method
+ 
+ - decoding nested values into self
+
+ sometimes we need to extract nested keys without preserving intermediate structure, for what we use custom `NestedKeyDecodable` helper and cusotm parsing methods
```
mluisbrown:
```suggestion
 sometimes we need to extract nested keys without preserving intermediate structure, for which we use custom `NestedKeyDecodable` helper and custom parsing methods
```

Cookbook/Proposals/PropertyWrappers.md
```diff
@@ -0,0 +1,155 @@
...
+ 
+ - decoding self values into nested objected
+
+ opposite to the previous keys sometimes we decode nested objects from the flat json structure, for which we directly pass decoder to `init(decoder:)`
+ 
+All these cases require us to implement decoding for all other properties as well, which could otherwise be decoded by compiler generated code.
+ 
+### Detailed design
+
+Swift compiler is capable of generating decoding code for properties wrapped with property wrappers in a non disruptive way (i.e. it does not require changing coding keys to match names of wrapper properties generated by the compilers). For that property wrappers should implemented coding protocols. Additionally it's possible to extendion standard library decoding containers with decoding methods for specific types of wrappers without writing decoding code for the whole type - container will pick up extension methods if the types in the signature match. This way we can customise decoding for non-standard cases in various ways.
```
mluisbrown:
```suggestion
The Swift compiler is capable of generating decoding code for properties wrapped with property wrappers in a non disruptive way (i.e. it does not require changing coding keys to match names of wrapper properties generated by the compilers). For that property wrappers should implement coding protocols. Additionally it's possible to extend standard library decoding containers with decoding methods for specific types of wrappers without writing decoding code for the whole type - the container will pick up extension methods if the types in the signature match. This way we can customise decoding for non-standard cases in various ways.
```

Cookbook/Proposals/PropertyWrappers.md
```diff
@@ -0,0 +1,155 @@
...
+
+Simplest example of a property wrapper would be if we make a `ISO8601Date` a property wrapper. This only requires to add a `@propertyWrapper` annotation to this type. It already implements decoding for dates so we don't need to change that. What changes is only how we declare properties of this type.
```
mluisbrown:
```suggestion
The simplest example of a property wrapper would be if we make an `ISO8601Date` property wrapper. This only requires to add a `@propertyWrapper` annotation to this type. It already implements decoding for dates so we don't need to change that. What changes is only how we declare properties of this type.
```

Cookbook/Proposals/PropertyWrappers.md
```diff
@@ -0,0 +1,155 @@
...
+```swift
+struct Appointment: Decodable {
+  @ISO8601Date
+  let time: Date
+}
+```
+
+Reference: https://speakerdeck.com/alisoftware/and-thats-a-wrap?slide=25
+
+Another example is pruned/strict arrays decoding. Pruned strategy means that any array element that fails decoding will be discarded when with strict strategy the whole decoding will fail on any failure in decoding of any array element (default behaviour). To implement prune strategy we use `FailableDecodable` that decodes its value with `value = try? container.decode(T.self)` or `decodeArraySafely` method. With property wrapper we wouldn't need to to call this method manually or change the type of the property. We can also implement more strategies, like non-empty, using the same property wrapper.
```
mluisbrown:
```suggestion
Another example is pruned/strict array decoding. The pruned strategy means that any array element that fails decoding will be discarded whereas with the strict strategy the whole decoding will fail on any failure in decoding of any array element (default behaviour). To implement prune strategy we use `FailableDecodable` that decodes its value with `value = try? container.decode(T.self)` or `decodeArraySafely` method. With a property wrapper we wouldn't need to to call this method manually or change the type of the property. We can also implement more strategies, like non-empty, using the same property wrapper.
```

Cookbook/Proposals/PropertyWrappers.md
```diff
@@ -0,0 +1,155 @@
...
+After:
+
+```swift
+struct DemographicsListDTO: Decodable {
+  @ArrayDecodable<Prune, DemographicDTO>
+  let demographics: [DemographicDTO]
+}
+```
+
+Providing defaults for absent values has a bit tricky ergonomics, as this approach does not allow to pass arbitrary parameters to property wrapper initialisers when they are being decoded (so `@Default(value: true)` is possible programmatically but won't affect decoding). So instead of using values we use types which would correspond to common defaults, i.e. `True`, `False`, `Empty` etc:
```
mluisbrown:
```suggestion
Providing defaults for absent values has slightly tricky ergonomics, as this approach does not allow to pass arbitrary parameters to property wrapper initialisers when they are being decoded (so `@Default(value: true)` is possible programmatically but won't affect decoding). So instead of using values we use types which would correspond to common defaults, i.e. `True`, `False`, `Empty` etc:
```

Cookbook/Proposals/PropertyWrappers.md
```diff
@@ -0,0 +1,155 @@
...
+    case isRead = "status"
+  }
+  
+  enum IsReadKeys: String, CodingKey, CaseIterable {
+    case isRead = "is_read"
+  }
+}
+```
+
+Here we specify additional coding keys type that will be used to decode `isRead`. Then container extension method takes care of traversing json using this additional keys and their order (via CaseIterable) as a key path.
```
mluisbrown:
```suggestion
Here we specify an additional coding keys type that will be used to decode `isRead`. Then the container extension method takes care of traversing json using these additional keys and their order (via CaseIterable) as a key path.
```

Cookbook/Proposals/PropertyWrappers.md
```diff
@@ -0,0 +1,155 @@
...
+
+1. We can come up we a single property wrapper to describe decoding strategy and provide the actual strategy as a generic parameter:
+
+```swift
+@Decoding<Default<Empty>>
+@Decoding<Array<Prune>>
+@Decoding<Nested<IsReadKeys>>
+```
+
+This might be a good way to unify approach, but it's a bit harder to read and it we would need to figure out how to constrain particular strategies to particular types of wrapped values, which seems to be easier with a more straight forward approach.
```
mluisbrown:
```suggestion
This might be a good way to unify the approach, but it's a bit harder to read and we would need to figure out how to constrain particular strategies to particular types of wrapped values, which seems to be easier with a more straight forward approach.
```

mluisbrown (approved):

Ben-Henshall (approved):
beautiful

MichalTKwiecien (approved):
🥳

Cookbook/Proposals/PropertyWrappers.md
```diff
@@ -0,0 +1,155 @@
...
+  @NestedDecodable<Bool, IsReadKeys>
+  var isRead: Bool // { "status": { "is_read": true } }
+  
+  enum CodingKeys: String, CodingKey {
+    case isRead = "status"
+  }
+  
+  enum IsReadKeys: String, CodingKey, CaseIterable {
+    case isRead = "is_read"
+  }
```
chitrakotwani: Everything else is a lovely improvement! But I am not too sure about property wrapper for nested types, it does not look very intuitive and there is some boiler plate code involved.

chitrakotwani (approved):

Cookbook/Proposals/PropertyWrappers.md
```diff
@@ -0,0 +1,155 @@
...
+
+This might be a good way to unify the approach, but it's a bit harder to read and we would need to figure out how to constrain particular strategies to particular types of wrapped values, which seems to be easier with a more straight forward approach.
+
+2. [KeydCodable](https://github.com/dgrzeszczak/KeyedCodable) provides an alternative approach where custom keys type and custom decoders are used along with some property wrapper. While it's a nice solution it requires much more code to implement (most of it - boilerplate) or we would need to depend on this library which is undesirable when the same can be achieved with simple property wrappers.
+
+### References
+
+https://github.com/gonzalezreal/DefaultCodable
+https://github.com/GottaGetSwifty/CodableWrappers
+https://github.com/dgrzeszczak/KeyedCodable
```
daniloaliberti:
```suggestion
https://github.com/gonzalezreal/DefaultCodable

https://github.com/GottaGetSwifty/CodableWrappers

https://github.com/dgrzeszczak/KeyedCodable
```

to avoid this 
Screen Shot 2020-03-05 at 11 38 07

daniloaliberti (approved):
🚀

TheAdamBorek (approved):

AnnKatF (approved):
~Just one thing - Please link this document to our cookbook index.
Thanks!~
This will be automatically linked in the proposals. 🤦‍♀ 

🚢 it!

Cookbook/Proposals/PropertyWrappers.md
ilyapuchka: Can you elaborate what is not intuitive and what boilerplate is needed? The only thing that is needed is to define CodingKeys in a straightforward way, and only to do that for the nested key. Comparing with a boilerplate we otherwise have to write to decode it manually its much less. There are also alternative approaches to this case https://twitter.com/ilyapuchka/status/1228062000682610690?s=20

ilyapuchka: For simple cases though we already have a solution (`decodeNested`) that we can keep using.

ilyapuchka:
Will create a story in the bau board to tackle that

PR #347, opened 2020-02-19
Update squads and members

Naiko:
Update squads and members on the README

### DIFF
--- a/scripts/squads.yml
+++ b/scripts/squads.yml
@@ -138,6 +138,10 @@ roles:
 # List of the Squads and their members
 #
 squads:
+  - name: Activate
+    desc: Onboard users so that they understand the value we provide.
+    members:
+    - *Sergey
   - name: Appointments 1
     desc: Face to face appointment booking.
     members:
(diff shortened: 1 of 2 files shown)


NSMyself (changes requested):
I’m afraid I have left Triage. I am now working for the assistants tribe

danspndl (approved):
👌 Thanks!

scripts/squads.yml
```diff
@@ -65,6 +58,9 @@ team:
   - &James
     name: James Birtwell
     github: jimmybee
+  - &Javier
+    name: Javier Hernández
+    github: jaherhi
```
jaherhi: Thanks!

jaherhi (approved):

NSMyself (approved):
👌

scripts/squads.yml
```diff
@@ -192,6 +177,18 @@ squads:
...
+    desc: Engineering work like tooling, CI and development processes.
+    members:
+      - *David
+      - *Ilya
+      - *Olivier
+  - name: Onboarding and Navigation
+    desc: General user experience in the app.
+    members:
+      - *Sergey
+      - *Yuri
```
karabatov: So instead of “Onboarding and navigation” we should have the following:
```
  - name: Activate
     desc: Onboard users so that they understand the value we provide.
     members:
       - *Sergey
  - name: Retain
     desc: Engage users with our key features in a dynamic, personalised manner.
     members:
       - *Yuri
```

PR #350, opened 2020-02-22
[IOSP-593] Update Danger documentation with new rules [check_absolute_paths] [swiftlint]

AliSoftware:
* While fixing absolute paths in pbxproj via https://github.com/babylonhealth/babylon-ios/pull/10686 I added a new Danger rule to prevent that from happening again in the future
* I also took the occasion to properly extract the `swiftlint` checks in a dedicated file to keep things tidy and consistent with the expected organisation of our Dangerfile

This PR updates the doc accordingly, documenting both the new `check_absolute_paths` rule and adding missing documentation for the existing `swiftlint` rule.


danspndl (approved):
✨

NSMyself (approved):

PR #369, opened 2020-04-07
Update README.md

chitrakotwani:


j531 (approved):

zopa-simon-cass (commented):
Vojtech also needs to be removed 😿

danspndl (changes requested):
@chitrakotwani `scripts/squads.yaml` should be updated, then run `scripts/squads.rb` to update the README

chitrakotwani:
> scripts/squads.yaml should be updated, then run scripts/squads.rb to update the README

We need a doc on how to update doc :trollface: 😆

PR #387, opened 2020-06-01
Update Apple Pay document

jaherhi:
Updated lane name in the Apple Pay document, and added a section around code signing as we're not using our default configuration for running the app in this scenario.

Also updated the Fastlane document, as I got an extra step when I added my device.


gtsiap (approved):

viorel15 (approved):
👍

PR #388, opened 2020-06-05
Update client release cut steps

danspndl:
## Why?

We tried the automated cut with the Telus release and it didn't work perfectly 😬 

## How?

Added explainer to the `branch` parameter and which value to use in certain scenarios

With ❤️


zopa-simon-cass (approved):
:elmorise:

brocoo (approved):

PR #401, opened 2020-10-02
Bye 👋

gtsiap:
Delete any references to me from playbook.

### DIFF
--- a/scripts/squads.yml
+++ b/scripts/squads.yml
@@ -26,9 +26,6 @@ team:
   - &Emese
     name: Emese Toth
     github: emeseuk
-  - &Giorgos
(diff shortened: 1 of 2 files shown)


konrad-em (approved):
😢

scripts/squads.yml
```diff
@@ -106,7 +103,6 @@ squads:
   - name: Experience
     desc: Engage users with our key features in a dynamic, personalised manner.
     members:
-      - *Giorgos
```
Naiko, line 109: Should we delete that squad altogether?

sdaniel55: yes, the script will fail otherwise

danspndl (approved):
😢

